PR #19, opened 2019-08-06
topolvm-hook: reimplement using kubebuilder v2

ymmt2005:
`topolvm-hook` was written as a plain webhook server using [k8s.io/client-go](https://github.com/kubernetes/client-go).

This pull request rewrites it using [kubebuilder](https://github.com/kubernetes-sigs/kubebuilder) and [controller-runtime](https://github.com/kubernetes-sigs/controller-runtime).

As a consequence, the new implementation:
- fixes a bug that drops existing resources of Pod.
- fixes a bug that crashes the server if PVC has no storage class.
- rotates TLS certificates.
- exports Prometheus metrics at `/metrics` on port 8080.
- has integration tests using the real apiserver and etcd.


hook/run.go
```diff
@@ -0,0 +1,71 @@
...
+	if err := clientgoscheme.AddToScheme(scheme); err != nil {
+		panic(err)
+	}
+
+	// +kubebuilder:scaffold:scheme
+}
+
+// Run runs the webhook server.
+func Run(cfg *rest.Config, webhookHost string, webhookPort int, metricsAddr, certDir string) error {
+	ctrl.SetLogger(zap.Logger(false))
```
zoetrope: It is difficult to read a non-development log.
I want to be able to set a flag like topolvm-node.
https://github.com/cybozu-go/topolvm/blob/c015ea67a8f789f0eb53c498ae92cb55b9311192/pkg/topolvm-node/cmd/root.go#L75

ymmt2005:
I will address the command in another PR.

PR #23, opened 2019-08-13
filesystem: stabilize mount point detection

ymmt2005:
The device filepath appearing in /proc/mounts can be any file
in the filesystem if the file is the device file having the
same major/minor device numbers.


ymmt2005:
All green!
https://circleci.com/gh/cybozu-go/workflows/topolvm/tree/no-proc-mounts

PR #29, opened 2019-08-20
Kuberntes 1.15

masa213f:
To update topolvm to apply Kubernetes 1.15, this pull request made the following changes.

- Add `preserveUnknownFields: false` to CRDs in order to use `apiextensions.k8.io/v1` in future 
- Update sidecar containers for TopoLVM
- Use matchPolicy: Equivalent for webhook configurations in e2e-test for TopoLVM

### DIFF
--- a/e2e/Makefile
+++ b/e2e/Makefile
@@ -3,7 +3,7 @@ GOFLAGS=-mod=vendor
 export GOFLAGS
 GINKGO=$(GOPATH)/bin/ginkgo
 KUBECTL=/usr/local/bin/kubectl
-KUBERNETES_VERSION=1.14.3
+KUBERNETES_VERSION=1.15.0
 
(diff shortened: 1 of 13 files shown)


e2e/Makefile
```diff
@@ -3,7 +3,7 @@ GOFLAGS=-mod=vendor
 export GOFLAGS
 GINKGO=$(GOPATH)/bin/ginkgo
 KUBECTL=/usr/local/bin/kubectl
-KUBERNETES_VERSION=1.14.3
+KUBERNETES_VERSION=1.15.0
```
ymmt2005: Should we use 1.15.3?
kind can also be updated to v0.5.1.

binoue: This is used in line 41 and when we checked, `kindest/node:v1.15.3` didn't exist.
However, as the image was made yesterday, we use `KUBERNETES_VERSION=1.15.3`

PR #44, opened 2019-09-05
Upgrade `kind` version in example

binoue:
When upgrading Kubernetes from 1.14 to 1.15, we updated apiVersion of `ClusterConfiguration` from `kubeadm.k8s.io/v1beta1` to `kubeadm.k8s.io/v1beta2` with `kind` upgrade from v0.4.0 to v0.5.1

Unfortunately, `topolvm` example is still using `kind` v0.4.0 and it can not use `kubeadm.k8s.io/v1beta2` of `ClusterConfiguration`.
So currently, example is not working.

Thus this pull request made the following change.
- upgrade `kind` version from v0.4.0 to v0.5.1

### DIFF
--- a/example/Makefile
+++ b/example/Makefile
@@ -3,7 +3,8 @@ GOFLAGS=-mod=vendor
 export GOFLAGS
 GINKGO=$(GOPATH)/bin/ginkgo
 KUBECTL=/usr/local/bin/kubectl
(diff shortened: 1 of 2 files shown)


binoue:
> I think it is better to use kind@0.5.1 and k8s@1.15 in the example.

Okay. I will fix that way.

PR #51, opened 2019-09-12
Improve resource deletion with node

mitsutaka:
Topolvm needs manual operation if node resource is deleted or executing `kubectl drain` as shown in [this issue](https://github.com/cybozu-go/topolvm/issues/41) 

To fix this, this pull request made the following changes.
- add a new controller` topolvm-controller` which deletes pvcs before deleting nodes.
- add a finalizer to clean up stale `logicalvolume` resource.
- change `logicalvolume` CustomResourceDefinition resource from namespace-wide resource to cluster-wide resource.  
- reflect changes to documents

Note:
- This pull  request contains breaking change because `logicalvolume` CRD changed. Thus before using `topolvm` including this pull request, it is required to delete all created `logicalvolumes` in advance.
- Only deleting pvcs when deleting nodes creates pending pods. #53 will solve this problem.


masa213f:
duplicated #53

PR #86, opened 2020-01-28
Add support for volume tags to lvmd

matthias50:
In support of: https://github.com/cybozu-go/topolvm/issues/75


lvmd/command/lvm.go
```diff
@@ -27,6 +27,7 @@ var ErrNotFound = errors.New("not found")
 func CallLVM(cmd string, args ...string) error {
 	args = append([]string{cmd}, args...)
 	c := exec.Command(lvm, args...)
+	fmt.Printf("Invoking LVM command: %v\n", c.Args)
```
ymmt2005: Does this a leftover for debugging?

matthias50: I originally put it in there for debugging, but it seemed to me it would be useful to leave in so that you can see exactly what command are being sent to lvm so I intentionally left it in.

ymmt2005: Got it.  For logging purposes, `lvmd` uses `github.com/cybozu-go/log` package elsewhere.
Please replace `fmt.Printf` with `log.Info`, like this:

```suggestion
	log.Info("invoking LVM command", map[string]interface{}{
	    "args": args,
	})
```

matthias50: Got it. For updates to a PR in process... Do you want to see a separate commit on my branch with the updates because of comments or a force push of an amended commit so there will only be one commit in the end? I'm ok with either way. Just want to know which way you want me to go.

ymmt2005: I prefer the latter; i.e., amended & force push.  Thanks!

matthias50: As an FYI, this does result in having duplicated timestamps in the lvmd logs. For example:

```
Jan 29 17:27:40 ip-10-126-138-175 lvmd[1798]: 2020-01-29T17:27:40.357761Z ip-10-126-138-175 lvmd info: "invoking LVM command" args="[lvcreate -n f69fa870-6926-433f-8608-d6b1368725a9 -L 1g -W y -y myvg2]"
```

ymmt2005: Are you watching `journalctl` log or syslog?  Then, they add a timestamp for each line of logs.

`cybozu-go/log` is designed to be self-contained, therefore has its own timestamp; if logs are output to files instead of stderr (and processed by journald or systemd), each log line has just one timestamp.

matthias50: I'm watching journalctl so that seems to be why I'm seeing that.

satoru-takeuchi:
LGTM. Thank you very much for your contribution. @matthias50

PR #89, opened 2020-01-31
Update CSIDriver config for k8s 1.16 for e2e while leaving legacy alone

matthias50:
In support of: https://github.com/cybozu-go/topolvm/issues/75. e2e
tests for this PR require fields only present in 1.16 or greater
so the e2e config needs to vary based on release.


e2e/legacy/controller.yaml
```diff
@@ -0,0 +1,6 @@
+- op: replace
+  path: /spec/podInfoOnMount
+  value: false
+
+- op: remove
+  path: /spec/volumeLifecycleModes
```
ymmt2005: I cannot find any references of `spec.volumeLifecycleModes` in the current manifests.
Why this `op` is needed?

matthias50: I added this to```spec.volumeLifecycleModes``` in e2ebase. I haven't touched the ```deploy/manifests``` yet. Should the changes I made in e2ebase instead go in the ```deploy/manifests```? Perhaps along with an update to https://github.com/cybozu-go/topolvm/blob/master/deploy/README.md instructing users to remove this if they are deploying on k8s versions prior to 1.16?

ymmt2005: Ah, got it.  Thanks.

PR #99, opened 2020-02-13
webhook: Fix to allow creating Pods before their PVCs

binoue:
As creating Pods before their PVCs is a valid behavior in other StorageClasses, we modified the TopoLVM webhook to allow the behavior in this pull request.

Along with this change, we updated `limitation.md`.

### DIFF
--- a/docs/limitations.md
+++ b/docs/limitations.md
@@ -11,3 +11,12 @@ Ref: https://kubernetes.io/docs/concepts/storage/volumes/#csi
 
 > The `csi` volume type does not support direct reference from Pod and may
 > only be referenced in a Pod via a `PersistentVolumeClaim` object. 
+
+Pod without PVC
+---------------
+
+TopoLVM expects that PVCs are created in advance of their Pods.
+However, the TopoLVM webhook does not prevent the creation of Pods before their PVCs. This is because such usages are valid in other StorageClasses and the webhook cannot identify the StorageClasses without PVCs.
+TopoLVM does not care about the created pods without their PVCs even if the PVCs are deployed afterward.
+
+The typical usage of TopoLVM is using StatefulSet with volumeClaimTemplate.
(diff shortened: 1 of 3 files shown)


docs/limitations.md
```diff
@@ -11,3 +11,12 @@ Ref: https://kubernetes.io/docs/concepts/storage/volumes/#csi
 
 > The `csi` volume type does not support direct reference from Pod and may
 > only be referenced in a Pod via a `PersistentVolumeClaim` object. 
+
+Pod without PVC
+---------------
+
+TopoLVM expects that PVCs are created in advance of their Pods.
+However, the TopoLVM webhook does not prevent the creation of Pods before their PVCs. This is because such usages are valid in other StorageClasses and the webhook cannot identify the StorageClasses without PVCs.
```
ymmt2005:
```suggestion
However, the TopoLVM webhook does not block the creation of a Pod when there are missing PVCs for the Pod. This is because such usages are valid in other StorageClasses and the webhook cannot identify the StorageClasses without PVCs.
```

docs/limitations.md
```diff
@@ -11,3 +11,12 @@ Ref: https://kubernetes.io/docs/concepts/storage/volumes/#csi
...
+TopoLVM does not care about the created pods without their PVCs even if the PVCs are deployed afterward.
```
ymmt2005:
```suggestion
For such Pods, TopoLVM's extended scheduler will not work.
```

morimoto-cybozu:
The added test looks good.
Why did you modify `getPod()`?

binoue:
> The added test looks good.
> Why did you modify `getPod()`?

Considering the possibility that we expect to return error by the `getPod()` function, I thought it would be better to modify the return values of the `getPod()` function.

However, as that is currently not necessary, I rolled the function back.

PR #114, opened 2020-03-05
Make sure to install custom-checker when make test.

kfyharukz:


kfyharukz:
I think golang version is not so frequently changed, so it is enough that `custom-checker` would be deleted when `make clean`, isn't it? @ymmt2005

ymmt2005:
I don't think to delete the binary in `make clean` is enough because
people don't do `make clean` always.

Could you explain why you avoided `go run`?
If `go run` cannot be used anyways, I'd recommend to run `go install` everytime
when we run the check.  `go install` effectively caches the object files for each
Go language version.

PR #118, opened 2020-03-12
Update controller-runtime, grpc-go, and controller-tools 

binoue:
This pull request made the following changes:
- update controller-runtime to v0.5.1
- update grpc-go to v1.28.0
- update controller-tools to v0.2.7
- modify webhook test by setting up with WebhookInstallOptions
- add the `matchpolicy` field to `MutatingWebhookConfiguration `


hook/mutate_persistentvolumeclaim.go
```diff
@@ -28,7 +28,7 @@ func PVCMutator(c client.Client, dec *admission.Decoder) http.Handler {
 	return &webhook.Admission{Handler: persistentVolumeClaimMutator{c, dec}}
 }
 
-// +kubebuilder:webhook:path=/pvc/mutate,mutating=true,failurePolicy=fail,groups="",resources=persistentvolumeclaims,verbs=create,versions=v1,name=pvc-hook.topolvm.cybozu.com
+// +kubebuilder:webhook:path=/pvc/mutate,mutating=true,failurePolicy=fail,matchPolicy="Equivalent",groups="",resources=persistentvolumeclaims,verbs=create,versions=v1,name=pvc-hook.topolvm.cybozu.com
```
masa213f:
```suggestion
// +kubebuilder:webhook:path=/pvc/mutate,mutating=true,failurePolicy=fail,matchPolicy=equivalent,groups="",resources=persistentvolumeclaims,verbs=create,versions=v1,name=pvc-hook.topolvm.cybozu.com
```

I think it is good to uncapitalize the value and remove double-quotes just like other markers.

hook/mutate_pod.go
```diff
@@ -20,7 +20,7 @@ import (
 
 var pmLogger = logf.Log.WithName("pod-mutator")
 
-// +kubebuilder:webhook:path=/pod/mutate,mutating=true,failurePolicy=fail,groups="",resources=pods,verbs=create,versions=v1,name=pod-hook.topolvm.cybozu.com
+// +kubebuilder:webhook:path=/pod/mutate,mutating=true,failurePolicy=fail,matchPolicy="Equivalent",groups="",resources=pods,verbs=create,versions=v1,name=pod-hook.topolvm.cybozu.com
```
masa213f:
```suggestion
// +kubebuilder:webhook:path=/pod/mutate,mutating=true,failurePolicy=fail,matchPolicy=equivalent,groups="",resources=pods,verbs=create,versions=v1,name=pod-hook.topolvm.cybozu.com
```

hook/mutate_persistentvolumeclaim.go
dulltz: Note: both of `Equivalent` and `equivalent` are valid.

https://github.com/kubernetes-sigs/controller-tools/blob/1c6d3844762fed588be5123e5b38ccbeb25d94e1/pkg/webhook/parser.go#L190-L201

dulltz: > I think it is good to use lowercase value and remove double-quotes just like other markers.

 I agree with it.

PR #141, opened 2020-05-12
Move to a new home

llamerada-jp:
change path of container image in quay.io


llamerada-jp:
use github actions for release workflow

PR #144, opened 2020-05-28
node: recreate device file if needed when expanding volume

dulltz:
Fix https://github.com/cybozu-go/topolvm/issues/143

### DIFF
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -887,6 +887,36 @@ spec:
 			return nil
 		}, timeout).Should(Succeed())
 
+		By("deleting topolvm-node Pods to clear /dev/topolvm/*")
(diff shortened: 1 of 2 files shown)


e2e/e2e_test.go
```diff
@@ -887,6 +887,36 @@ spec:
 			return nil
 		}, timeout).Should(Succeed())
 
+		By("deleting topolvm-node Pods to clear /dev/topolvm/*")
+		stdout, stderr, err = kubectlWithInput([]byte(podYAML), "delete", "-n", ns, "pod", "-l=app.kubernetes.io/name=node")
```
zoetrope, line 891: Is it necessary to pass `podYaml`?

dulltz: no, I fixed it. thanks

PR #161, opened 2020-07-15
GitHub Actions

masa213f:
- Add GitHub Actions workflows
  - `build` ... Execute unit test and e2e test.
  - `release` ... Build a container image and upload asset to GitHub Release.
- Refactor Makefile
  - Add some make targets.
  - Separate a makefile for csi-sidecars.
  - Get version variable value when building.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -10,7 +10,7 @@ jobs:
         with:
           go-version: 1.13
         id: go
-      - run: make setup SUDO=""
+      - run: make setup
       - run: |
             make manifests generate
             diffs=$(git diff --name-only)
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -1,0 +1,19 @@
+name: Main
+on: [push]
+jobs:
+  build:
+    name: build
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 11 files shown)


zoetrope:
When would you delete .circleci/config.yaml?
I worry about the release conflicting.

ysksuzuki:
Looks good to me other than ikezoe-san's comments above.

tapih:
> When would you delete .circleci/config.yaml?
> I worry about the release conflicting.

I can remove circleci config in this PR, but the change will be huge.
So, after this PR is merged, I am going to make new PR just for removing the circleci config.

tapih:
> Currently we are deciding whether or not to run a test depending on `CIRCLECI` environment variables.
> 
> https://github.com/cybozu-go/topolvm/blob/9d5902a2d133b6775bde8db3ed1130c178a31f39/lvmd/lvservice_test.go#L20
> 
> 
> https://github.com/cybozu-go/topolvm/blob/c0c629cf4faf854bdceeef6371854c2b8346b0ba/filesystem/interface_test.go#L46
> 
> Do you need to change this for GitHub Actions?

All the workflows run on a virtual machine, so I think the current code runs tests as we intended.
`circleci := os.Getenv("CIRCLECI") == "true" ` will be deleted at the same time I purge the circleci config.

PR #164, opened 2020-07-16
chmod volume permission to 2777.

kfyharukz:
Implements https://github.com/cybozu-go/topolvm/issues/152
Signed-off-by: Kazuhito MATSUDA <kazuhito.matuda@gmail.com>


driver/node.go
```diff
@@ -182,7 +182,7 @@ func (s *nodeService) nodePublishFilesystemVolume(req *csi.NodePublishVolumeRequ
 		}
 	}
 
-	err = os.MkdirAll(req.GetTargetPath(), 0755)
+	err = os.MkdirAll(req.GetTargetPath(), 2777)
```
ymmt2005, line 185: What happens if the target directory already exists?
Shouldn't we use `os.Chmod` for that case?

ymmt2005: This is changing the mode of directory _before_ mount.
What we need to do is to change the directory mode _after_ mount.

PR #176, opened 2020-08-01
Add FAQ entry to describe why TopoLVM doesn't use extended resources

ymmt2005:


ymmt2005:
@kfyharukz friendly ping

PR #182, opened 2020-08-11
Fix to accept implicit StorageClassName

### LINKED ISSUE #180: Topolvm prevent us to create a PVC with an implicit storageClassName
**Describe the bug**

As stated in the [kubernetes doc](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistent-volumes), it is perfectly legal to have a PV and a PVC linked together through a storageClassName which is a simple name, without explicit storageClass definition. 

It seems this is the standard way to provide access to an NFS volume. 

But, the topolvm admission webhook prevent us to do so. 

**Environments**
- Kubernetes:
- centos78:
- topolvm v0.5.2

**To Reproduce**

On a cluster with topolvm installed:

1. Apply the following PV:
```
apiVersion: v1
kind: PersistentVolume
metadata:
  name: nfs1data1
spec:
  capacity:
    storage: 20Gi
  volumeMode: Filesystem
  accessModes:
    - ReadWriteMany
  persistentVolumeReclaimPolicy: Retain   # Retain, Recycle or Delete
  storageClassName: nfs1data1 
  nfs:
    path: /data01/vol1
    server: nfs1
```
2. Apply the following PVC
```
kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  name: nfs1data1claim
spec:
  accessModes:
  - ReadWriteMany
  resources:
    requests:
      storage: 1Gi
  storageClassName: nfs1data1
```    
Got:
Error from server: error when creating "pvc.yaml": admission webhook "pvc-hook.topolvm.cybozu.com" denied the request: StorageClass.storage.k8s.io "nfs1data1" not found

**Expected behavior**

As my PVC is 'legal', I would like topolvm to let it pass through

**Additional context**

 None

chez-shanpu:
fix #180

- Accept implicit StorageClassName
- Add test

### DIFF
--- a/hook/mutate_persistentvolumeclaim_test.go
+++ b/hook/mutate_persistentvolumeclaim_test.go
@@ -66,7 +66,7 @@ var _ = Describe("pvc mutation webhook", func() {
 		Expect(hasFinalizer).Should(Equal(false), "finalizer should not be set for storageclass=%s", hostLocalStorageClassName)
 	})
 
-	It("should not have topolvm.cybozu.com/pvc finalizer when pvc has a the StorageClassName that does not exist", func() {
+	It("should not have topolvm.cybozu.com/pvc finalizer when the specified StorageClass does not exist", func() {
 		pvcName := "unexists-storageclass-pvc"
 		createPVC(unexistStorageClassName, pvcName)
 		pvc, err := getPVC(pvcName)
--- a/hook/mutate_persistentvolumeclaim_test.go
+++ b/hook/mutate_persistentvolumeclaim_test.go
@@ -66,6 +66,15 @@ var _ = Describe("pvc mutation webhook", func() {
 		Expect(hasFinalizer).Should(Equal(false), "finalizer should not be set for storageclass=%s", hostLocalStorageClassName)
 	})
 
+	It("should not have topolvm.cybozu.com/pvc finalizer when pvc has a the StorageClassName that does not exist", func() {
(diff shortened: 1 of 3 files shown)


hook/mutate_persistentvolumeclaim_test.go
```diff
@@ -66,6 +66,15 @@ var _ = Describe("pvc mutation webhook", func() {
 		Expect(hasFinalizer).Should(Equal(false), "finalizer should not be set for storageclass=%s", hostLocalStorageClassName)
 	})
 
+	It("should not have topolvm.cybozu.com/pvc finalizer when pvc has a the StorageClassName that does not exist", func() {
```
ymmt2005, line 69:
```suggestion
	It("should not have topolvm.cybozu.com/pvc finalizer when the specified StorageClass does not exist", func() {
```

hook/suite_test.go
```diff
@@ -42,6 +42,7 @@ const (
 	topolvmProvisioner3StorageClassName         = "topolvm-provisioner3"
 	topolvmProvisionerImmediateStorageClassName = "topolvm-provisioner-immediate"
 	hostLocalStorageClassName                   = "host-local"
+	unexistStorageClassName                     = "doesnot-exists"
```
ymmt2005, line 45: `doesnot-exists` looks strange.  The same applies to `unexist`.

```suggestion
	missingStorageClassName                     = "missing-storageclass"
```

PR #183, opened 2020-08-11
Add support to run example on Vagrant

frederiko:
This PR adds support to run the example, e2e, using Vagrant. This is useful for those not running on Linux with no LVM support.

### DIFF
--- a/e2e/Makefile
+++ b/e2e/Makefile
@@ -29,6 +29,7 @@ endif
 GO_FILES := $(shell find .. -path ../vendor -prune -o -path ../e2e -prune -o -name '*.go' -print)
 CA_FILES=./certs/ca.csr ./certs/ca.pem ./certs/ca-key.pem
 SERVER_CERT_FILES=./certs/server.csr ./certs/server.pem ./certs/server-key.pem
+BACKING_STORE=./build
 
 $(KUBECTL):
 	$(SUDO) curl -sfL https://storage.googleapis.com/kubernetes-release/release/v$(KUBERNETES_VERSION)/bin/linux/amd64/kubectl -o $(KUBECTL)
@@ -71,28 +72,28 @@ shutdown-kind:
 start-lvmd:
 	mkdir -p build /tmp/build
 	go build -o build/lvmd ../pkg/lvmd
-	if [ -f /tmp/build/backing_store* ]; then $(MAKE) stop-lvmd; fi
+	if [ -f $(BACKING_STORE)/backing_store* ]; then $(MAKE) stop-lvmd; fi
 
 	for i in $$(seq 3); do \
 		mkdir -p /tmp/topolvm/worker$$i; \
 		mkdir -p /tmp/topolvm/lvmd$$i; \
-		truncate --size=20G /tmp/build/backing_store$${i}_1; \
-		$(SUDO) losetup -f /tmp/build/backing_store$${i}_1; \
-		$(SUDO) vgcreate -y node$${i}-myvg1 $$($(SUDO) losetup -j /tmp/build/backing_store$${i}_1 | cut -d: -f1); \
+		truncate --size=20G $(BACKING_STORE)/backing_store$${i}_1; \
+		$(SUDO) losetup -f $(BACKING_STORE)/backing_store$${i}_1; \
+		$(SUDO) vgcreate -y node$${i}-myvg1 $$($(SUDO) losetup -j $(BACKING_STORE)/backing_store$${i}_1 | cut -d: -f1); \
 		$(SUDO) lvcreate -y -n csi-node-test-block -L 1G node$${i}-myvg1; \
 		$(SUDO) lvcreate -y -n csi-node-test-fs -L 1G node$${i}-myvg1; \
 	done
 
 	# Create additional Volume Groups
-	truncate --size=10G /tmp/build/backing_store1_2; \
-	$(SUDO) losetup -f /tmp/build/backing_store1_2; \
-	$(SUDO) vgcreate -y node1-myvg2 $$($(SUDO) losetup -j /tmp/build/backing_store1_2 | cut -d: -f1); \
-	truncate --size=10G /tmp/build/backing_store2_2; \
-	$(SUDO) losetup -f /tmp/build/backing_store2_2; \
-	$(SUDO) vgcreate -y node2-myvg2 $$($(SUDO) losetup -j /tmp/build/backing_store2_2 | cut -d: -f1); \
-	truncate --size=10G /tmp/build/backing_store3_3; \
-	$(SUDO) losetup -f /tmp/build/backing_store3_3; \
-	$(SUDO) vgcreate -y node3-myvg3 $$($(SUDO) losetup -j /tmp/build/backing_store3_3 | cut -d: -f1); \
+	truncate --size=10G $(BACKING_STORE)/backing_store1_2; \
+	$(SUDO) losetup -f $(BACKING_STORE)/backing_store1_2; \
+	$(SUDO) vgcreate -y node1-myvg2 $$($(SUDO) losetup -j $(BACKING_STORE)/backing_store1_2 | cut -d: -f1); \
+	truncate --size=10G $(BACKING_STORE)/backing_store2_2; \
+	$(SUDO) losetup -f $(BACKING_STORE)/backing_store2_2; \
+	$(SUDO) vgcreate -y node2-myvg2 $$($(SUDO) losetup -j $(BACKING_STORE)/backing_store2_2 | cut -d: -f1); \
+	truncate --size=10G $(BACKING_STORE)/backing_store3_3; \
+	$(SUDO) losetup -f $(BACKING_STORE)/backing_store3_3; \
+	$(SUDO) vgcreate -y node3-myvg3 $$($(SUDO) losetup -j $(BACKING_STORE)/backing_store3_3 | cut -d: -f1); \
 
 	for i in $$(seq 3); do \
 		$(SUDO) systemd-run --unit=lvmd$$i.service $(shell pwd)/build/lvmd --config=$(shell pwd)/lvmd$$i.yaml; \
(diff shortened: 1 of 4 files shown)


zoetrope:
@frederiko 
Thank you for your pull request. I think this is a very good idea.
Could you describe how to use this Vagrantfile in `example/README.md`?

frederiko:
Yes, I can. Will do.

frederiko:
@zoetrope I was thinking that, to keep things consistent, I was considering to add something like `make vm` (and its counterpart `make vm-clean`) to get all the Vagrant stuff up and running. Would it be overkill?

frederiko:
> @zoetrope I was thinking that, to keep things consistent, I was considering to add something like `make vm` (and its counterpart `make vm-clean`) to get all the Vagrant stuff up and running. Would it be overkill?

On a second thought, it would be overkill, as the user can also use Vagrant in Linux/Windows/Macosx and it would be too just for the simple case of trying the example.

zoetrope:
@frederiko 
I think README is sufficient.
Windows users may not be able to use `make` command.

example/Vagrantfile
```diff
@@ -0,0 +1,30 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+Vagrant.configure("2") do |config|
+ config.vm.box = "ubuntu/xenial64"
```
zoetrope, line 5:
```suggestion
 config.vm.box = "ubuntu/bionic64"
```

I prefer to use bionic instead of xenial.

example/Vagrantfile
```diff
@@ -0,0 +1,30 @@
...
+    sudo apt-get update && sudo apt-get install -y apt-transport-https curl ca-certificates curl gnupg-agent software-properties-common make gcc
+    curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+    sudo add-apt-repository \
+      "deb [arch=amd64] https://download.docker.com/linux/ubuntu \
+      $(lsb_release -cs) \
+      stable"
+    curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
+    sudo echo "deb https://apt.kubernetes.io/ kubernetes-xenial main" > /etc/apt/sources.list.d/kubernetes.list
+    sudo apt-get update
+    sudo apt-get install -y docker-ce docker-ce-cli containerd.io
```
zoetrope, line 22: It's easier to use docker provisioner than apt.
You can write the following:

```Vagrantfile
config.vm.provision "docker"
```

example/Vagrantfile
```diff
@@ -0,0 +1,30 @@
...
+    systemctl daemon-reload
+    echo "Allowing docker as a non-root"
+    sudo usermod -aG docker vagrant && newgrp docker
+    wget -q https://dl.google.com/go/go1.13.3.linux-amd64.tar.gz && \
+      sudo tar xzf go1.13.3.linux-amd64.tar.gz -C /usr/local
+    echo "export PATH=$PATH:/usr/local/go/bin:/home/vagrant/go/bin" > /etc/profile.d/gopath.sh
+  SHELL
```
zoetrope, line 29: Please add the following config:
```
config.vm.synced_folder "../", "/vagrant"
```

This allows you to run example and e2e directly in a VM.

example/Vagrantfile
```diff
@@ -0,0 +1,30 @@
...
+
+  config.vm.provision "shell", inline: <<-SHELL
+    sudo apt-get update && sudo apt-get install -y apt-transport-https curl ca-certificates curl gnupg-agent software-properties-common make gcc
+    curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+    sudo add-apt-repository \
+      "deb [arch=amd64] https://download.docker.com/linux/ubuntu \
+      $(lsb_release -cs) \
+      stable"
+    curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
+    sudo echo "deb https://apt.kubernetes.io/ kubernetes-xenial main" > /etc/apt/sources.list.d/kubernetes.list
```
zoetrope, line 20: What do you need this repository for?

example/Vagrantfile
```diff
@@ -0,0 +1,30 @@
...
+    sudo apt-get update
+    sudo apt-get install -y docker-ce docker-ce-cli containerd.io
+    systemctl daemon-reload
+    echo "Allowing docker as a non-root"
+    sudo usermod -aG docker vagrant && newgrp docker
+    wget -q https://dl.google.com/go/go1.13.3.linux-amd64.tar.gz && \
+      sudo tar xzf go1.13.3.linux-amd64.tar.gz -C /usr/local
+    echo "export PATH=$PATH:/usr/local/go/bin:/home/vagrant/go/bin" > /etc/profile.d/gopath.sh
```
zoetrope, line 28: Don't you need permission to write the file?

```
echo "export PATH=$PATH:/usr/local/go/bin:/home/vagrant/go/bin" | sudo tee /etc/profile.d/gopath.sh
```

example/Vagrantfile
```diff
@@ -0,0 +1,30 @@
...
+      stable"
+    curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
+    sudo echo "deb https://apt.kubernetes.io/ kubernetes-xenial main" > /etc/apt/sources.list.d/kubernetes.list
+    sudo apt-get update
+    sudo apt-get install -y docker-ce docker-ce-cli containerd.io
+    systemctl daemon-reload
+    echo "Allowing docker as a non-root"
+    sudo usermod -aG docker vagrant && newgrp docker
+    wget -q https://dl.google.com/go/go1.13.3.linux-amd64.tar.gz && \
+      sudo tar xzf go1.13.3.linux-amd64.tar.gz -C /usr/local
```
zoetrope, line 27: We want to use latest golang, how about using snap?

```
sudo snap install --channel=1.13/stable --classic go
```

example/Vagrantfile
zoetrope: I found that I cannot create a volume group on the synced folder.
I need to increase the disk size of VM and start example there.

I installed vagrant-disksize plugin:
```
vagrant plugin install vagrant-disksize
```

and, increased the disk size:
```
config.disksize.size = '50GB'
```

zoetrope:
Finally, it worked with the following file:

```vagrantfile
# -*- mode: ruby -*-
# vi: set ft=ruby :

Vagrant.configure("2") do |config|
  config.vm.box = "ubuntu/bionic64"
  config.disksize.size = '50GB'

  config.vm.provider "virtualbox" do |vb|
    vb.memory = (1024 * 5).to_s # make sure there's enough memory for all components
    vb.cpus = 2
  end

  config.vm.provision "docker"
  config.vm.synced_folder "../", "/vagrant"
  config.vm.provision "shell", inline: <<-SHELL
    sudo apt-get update && sudo apt-get install -y apt-transport-https curl ca-certificates curl gnupg-agent software-properties-common build-essential jq
    sudo snap install --channel=1.13/stable --classic go
    echo "export PATH=$PATH:/home/vagrant/go/bin" | sudo tee /etc/profile.d/gopath.sh
  SHELL
end
```

example/Vagrantfile
frederiko: For `kubectl`, although I could get the binary directly.

example/Vagrantfile
frederiko: Interesting. Looking at the docs it does install docker (which I did not know). I'd only argue that it can be confusing to users as - at least in my case - I thought you would provision everything as containers in the VM, when in fact you're only using it do install docker. So, in that case, what do you think of consolidating everything to only one provisioner? I just thought it would read better.

example/Vagrantfile
frederiko: Oops.. not need. I just saw that `make setup` does it

example/Vagrantfile
frederiko: I don't think I need, as the provisioner runs as privileged. In fact, even `sudo` does not make sense.

example/Vagrantfile
frederiko: Quite strange, with bionic64, it stopped working for me when running the `example`. I'm curious if you had to change something on `make run` to get the `backing_store` created outside `/vagrant/example`. That's the only way I could get it working with bionic64. On Xenial64, it works as expected.

frederiko: @zoetrope the only way I could it make the `example` to work with `precise64` on my local ubuntu box was by setting the backing_store to be created in the / disk. As you've mentioned, using the shared Vagrant volume does not work (surprisingly worked for me on xenial64). Would you be okay with changing the backing_store to somewhere else, like `/tmp/topo-build/backing_store`, for example?

example/Vagrantfile
zoetrope: Sorry it's my misunderstanding.

example/Vagrantfile
zoetrope: I see. I think it's fine as it is.

example/Vagrantfile
zoetrope: Thank you for investigating! Hmm, that's strange...

I think it's a good idea to create the `backing_store` outside `/vagrant/example`.

zoetrope:
@frederiko 
Thanks!
However, I'm concerned that it may not be possible to `make setup` on `/tmp` in some environments.
So, could you change Makefile to specify BACKING_STORE_DIR variable as follows?

```Makefile
BACKING_STORE_DIR = ./build

start-lvmd:
	mkdir -p build $(BACKING_STORE_DIR)
…
```

In a Vagrant environment, we will be able to run make as follows:

```
make BACKING_STORE_DIR=/tmp
```

frederiko:
In which environments, could you give and example? If we assume that this is a Ubuntu vm using that particular vagrant image, or any Linux node that is required to run lvmd, I can’t see why it wouldn’t work defaulting to /tmp (unless /tmp runs on tmpfs?, idk). In any case, I will address the suggested changes. It should be quick.

zoetrope:
Exactly, I'm worried that `/tmp` runs on tmpfs.

zoetrope (changes requested):
> In any case, I will address the suggested changes. It should be quick.

Thank you for your response.

e2e/Makefile
```diff
@@ -79,30 +80,30 @@ shutdown-kind:
 	for d in $$(mount | grep /lib/kubelet | cut -d ' ' -f 3); do $(SUDO) umount $$d; done
 
 start-lvmd:
-	mkdir -p build
+	mkdir -p build /tmp/build
```
zoetrope, line 83:
```suggestion
	mkdir -p build $(BACKING_STORE)
```

zoetrope (changes requested):

zoetrope (approved):
LGTM

morimoto-cybozu (approved):
LGTM

Issue #195, opened 2020-08-31
lvmd daemonset


kfox1111:
## What
It would be nice if you could deploy lvmd with Kubernetes

## How
lvmd should be deployable thusly:
* shipping lvmd & nsenter in a container
* setting the lvmd pod to hostPID: true (so it can see pid1 on the host)
* setting the lvmd pod to be privileged (so it can use nsenter)
* mounting /bin or some other directory from the host into the container at /hostbin as writable
* copying lvmd to /hostbin
* running: nsenter -t 1 -a /bin/lvmd (startup lvmd on the host)

It should then run lvmd as if it was on the host as all the containerization will be stripped back to the namespaces associated with the hosts pid1 process.

## Checklist

- [x] Finish implementation of the issue
- [ ] Test all functions
- [ ] Write some documents for this change
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

llamerada-jp:
Hi @kfox1111 

I'm very excited about your idea. I think it requires some improvement for the following reasons.
- Some distributions prohibit write to system directories like `/bin`.
- Running `lvmd` as a whole with `nsenter -t 1` completely breaks the container boundary.
  It loses the benefit of container separation, for example, it suffers from a collision of a listening port.

Instead of running `lvmd` in the host namespaces, I'd like to call `lvm` and `blockdev` commands
with `nsenter -t 1 -a` from inside `lvmd`. In this way, we can avoid the above problems.
What do you think about this improvement?

kfox1111:
It breaks the container boundry, but that is often done with really low level containers such as calico where it copies binaries on to the host to make cni work, or some of the host upgrade orchestration containers. Arguably this use case fits that use case. Low level storage access.

I had looked through the code yesterday and noticed it could probably be done by wrapping the exec of the lvm and blockdev commands with nsenter. But I was not sure there wasn't any other functions in the code base accessing things directly that would need to have the same view as the host. It would take good knowledge of the code base to decide if that was safe.

I did get a very rough prototype working and emailed it to @ymmt2005. Not sure what the best way to integrate it with the rest of the code. maybe an overlay

In the prototype I put it in /usr/libexec/lvmd as that is the FHS standard location for those sorts of things. Your right, it may not be writable though. It probably needs to be made configurable. That kind of configurability can be harder to do with kustomize then something like helm. Which is probably a whole other discussion. (The clusterrole names are questionable in my environment...)

[...]

PR #215, opened 2020-10-05
 Build hypertopolvm inside docker

onokatio:
support docker multi-stage build.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+build
+bin
+include
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,0 @@
-*
-!build
-!LICENSE
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -31,7 +31,7 @@ jobs:
       - run: make setup
       - run: make build TOPOLVM_VERSION=${{ steps.check_version.outputs.version }}
       - run: tar czf lvmd-${{ steps.check_version.outputs.version }}.tar.gz -C ./build lvmd
-      - run: make image
+      - run: make image TOPOLVM_VERSION=${{ steps.check_version.outputs.version }}
       - run: make tag push IMAGE_TAG=${{ steps.check_version.outputs.version }}
       - name: Push branch tag
         if: ${{ steps.check_version.outputs.prerelease == 'false' }}
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -29,7 +29,7 @@ jobs:
         with:
           go-version: 1.13
       - run: make setup
-      - run: make build TOPOLVM_VERSION=${{ steps.check_version.outputs.version }}
+      - run: make build/lvmd TOPOLVM_VERSION=${{ steps.check_version.outputs.version }}
       - run: tar czf lvmd-${{ steps.check_version.outputs.version }}.tar.gz -C ./build lvmd
       - run: make image TOPOLVM_VERSION=${{ steps.check_version.outputs.version }}
       - run: make tag push IMAGE_TAG=${{ steps.check_version.outputs.version }}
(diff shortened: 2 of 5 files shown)


toshipp (changes requested):
Thanks for your PR!

I tested the PR in my local environment, `make image` is failed.
We added the dockerignore file and it contains wildcard ignore,
so that `ADD . /workdir` can not add needed files to build.

All you need to do is following.

- remove wildcard ignore from dockerignore
- csi sidecars should be built in the container then copy them to the final container image
- GitHub workflow should be changed
  - `make build` before `make image` should be deleted.
  - specify TOPOLVM_VERSION to `make image` for github release workflow because `make build` is now in a container.
     ref. https://github.com/topolvm/topolvm/blob/master/.github/workflows/release.yaml#L32-L34

Would you revise your PR?

onokatio:
@toshipp Thanks for replying.
OK. I'll do that.

.github/workflows/release.yaml
```diff
@@ -31,7 +31,7 @@ jobs:
       - run: make setup
       - run: make build TOPOLVM_VERSION=${{ steps.check_version.outputs.version }}
```
onokatio, line 32: `lvmd` is stil needed as standalone binary so this line is not removed yet.
I'll make new PullRequest after https://github.com/topolvm/topolvm/issues/195 is implemented

toshipp: Yeah, it is still needed.
Moreover, after #195 is done, someone (including ours) need to migrate to use daemonset,
there is no need to rush to delete lines.

onokatio:
I think now all points are fixed.

Dockerfile
```diff
@@ -1,3 +1,11 @@
+# Build Container
+FROM golang:buster AS build-env
```
toshipp, line 2: we now use golang 1.13, so use `golang:1.13-buster` instead.

Dockerfile
```diff
@@ -1,3 +1,11 @@
...
+
+ADD . /workdir
```
toshipp, line 4: COPY is preferable.
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy

.github/workflows/release.yaml
```diff
@@ -29,9 +29,9 @@ jobs:
         with:
           go-version: 1.13
       - run: make setup
-      - run: make build TOPOLVM_VERSION=${{ steps.check_version.outputs.version }}
+      - run: make build/lvmd TOPOLVM_VERSION=${{ steps.check_version.outputs.version }}
       - run: tar czf lvmd-${{ steps.check_version.outputs.version }}.tar.gz -C ./build lvmd
-      - run: make image
+      - run: make image TOPOLVM_VERSION=${{ steps.check_version.outputs.version }}
```
toshipp, line 34: Specified TOPOLVM_VERSION is not used in the `image` target in Makefile.
To use the variable, you also have to modify Makefile and dockerfile.

toshipp (changes requested):
@onokatio
I put some other comments.

onokatio:
Fixed dockerfile and Makefile.
(Sorry force pushing. I missed committing some debug line)

.dockerignore
```diff
@@ -1,3 +0,0 @@
-*
-!build
```
toshipp, line 2: I missed that ignoring the `build` directory is required because if `make build` runs before `make image`,
the built artifacts are copied into a container, so binaries that should be built in the container are never built.

toshipp (changes requested):

.dockerignore
onokatio: I missed too. Thanks.

toshipp (approved):
LGTM

llamerada-jp (approved):
LGTM

zoetrope (approved):
LGTM

PR #221, opened 2020-10-20
modify manifest for deployment-scheduler

llamerada-jp:
- split `scheduler-potions.yaml` file under `topolvm/deployment/` to set listen at `0.0.0.0` for Deployment.
- split manifest file under `topolvm/e2e` for test.
- run e2e test both of for daemonset and for deployment.


toshipp (approved):
LGTM

e2e/scheduler_test.go
```diff
@@ -28,21 +26,26 @@ func testScheduler() {
...
 
-			pod := podlist.Items[0]
-			for _, cond := range pod.Status.Conditions {
-				fmt.Println(cond)
-				if cond.Type == corev1.PodReady && cond.Status == corev1.ConditionTrue {
-					return nil
+			for _, pod := range podlist.Items {
+				podReady := false
+				for _, cond := range pod.Status.Conditions {
+					fmt.Println(cond)
```
zoetrope, line 36: for debug?

.github/workflows/main.yaml
```diff
@@ -24,8 +24,10 @@ jobs:
     strategy:
       matrix:
         test_kubernetes_target: [current, prev, prev2]
+        test_scheduler_manifest: [daemonset, deployment]
```
zoetrope, line 27: Do you need to run the test on all combinations?
I feel it is too much.

e2e/scheduler_test.go
llamerada-jp: I think so, this line was already in place before the change.
And changed it to use `GinkgoWriter`.

.github/workflows/main.yaml
llamerada-jp: We will revise them if there are any problems.

zoetrope (approved):

toshipp (approved):

PR #230, opened 2020-12-08
Improve stale LV deletion method

### LINKED ISSUE #178: Improve the deletion method of stale LogicalVolumes
## What

In the current implementation, if logical volumes cannot be deleted for 24 hours (adjustable in `stale-period` option), topolvm-controller gives up the deletion of the volumes and delete the related LogicalVolumes CR.

https://github.com/topolvm/topolvm/blob/v0.5.2/docs/topolvm-controller.md#delete-stale-logicalvolumes

This method may create in the orphaned logical volumes that are not tied to LogicalVolume CRs.
So we need to improve the deletion method to avoid creating the orphaned logical volumes.

## How

- Remove `stale-period` option and related logic

## Checklist

- [ ] Test all functions

bells17:
resolve #178 

- [x] remove LogicalVolumeCleanupReconciler
- [x] delete LogicalVolume resources connected to the deleting node when deleting a node


masa213f (commented):
`--cleanup-interval` option may be unnecessary too.
Could you check it? And please remove the option if it is unnecessary.

controllers/node_controller.go
```diff
@@ -123,11 +125,60 @@ func (r *NodeReconciler) doFinalize(ctx context.Context, log logr.Logger, node *
 			return ctrl.Result{}, err
 		}
 		log.Info("deleted PVC", "name", pvc.Name, "namespace", pvc.Namespace)
+
+		r.Get(ctx, client.ObjectKey{Name: pvc.Spec.VolumeName}, &lv)
```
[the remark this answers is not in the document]
masa213f: > We can use spec.nodeName in LogicalVolumeSpec as a filter.

I think so. If there are orphaned LogicalVolume CRs, they cannot be deleted in this logic.
It is better to list all LogicalVolume CRs and filter by .spec.nodeName.

e2e/cleanup_test.go
```diff
@@ -276,4 +274,14 @@ spec:
 		stdout, stderr, err := kubectl("delete", "ns", cleanupTest)
 		Expect(err).ShouldNot(HaveOccurred(), "stdout=%s, stderr=%s", stdout, stderr)
 	})
+
+	It("should delete unnecessary LVs", func() {
```
toshipp, line 278: I think this is not a test but a cleanup logic.
If so, use [AfterSuite](https://onsi.github.io/ginkgo/#global-setup-and-teardown-beforesuite-and-aftersuite) or something instead.

bells17:
@llamerada-jp @toshipp @masa213f I fixed some problems. Would you please review this PR again?

controllers/node_controller.go
```diff
@@ -123,11 +124,64 @@ func (r *NodeReconciler) doFinalize(ctx context.Context, log logr.Logger, node *
 			return ctrl.Result{}, err
 		}
 		log.Info("deleted PVC", "name", pvc.Name, "namespace", pvc.Namespace)
+
```
masa213f, line 127: This is unnecessary blank lines.

controllers/node_controller.go
```diff
@@ -123,11 +124,64 @@ func (r *NodeReconciler) doFinalize(ctx context.Context, log logr.Logger, node *
...
+	}
+
+	lvList := new(topolvmv1.LogicalVolumeList)
+	err = r.List(ctx, lvList, client.MatchingFields{KeyLogicalVolumeNode: node.Name})
+	if err != nil {
+		log.Error(err, "failed to get LogicalVolumes")
```
masa213f, line 133: In my opinion, it would be better to use the same error messages as in other parts.

```suggestion
		log.Error(err, "unable to fetch LogicalVolumeList")
```

controllers/node_controller.go
```diff
@@ -123,11 +124,64 @@ func (r *NodeReconciler) doFinalize(ctx context.Context, log logr.Logger, node *
...
+		return ctrl.Result{}, err
+	}
+
+	for _, lv := range lvList.Items {
+		err = r.cleanupLogicalVolume(ctx, log, &lv)
+		if err != nil {
+			log.Error(err, "unable to cleanup LogicalVolume")
```
masa213f, line 140: This error message is redundant. Because error messages are logged in the function.

masa213f (changes requested):

controllers/node_controller.go
masa213f: I'm sorry. In other places, it's more likely to start with `failed to ~`. You don't have to fix it.

docs/topolvm-controller.md
```diff
@@ -158,7 +158,7 @@ Controllers
 
 `topolvm-metrics` adds `topolvm.cybozu.com/node` finalizer.
 
-When a Node is being deleted, the controller deletes all PVCs for TopoLVM
+When a Node is being deleted, the controller deletes all PVCs and Logical Volumes for TopoLVM
```
toshipp, line 161: Space confuses me (but the naming is actually confusing...)
Would you use `LogicalVolumes`?

toshipp (changes requested):

e2e/cleanup_test.go
```diff
@@ -276,4 +274,14 @@ spec:
 		stdout, stderr, err := kubectl("delete", "ns", cleanupTest)
 		Expect(err).ShouldNot(HaveOccurred(), "stdout=%s, stderr=%s", stdout, stderr)
 	})
+
+	AfterSuite(func() {
```
toshipp, line 278: Sorry, my suggestion was a mistake.
`AfterSuite` is a global teardown, so it should be used at the top level.
In this case, I think calling umount and lvremove directly is sufficient.

toshipp: Fmm, it seems each `It` is independent.
https://github.com/onsi/ginkgo/issues/70#issuecomment-42882976
So, the current test is broken...?

docs/topolvm-controller.md
bells17: fixed it.

controllers/node_controller.go
bells17: fixed it.

controllers/node_controller.go
bells17: fixed it.

controllers/node_controller.go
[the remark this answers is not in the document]
bells17: fixed it.

e2e/cleanup_test.go
toshipp: By Ginkgo default, `It` runs on sequence order.
@bells17 I think the first implementation is better, would you remove this [commit](https://github.com/topolvm/topolvm/pull/230/commits/2bf86f4aec7eba7d1c58c95b9e5d2a4403af1105)?

masa213f (dismissed):
Thank you!

e2e/cleanup_test.go
bells17: fixed it.

bells17:
@llamerada-jp @toshipp @masa213f I fixed some problems. Would you please review this PR again?

controllers/constants.go
```diff
@@ -1,9 +1,12 @@
 package controllers
 
 const (
-	// KeySelectedNode is an indexing key for the controller
+	// KeySelectedNode is a PVC resource indexing key for the controller
 	KeySelectedNode = "metadata.annotations.selected-node"
 
+	// KeyLogicalVolumeNode is a Logical Volume resource indexing key for the controller
+	KeyLogicalVolumeNode = "spec.nodeName"
```
llamerada-jp, line 8: Would you consider making `KeySelectedNode` and `KeyLogicalVolumeNode` package private, since it is not referenced by other packages.

e2e/cleanup_test.go
```diff
@@ -276,4 +274,14 @@ spec:
 		stdout, stderr, err := kubectl("delete", "ns", cleanupTest)
 		Expect(err).ShouldNot(HaveOccurred(), "stdout=%s, stderr=%s", stdout, stderr)
 	})
+
+	It("cleanup", func() {
```
llamerada-jp, line 278:
```suggestion
	It("should cleanup volumes", func() {
```

bells17:
@llamerada-jp I updated. Could you review it again?

e2e/cleanup_test.go
toshipp: @llamerada-jp
This is a not test, but cleanup logic.
I think using `Should` is confusing.

toshipp: Since this is a cleanup process, we should use `AfterEach` rather than `It`.
However, if we use `AfterEach`, the cleanup will be run by each `It`.
So, we use `It` for cleanup, although it is a strange way to use `It`.

toshipp: I discussed this with my teams, we concluded to use the same message as the others to avoid confusion.

toshipp (approved):
LGTM

llamerada-jp (approved):
LGTM

PR #233, opened 2020-12-21
Use MayRunAs instead of MustRunAs to avoid recursive chown

tapih:
Related to https://github.com/cybozu-go/neco/issues/1018
Signed-off-by: H.Muraoka <hiroshi-muraoka@cybozu.co.jp>


zoetrope (approved):

toshipp (approved):
LGTM

PR #256, opened 2021-01-30
Update the CSI spec to v1.3.0

bells17:
Hi, I updated the CSI spec to v1.3.0.
(this compare is this: https://github.com/container-storage-interface/spec/compare/release-1.1...release-1.3)

This update is needed to #247 .
When uses k8s.io/mount-utils or k8s.io/utils/exec package, we need to use CSI spec to v1.2.0 or later.

Currently, tololvm use below sidecars version:

- external-provisioner: 2.1.0
- external-attacher: 3.1.0
- external-resizer: 1.1.0

these sidecars use the CSI spec v1.3.0 package.

- https://github.com/kubernetes-csi/external-provisioner/blob/v2.1.0/go.mod
- https://github.com/kubernetes-csi/external-attacher/blob/v3.1.0/go.mod
- https://github.com/kubernetes-csi/external-resizer/blob/v1.1.0/go.mod

So, this PR updated the CSI spec to v1.3.0, not v1.2.0.

But, on the other hand, the Kubernetes versions that supported by topolvm now is below:

- v1.19
- v1.18
- v1.17

these versions use the CSI spec v1.2.0.

- https://github.com/kubernetes/kubernetes/blob/v1.19.7/go.mod
- https://github.com/kubernetes/kubernetes/blob/v1.18.15/go.mod
- https://github.com/kubernetes/kubernetes/blob/v1.17.17/go.mod

But I think the updated contents of Node Service in the CSI spec is only adding a few fields.
So I think we don't have any problem using the CSI spec v1.3.0 unless we use newly added fields.

And the CSI sidecars already use the CSI spec v1.3.0, so I think we don't need to update the CSI sidecars version.


ymmt2005:
@bells17 
TopoLVM generates CSI protobuf code by itself with `Makefile`.

The dependency you are updating comes from a mistake in `e2e/publish_test.go`.

```console
$ go mod why github.com/container-storage-interface/spec/...
# github.com/container-storage-interface/spec/lib/go/csi
github.com/topolvm/topolvm/e2e
github.com/topolvm/topolvm/e2e.test
github.com/container-storage-interface/spec/lib/go/csi

$ git grep 'spec/lib/go/csi' e2e
e2e/publish_test.go:    "github.com/container-storage-interface/spec/lib/go/csi"
```

Please fix `e2e/publish_test.go` to reference `github.com/topolvm/topolvm/csi` package properly,
and regenerate protobuf code in `csi` by updating `Makefile`.

go.mod
```diff
@@ -6,7 +6,7 @@ replace launchpad.net/gocheck => github.com/go-check/check v0.0.0-20180628173108
 
 require (
 	cloud.google.com/go v0.63.0 // indirect
-	github.com/container-storage-interface/spec v1.1.0
+	github.com/container-storage-interface/spec v1.3.0 // indirect
```
bells17, line 9: This was added when I run "make test".
(not e2e)

bells17:
Thank you @ymmt2005 . I updated `CSI_VERSION` on `Makefile` and run "rm csi.proto; make generate".
And I fixed to use `github.com/topolvm/topolvm/csi` in `e2e/publish_test.go`.

go.mod
ymmt2005: That's because we are using a very old `csi-test` package.
We should update it to v4 at the same time.

Here is my suggestion:
```diff
diff --git a/e2e/suite_test.go b/e2e/suite_test.go
index e6c8295..bb6c3f2 100644
--- a/e2e/suite_test.go
+++ b/e2e/suite_test.go
@@ -9,7 +9,7 @@ import (
        "testing"
        "time"

-       "github.com/kubernetes-csi/csi-test/pkg/sanity"
+       "github.com/kubernetes-csi/csi-test/v4/pkg/sanity"
        . "github.com/onsi/ginkgo"
        . "github.com/onsi/gomega"
        appsv1 "k8s.io/api/apps/v1"
@@ -134,13 +134,13 @@ var _ = Describe("TopoLVM", func() {
                        }).Should(Succeed())
                })

-               sanity.GinkgoTest(&sanity.Config{
-                       Address:           "/tmp/topolvm/worker1/plugins/topolvm.cybozu.com/node/csi-topolvm.sock",
-                       ControllerAddress: "/tmp/topolvm/worker1/plugins/topolvm.cybozu.com/controller/csi-topolvm.sock",
-                       TargetPath:        "/tmp/topolvm/worker1/plugins/topolvm.cybozu.com/node/mountdir",
-                       StagingPath:       "/tmp/topolvm/worker1/plugins/topolvm.cybozu.com/node/stagingdir",
-                       TestVolumeSize:    1073741824,
-                       IDGen:             &sanity.DefaultIDGenerator{},
-               })
+        tc := sanity.NewTestConfig()
+        tc.Address = "/tmp/topolvm/worker1/plugins/topolvm.cybozu.com/node/csi-topolvm.sock"
+        tc.ControllerAddress = "/tmp/topolvm/worker1/plugins/topolvm.cybozu.com/controller/csi-topolvm.sock"
+        tc.TargetPath = "/tmp/topolvm/worker1/plugins/topolvm.cybozu.com/node/mountdir"
+        tc.StagingPath = "/tmp/topolvm/worker1/plugins/topolvm.cybozu.com/node/stagingdir"
+        tc.TestVolumeSize = 1073741824
+        tc.IDGen = &sanity.DefaultIDGenerator{}
+               sanity.GinkgoTest(&tc)
        })
 })
diff --git a/go.mod b/go.mod
index 35c344e..26bfeee 100644
--- a/go.mod
+++ b/go.mod
@@ -6,13 +6,12 @@ replace launchpad.net/gocheck => github.com/go-check/check v0.0.0-20180628173108

 require (
        cloud.google.com/go v0.63.0 // indirect
-       github.com/container-storage-interface/spec v1.3.0 // indirect
        github.com/cybozu-go/log v1.5.0
        github.com/cybozu-go/well v1.10.0
        github.com/go-logr/logr v0.1.0
        github.com/golang/protobuf v1.4.2
        github.com/google/go-cmp v0.5.2
-       github.com/kubernetes-csi/csi-test v2.2.0+incompatible
+       github.com/kubernetes-csi/csi-test/v4 v4.0.2
        github.com/onsi/ginkgo v1.14.1
        github.com/onsi/gomega v1.10.2
        github.com/prometheus/client_golang v1.1.0
```

bells17: Thank you. I updated csi-test package version.

e2e/suite_test.go
```diff
@@ -134,13 +134,14 @@ var _ = Describe("TopoLVM", func() {
...
-			IDGen:             &sanity.DefaultIDGenerator{},
-		})
+		tc := sanity.NewTestConfig()
+		tc.Address = "/tmp/topolvm/worker1/plugins/topolvm.cybozu.com/node/csi-topolvm.sock"
+		tc.ControllerAddress = "/tmp/topolvm/worker1/plugins/topolvm.cybozu.com/controller/csi-topolvm.sock"
+		tc.TargetPath = "/tmp/topolvm/worker1/plugins/topolvm.cybozu.com/node/mountdir"
+		tc.StagingPath = "/tmp/topolvm/worker1/plugins/topolvm.cybozu.com/node/stagingdir"
+		tc.TestVolumeSize = 1073741824
+		tc.IDGen = &sanity.DefaultIDGenerator{}
+		tc.IdempotentCount = 0
```
ymmt2005, line 144: This is to skip idempotency tests.
I think we should make them passed.
https://pkg.go.dev/github.com/kubernetes-csi/csi-test/v4/pkg/sanity#TestConfig

toshipp: Agreed.
I inspected a failed test and found an error message that is `LogicalVolume.topolvm.cybozu.com "sanity-node-full-f497324e-e3ef8e64" not found`.
https://github.com/topolvm/topolvm/runs/1802349235
It seems that CreateVolume is failed or getting created LogicalVolume is failed.

toshipp: I'm not sure that `LogicalVolume.topolvm.cybozu.com` is correct.
In CRD, I can not find it, but found `plural: logicalvolumes`.

driver/k8s/logicalvolume_service.go
```diff
@@ -140,7 +140,28 @@ func (s *LogicalVolumeService) DeleteVolume(ctx context.Context, volumeID string
...
+			return ctx.Err()
+		case <-time.After(100 * time.Millisecond):
+		}
+
+		err := s.Get(ctx, client.ObjectKey{Name: lv.Name}, new(topolvmv1.LogicalVolume))
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				return nil
+			}
+			logger.Error(err, "failed to get LogicalVolume", "name", lv.Name)
```
llamerada-jp, line 162: I think it is better to return an error immediately when getting the error without IsNotFound.

driver/k8s/logicalvolume_service.go
```diff
@@ -140,7 +140,28 @@ func (s *LogicalVolumeService) DeleteVolume(ctx context.Context, volumeID string
 		return err
 	}
 
-	return s.Delete(ctx, lv)
+	err = s.Delete(ctx, lv)
+	if err != nil {
```
ymmt2005, line 144: If err indicates `lv` is not found, we should return nil.

driver/k8s/logicalvolume_service.go
ymmt2005: @llamerada-jp Why?  For me, this seems legit.

ymmt2005 (changes requested):

driver/k8s/logicalvolume_service.go
toshipp: @ymmt2005 
I doubt that the client will retry when the connection error has happened, but give up if it is reached to max retry count.
Is it needed to distinguish between not found and other errors?

toshipp: Ah, you already say that is necessary, sorry.

driver/k8s/logicalvolume_service.go
ymmt2005, line 162:
```suggestion
			logger.Error(err, "failed to get LogicalVolume", "name", lv.Name)
			return err
```

bells17:
Thank you, folks.
I applied the suggestion commit.
Could you review this again?

ymmt2005:
@bells17 Please fix this too.
https://github.com/topolvm/topolvm/pull/256/files#r570101601

driver/k8s/logicalvolume_service.go
bells17: Um... It seems already modified to return nil when if err indicates `lv` is not found.
Did I misunderstand something?

https://github.com/topolvm/topolvm/pull/256/files#diff-78915383711edcc7723926712f85a2eb106b00aeb79b852c37d8560c16d7a765R160

ymmt2005 (approved):
LGTM

driver/k8s/logicalvolume_service.go
bells17: Sorry, I noticed my misunderstanding.
I fixed it.

toshipp (approved):

Issue #96, opened 2020-02-05
Support mountOptions


ymmt2005:
## What

Design and implement  [`mountOptions`](https://kubernetes.io/docs/concepts/storage/storage-classes/#mount-options) property of StorageClass that is ignored in the current implementation.

Resolves #94.

## How

The design should consider the following:

- How to distinguish flags for [`mount` system call](http://man7.org/linux/man-pages/man2/mount.2.html) and other options (`data`).
- CSI passes the read-only volume attribute independently from `mountOptions`.
    Should `ro` or `rw` mount option be ignored?
- What flags/options should be specified by default.
- How to support filesystem-specific options.  For example, XFS has `wsync` option.

## Checklist

- [ ] Finish implentation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

PR #260, opened 2021-02-14
Support mount option

bells17:
resolve: #96
old PR is here: https://github.com/topolvm/topolvm/pull/247

---

Hi, This PR adds support mount options using kubernetes/mount-util package.

> How to distinguish flags for mount system call and other options (data).

There is no way to distinguish flags for mount system call and other options (data) because kubernetes/mount-util package doesn't distinguish flags.

> CSI passes the read-only volume attribute independently from mountOptions. Should ro or rw mount option be ignored?

I added the validation case for mount option \"rw\" is specified even though the read-only mode is specified.

> What flags/options should be specified by default.

We don't need to specify some options by default manually because default options are set up from kubernetes/mount-util package.

> How to support filesystem-specific options. For example, XFS has wsync option.

Supporting filesystem-specific options can set in "mountOptions" field of StorageClass resource.


driver/node.go
```diff
@@ -485,24 +524,23 @@ func (s *nodeService) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
...
 		return nil, status.Errorf(codes.Internal, "filesystem %s is not mounted at %s", vid, vpath)
 	}
 
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
-	err = fs.Resize(vpath)
-	if err != nil {
+	r := filesystem.NewResizeFs(&s.mounter)
+	if _, err := r.Resize(device, req.GetVolumePath()); err != nil {
```
llamerada-jp, line 543:
```suggestion
	if _, err := r.Resize(device, vpath); err != nil {
```

filesystem/resizefs.go
```diff
@@ -0,0 +1,73 @@
+/*
```
toshipp, line 1: This is copied from k8s repository and it is licensed under apache license.
Would you copy the [license file](https://github.com/kubernetes/kubernetes/blob/v1.19.7/LICENSES/LICENSE) into root directory?

e2e/publish_test.go
```diff
@@ -279,4 +279,72 @@ spec:
...
+		}).Should(Succeed())
+
+		cl.register(volumeID, mountTargetPath)
+
+		By("mount option \"rw\" is specified even though read only mode is specified")
+		mountVolCap := &csi.VolumeCapability{
+			AccessType: &csi.VolumeCapability_Mount{
+				Mount: &csi.VolumeCapability_MountVolume{
+					FsType:     "xfs",
+					MountFlags: []string{"wr"},
```
toshipp, line 328: Is this typo for `rw`?

toshipp (changes requested):
Would you add a e2e test case for specifying mount option?

driver/node.go
```diff
@@ -172,25 +181,55 @@ func (s *nodeService) nodePublishFilesystemVolume(req *csi.NodePublishVolumeRequ
...
-			return nil, status.Errorf(codes.Internal, "failed to create filesystem: volume=%s, error=%v", req.GetVolumeId(), err)
+
+	for _, f := range mountOption.MountFlags {
+		hasOption := false
+		for _, o := range mountOptions {
+			if o == f {
+				hasOption = true
+				break
+			}
+		}
```
llamerada-jp, line 196: In the mount command, if conflicting options are specified, the option specified later becomes effective.
There is no need to remove duplicate options.

satoru-takeuchi:
I don't have any addditional comment for now.

satoru-takeuchi (commented):
test

e2e/publish_test.go
```diff
@@ -279,4 +281,164 @@ spec:
...
+		stdout, stderr, err := kubectl("get", "pvc", "topo-pvc-mount-option", "-o", "yaml")
+		Expect(err).ShouldNot(HaveOccurred(), "stdout=%s, stderr=%s", stdout, stderr)
+
+		var pvc corev1.PersistentVolumeClaim
+		err = yaml.Unmarshal(stdout, &pvc)
+		Expect(err).ShouldNot(HaveOccurred())
+
+		fmt.Printf("pvc stdout: %+v\n", string(stdout))
+		fmt.Printf("pvc: %+v\n", pvc)
+		fmt.Printf("pvc.Spec.ClaimRef: %+v\n", pvc.Spec.VolumeName)
```
toshipp, line 351: Is this code for temporary debugging?
If for logging, use [ginkgo.writer](https://pkg.go.dev/github.com/onsi/ginkgo@v1.15.0/internal/writer) for better logging.

toshipp (changes requested):

llamerada-jp (changes requested):

e2e/publish_test.go
bells17: Thank you. I fixed it.

driver/node.go
```diff
@@ -338,7 +366,7 @@ func (s *nodeService) isEphemeralVolume(volume *proto.LogicalVolume) bool {
 
 func (s *nodeService) nodeUnpublishFilesystemVolume(req *csi.NodeUnpublishVolumeRequest, device string) (*csi.NodeUnpublishVolumeResponse, error) {
 	target := req.GetTargetPath()
-	if err := filesystem.Unmount(device, target); err != nil {
+	if err := s.mounter.Unmount(target); err != nil {
```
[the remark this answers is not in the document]
bells17: Thank you. I fixed it.
https://github.com/topolvm/topolvm/pull/260/commits/485c525f8b2bea4dee41b4b51cb92091bfa48bed

toshipp (dismissed):
LGTM

toshipp:
Oops, I found some docs describe mount options are not supported.
https://github.com/topolvm/topolvm/blob/main/docs/limitations.md#storageclass-mountoptions
https://github.com/topolvm/topolvm/blob/main/docs/user-manual.md#storageclass

@bells17 Would you remove the statements on docs?

driver/node.go
```diff
@@ -338,14 +366,22 @@ func (s *nodeService) isEphemeralVolume(volume *proto.LogicalVolume) bool {
...
+		return nil, status.Errorf(codes.Internal, "mount check failed: target=%s, error=%v", target, err)
 	}
-	if err := os.Remove(device); err != nil {
-		return nil, status.Errorf(codes.Internal, "remove device failed for %s: error=%v", device, err)
+
+	if mounted {
+		if err := s.mounter.Unmount(target); err != nil {
+			return nil, status.Errorf(codes.Internal, "unmount failed for %s: error=%v", target, err)
+		}
+		if err := os.RemoveAll(target); err != nil {
```
llamerada-jp, line 379: `RemoveAll` and `Remove` should be executed regardless of the `mountd` to recover removing the device file.
(`RemoveAll` and `Remove` do not need to check the target file because it returns nil when the target does not exist.)

llamerada-jp (changes requested):

bells17:
@toshipp @llamerada-jp Hi, I've fixed the points you pointed out. Would you confirm it?

driver/node.go
```diff
@@ -376,12 +376,13 @@ func (s *nodeService) nodeUnpublishFilesystemVolume(req *csi.NodeUnpublishVolume
...
-		}
-		if err := os.Remove(device); err != nil {
-			return nil, status.Errorf(codes.Internal, "remove device failed for %s: error=%v", device, err)
-		}
+	}
+
+	if err := os.RemoveAll(target); err != nil {
+		return nil, status.Errorf(codes.Internal, "remove dir failed for %s: error=%v", target, err)
+	}
+	if err := os.Remove(device); err != nil {
```
toshipp: It seems that `os.Remove` returns not found error but `os.RemoveAll` does not.
So please check if the err is not found.

toshipp (changes requested):

driver/node.go
toshipp: My test code:
```go
package main
import "os"
import "log"
func main() {
        e := os.Remove("not_found")
        if e != nil {
                log.Print("remove error:", e)
        }
        e = os.RemoveAll("not_found")
        if e != nil {
                log.Print("removeall error:", e)
        }
}
```
The result:
```
2021/03/03 02:30:28 remove error:remove not_found: no such file or directory
```

bells17: @toshipp Sorry...should I fix it like following?

```go
err = os.Remove(device)
if !os.IsNotExist(err) {
	return nil, status.Errorf(codes.Internal, "remove device failed for %s: error=%v", device, err)
}
```

ref: https://play.golang.org/p/XB9fTyhAgGe

toshipp: Yes.

bells17: @toshipp Thank you. I fixed it.

toshipp (approved):
LGTM

llamerada-jp (approved):
LGTM

PR #279, opened 2021-03-19
Make event handler be invoked on create event.

### LINKED ISSUE #278: PVC can remain in Terminating state due to Topolvm PVC finalizer
**Describe the bug**
PVC with the TopoLVM finalizer `topolvm.cybozu.com/pvc` can remain in the Terminating state.
It may be because that the TopoLVM PVC controller can miss the Delete event.

**Environments**
- Version: v0.8.0
- Kubernetes version: 1.19.7

**To Reproduce**
1. Prepare a StatefulSet with a generic ephemeral volume
1. Stop `kube-controller-manager`
2. Delete all Pod (including all TopoLVM components) with `--force` option
3. Start `kube-controller-manager`
4. Confirm the state:
    - The PVC of the generic ephemeral volume PVC is `Terminating` and only has the finalized `topolvm.cybozu.com/pvc`, however, the TopoLVM PVC controller doesn't try to delete it.

**Expected behavior**
The TopoLVM PVC controller should delete it.

**Additional context**
This bug also can reproduce with neco-apps reboot test.
https://github.com/cybozu-go/neco-apps/blob/main/test/reboot_test.go#L84

toshipp:
If pvc or node resources are created before topolvm-controller is started,
controller-runtime notifies it as a create event, so the predicate
function should returns true.

Signed-off-by: Toshikuni Fukaya <toshikuni-fukaya@cybozu.co.jp>

Fixes #278

### DIFF
--- a/controllers/node_controller.go
+++ b/controllers/node_controller.go
@@ -196,7 +196,7 @@ func (r *NodeReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	}
 
 	pred := predicate.Funcs{
-		CreateFunc:  func(event.CreateEvent) bool { return false },
+		CreateFunc:  func(event.CreateEvent) bool { return true },
 		DeleteFunc:  func(event.DeleteEvent) bool { return false },
 		UpdateFunc:  func(event.UpdateEvent) bool { return true },
 		GenericFunc: func(event.GenericEvent) bool { return false },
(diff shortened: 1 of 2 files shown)


llamerada-jp (approved):
LGTM

ymmt2005 (approved):
LGTM

PR #314, opened 2021-05-12
Add better error message when configured VG is not found

toelke:
I was under the impression that I could have different names for the VG on different nodes; this is the error message that would have helped my find the error quicker.

### DIFF
--- a/pkg/lvmd/cmd/root.go
+++ b/pkg/lvmd/cmd/root.go
@@ -79,6 +79,9 @@ func subMain() error {
 	for _, dc := range config.DeviceClasses {
 		_, err := command.FindVolumeGroup(dc.VolumeGroup)
 		if err != nil {
+			log.Error("Volume group not found:", map[string]interface{}{
+				"volume_group": dc.VolumeGroup,
+			})
 			return err
 		}
 	}


pkg/lvmd/cmd/root.go
```diff
@@ -79,6 +79,7 @@ func subMain() error {
 	for _, dc := range config.DeviceClasses {
 		_, err := command.FindVolumeGroup(dc.VolumeGroup)
 		if err != nil {
+			log.Error("Volume group not found:", dc.VolumeGroup);
```
toshipp, line 82: The last semicolon causes lint error, would you remove this?

I found another issue that the build is failed.
```
pkg/lvmd/cmd/root.go:82:43: cannot use dc.VolumeGroup (type string) as type map[string]interface {} in argument to "github.com/cybozu-go/log".
```

`log.Error` requires map but you specified string.
Could you revise this and test locally?

toelke:
Sorry, I did not have a go-toolchain handy.

toshipp (approved):
LGTM

satoru-takeuchi (approved):

PR #326, opened 2021-07-05
Improve makefile of example folder

llamerada-jp:
- use `:=` instead of `=`
- add TMPDIR as a variable
- add some parameters for the test of pvc-autoresizer

### DIFF
--- a/example/Makefile
+++ b/example/Makefile
@@ -1,21 +1,22 @@
 
-KUBERNETES_VERSION=1.21.1
-KUSTOMIZE_VERSION=3.8.9
-TOPOLVM_VERSION=0.8.3
+KUBERNETES_VERSION := 1.21.1
+KUSTOMIZE_VERSION := 3.8.9
+TOPOLVM_VERSION := 0.8.3
 
-KIND_CLUSTER_NAME=topolvm-example
+KIND_CLUSTER_NAME := topolvm-example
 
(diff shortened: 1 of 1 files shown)


example/Makefile
```diff
@@ -1,21 +1,22 @@
...
+KIND_CLUSTER_NAME ?= topolvm-example
 
-SUDO=sudo
-BINDIR=$(shell pwd)/bin
-KUSTOMIZE=$(BINDIR)/kustomize
-KUBECTL=$(BINDIR)/kubectl
-KIND=../bin/kind
+SUDO := sudo
+BINDIR := $(shell pwd)/bin
+TMPDIR := /tmp/topolvm
```
toshipp, line 10: Do you need to use `?=`?

satoru-takeuchi (approved):
No other comment in addition to Toshi's one.

example/Makefile
llamerada-jp: I confirmed that no problem caused using `:=` instead of `?=` in this Makefile. I replaced all `?=` to `:=`.

toshipp: The behavior is described here:  https://www.gnu.org/software/make/manual/make.html#Overriding

toshipp (approved):
LGTM

satoru-takeuchi (approved):

PR #345, opened 2021-08-16
Support ReadWriteOncePod

bells17:


.github/workflows/e2e-k8s.yaml
```diff
@@ -0,0 +1,27 @@
...
+      TEST_SCHEDULER_MANIFEST: ${{ matrix.test_scheduler_manifest }}
+      STORAGE_CAPACITY: ${{ matrix.storage_capacity }}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: 1.16
+      - run: make -C e2e setup
+      - run: make -C e2e start-lvmd
+      - run: make -C e2e test
```
toshipp, line 27: Would you add newline?

charts/topolvm/Chart.yaml
```diff
@@ -4,7 +4,7 @@ type: application
 home: https://github.com/topolvm/topolvm
 name: topolvm
 description: Topolvm
-version: 2.0.2
+version: 2.0.3
```
toshipp, line 7: Because files under charts are not updated, it is not needed, IMO.

toshipp (changes requested):
I reviewed it roughly and commented.

go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/cybozu-go/well v1.10.0
 	github.com/go-logr/logr v0.4.0
 	github.com/google/go-cmp v0.5.6
-	github.com/kubernetes-csi/csi-test/v4 v4.2.0
+	github.com/kubernetes-csi/csi-test/v4 v4.2.1-0.20210804222850-c5c0c94cd86e
```
bells17, line 10: csi-test is not support RWOP yet completely.
https://kubernetes.slack.com/archives/C09QZFCE5/p1628699178102700?thread_ts=1628698823.102500&cid=C09QZFCE5

toshipp:
@bells17 
Recently, CSI sidecars used by TopoLVM are updated to support CSI 1.5 and `SINGLE_NODE_*_WRITER`.
Would you update them? Their versions are described in https://github.com/topolvm/topolvm/blob/main/csi-sidecars.mk

bells17:
The changelogs that updated CSI sidecars to support CSI 1.5 and `SINGLE_NODE_*_WRITER` is here.

- https://github.com/kubernetes-csi/external-provisioner/blob/release-3.0/CHANGELOG/CHANGELOG-3.0.md
- https://github.com/kubernetes-csi/external-resizer/blob/release-1.3/CHANGELOG/CHANGELOG-1.3.md
- https://github.com/kubernetes-csi/livenessprobe/blob/master/CHANGELOG/CHANGELOG-2.4.md
- https://github.com/kubernetes-csi/node-driver-registrar/blob/release-2.3/CHANGELOG/CHANGELOG-2.3.md

llamerada-jp (commented):
Would you update the supported Kubernetes version and CSI version on README.md?

toshipp:
@bells17 
In CSI spec 1.5.0, if two NodePublishVolume for the same volume_id but for different target paths are called, the plugin SHOULD return FAILED_PRECONDITION.
https://github.com/container-storage-interface/spec/blob/v1.5.0/spec.md#nodepublishvolume

There is no such implementation, do you have any attempt to implement it?

toshipp:
@bells17 
I think checking nodePublisVolume twice is specified in CSI spec, so if we state the functionality is supported in a capability, we should do that as the spec says.

But, as you say, topolvm only works with kubernetes, so it can depend on k8s implementation.
The KEP says if SINGLE_NODE_MULTI_WRTER is not supported, k8s uses SINGLE_NODE_WRITER instead of SINGLE_NODE_SINGLE_WRITER.
So I assume ReadWriteOncePod can be used even when topolvm does not support the functionality.
>For ReadWriteOncePod, if the CSI driver supports the SINGLE_NODE_MULTI_WRTER capability, then ReadWriteOncePod will map to SINGLE_NODE_SINGLE_WRITER. If it does not, then ReadWriteOncePod will map to SINGLE_NODE_WRITER. This mapping is chosen because we can safely rely on Kubernetes to enforce the access mode outside of the CSI driver. It also has the advantage of enabling existing CSI drivers to start using ReadWriteOncePod.

toshipp:
@bells17 
Sorry, I missed controller.go. Do we need to change the file as well?

bells17:
I’m sorry too. I fixed it.

PR #347, opened 2021-08-18
Make allowedHostPaths property of node's PSP configurable

debackerl:
Hello,

While installing TopoLVM on RKE, I noticed that the node DaemonSet's list of allowedHostPaths could not be tweaked via the values.yaml of the Chart. That's a problem because my kubelet directory is `/var/opt/rke/var/lib/kubelet` instead of `/var/lib/kubelet`.

It's already possible to customize the allowedHostPaths of the lvmd DeamonSet, so I mostly did it the exact same way.

This is my first PR, apologies if you missed something :-)

PS: I'm also good with the change of MIT license to Apache License 2.0 for this and future contributions.

Kind regard,
Laurent

### DIFF
--- a/charts/topolvm/Chart.yaml
+++ b/charts/topolvm/Chart.yaml
@@ -4,7 +4,7 @@ type: application
 home: https://github.com/topolvm/topolvm
 name: topolvm
 description: Topolvm
-version: 2.0.2
+version: 2.1.0
 appVersion: 0.8.3
 kubeVersion: ">=1.18.0"
 sources:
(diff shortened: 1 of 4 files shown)


satoru-takeuchi (dismissed):

toshipp:
@debackerl Thank you for your contribution! It seems good.
BTW, the chart lint fails, would you update the version to 2.0.4?
(this is boring, we may change the lint in future though)

debackerl:
Sorry @toshipp I missed you remark about 2.0.4, I already did 2.1.0 since I was adding new variables, I thought it was adding features (minor version bump), not a simple bugfix (patch version bump).

toshipp:
@debackerl You are correct. I forget the chart major version is 2 rather than 0.

toshipp (approved):

satoru-takeuchi (approved):

PR #349, opened 2021-08-20
Allow customization of kubeletRegistrationPath

debackerl:
Hello,

As I am installing topolvm on RKE/Fedora CoreOS, I needed to tweak the kubelet-registration-path passed to the csi-registrar container.

Reason is that kubelet runs in a container on RKE, that container uses one binding in particular to access the UNIX socket:

docker inspect kubelet:
```
...
            {
                "Type": "bind",
                "Source": "/opt/rke/var/lib/kubelet",
                "Destination": "/opt/rke/var/lib/kubelet",
                "Mode": "shared,z",
                "RW": true,
                "Propagation": "shared"
            }
...
```

No other bind is matching the default /var/lib/kubelet/plugins/topolvm.cybozu.com/node/csi-topolvm.sock path. The closest I could find is

```
            {
                "Type": "bind",
                "Source": "/var/lib/kubelet/volumeplugins",
                "Destination": "/var/lib/kubelet/volumeplugins",
                "Mode": "shared,z",
                "RW": true,
                "Propagation": "shared"
            }
```

but that's for Flex Volumes. Not sure why they have two kubelet directories...

Anyway, thanks to the new node.kubeletRegistrationPath that I introduce, I can set it to /opt/rke/var/lib/kubelet/plugins/topolvm.cybozu.com/node/csi-topolvm.sock and that let the csi-registrar succeed.

Thank for reviewing :-)

Laurent

### DIFF
--- a/charts/topolvm/values.yaml
+++ b/charts/topolvm/values.yaml
@@ -180,7 +180,7 @@ node:
   # node.lvmdSocket -- Specify the socket to be used for communication with lvmd.
   lvmdSocket: /run/lvmd/lvmd.sock
 
-  # node.kubeletRegistrationPath -- path to the CSI UNIX socket as seen by kubelet
+  # node.kubeletRegistrationPath -- Path to the CSI UNIX socket as seen by kubelet
   kubeletRegistrationPath: /var/lib/kubelet/plugins/topolvm.cybozu.com/node/csi-topolvm.sock
 
   # node.securityContext. -- Container securityContext.
--- a/charts/topolvm/values.yaml
+++ b/charts/topolvm/values.yaml
@@ -180,6 +180,9 @@ node:
   # node.lvmdSocket -- Specify the socket to be used for communication with lvmd.
   lvmdSocket: /run/lvmd/lvmd.sock
 
+  # node.kubeletRegistrationPath -- path to the CSI UNIX socket as seen by kubelet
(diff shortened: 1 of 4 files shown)


debackerl:
Hello,

If I understand correctly your pipeline, the pipeline failed because I updated the README.md manually, while the pipeline is supposed to do it, so the following code from the pipeline got upset:
```
      - name: Run helm-docs
        run: helm-docs && git diff --no-patch --exit-code
```

Is that correct? Shall I re-commit by reverted the README.md?

Thank you :-)

toshipp:
@debackerl
In our workflow, developers should create README.md by helm-doc, which is installed via `make tools` (or download it manually)
So, I'd like you to revise the file by the tool.

And another point, we change our lint not to check helm version, we now bump the version just before a new release.
Would you revert the version change?

charts/topolvm/values.yaml
```diff
@@ -180,6 +180,9 @@ node:
   # node.lvmdSocket -- Specify the socket to be used for communication with lvmd.
   lvmdSocket: /run/lvmd/lvmd.sock
 
+  # node.kubeletRegistrationPath -- Path to the CSI UNIX socket as seen by kubelet
```
cupnes, line 183: Please add a period to the end of the line.

cupnes (changes requested):

satoru-takeuchi:
@debackerl Could you reflect the comments from toshi and cupnes?

Issue #335, opened 2021-07-30
License change to Apache License 2.0 from MIT License


toshipp:
## Summary
We are planning to change the license of TopoLVM from MIT to Apache License because the current license is not clear on how to handle patents and trademarks.

## What you need
If you are a contributor who disagrees with this change, please reply to this Issue by September 1st 2021 12:00AM GMT.
If you agree to it, you do not need any action.
@UZER0 @briantopping @funkypenguin @frederiko @kfox1111 @matthias50 @onokatio @toelke @sebgl 

### Note
The Apache License is almost the same as the MIT, and does not affect end users as they are free to use it.

PR #351, opened 2021-08-22
duplicate label causes YAML parsing errors

khrisrichardson:
`{{- include "topolvm.labels" . | nindent 4 }}` also includes the `app.kubernetes.io/name` label, and the duplication causes YAML parsing errors

### DIFF
--- a/charts/topolvm/templates/mutatingwebhooks.yaml
+++ b/charts/topolvm/templates/mutatingwebhooks.yaml
@@ -8,7 +8,6 @@ metadata:
     cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/{{ template "topolvm.fullname" . }}-mutatingwebhook
   labels:
     {{- include "topolvm.labels" . | nindent 4 }}
-    app.kubernetes.io/name: {{ template "topolvm.fullname" . }}-hook
 webhooks:
   {{- if .Values.webhook.pvcMutatingWebhook.enabled }}
   - name: pvc-hook.topolvm.cybozu.com


toshipp (approved):
LGTM

toshipp:
Note: It seems that kubernetes accepts a yaml that has duplicated keys, but only uses the last one.

toshipp:
@khrisrichardson 
I will change our license from MIT to APL2.
background: https://github.com/topolvm/topolvm/issues/335
Would you provide your contribution under APL2?

llamerada-jp (approved):
LGTM

khrisrichardson:
@toshipp ironically the [YAML spec](http://yaml.org/spec/1.1/#id932806) indicates that the first entry wins, which is why certain parsers are less ambiguous with regards to handling duplicates. In my case, even though topolvm switched from kustomize to helm, I'm still leveraging kustomize's `helmCharts` directive to use the chart as a base layer, and kustomize's dependence on go-yaml surfaced the conflicting label.

Regarding MIT vs. Apache licensing, I don't know whether it's worth squabbling over in regards to this commit, which simply removed one line from a template.

toshipp:
@khrisrichardson 
Thank you for your explanation. I did not know kustomize supporting helm as a base.

On the topic of license changes, according to our legal department, files containing different licenses must be clearly marked so that they can be recognized. It is hard to maintain, so we want to change a license entirely.
Would you agree to the change?

toshipp:
@khrisrichardson 
I checked with the legal department and they said that the commit is not considered copyrightable.
Therefore, I will merge it after the deadline of #335.
If you have any explicit agreement or concerns, please comment.

PR #352, opened 2021-08-26
Fix YAML example in deploy/README.md

nonylene:
I think `controller.storageCapacityTracking.enabled` should be `true` in this YAML example.

> If you want to use Storage Capacity Tracking instead of using topolvm-scheduler, you need set the controller.storageCapacityTracking.enabled=true and scheduler.enabled=false in the Helm Chart values.
>
> https://github.com/topolvm/topolvm/blob/main/deploy/README.md


satoru-takeuchi (approved):

cupnes (approved):
LGTM

PR #356, opened 2021-09-02
Replace resizefs.go with mount-utils

toshipp:
We copied resizefs.go from kubernetes, but it migrated to mount-utils.
Therefore, we simply use mount-utils instead.


llamerada-jp:
MEMO:

`Resize` method in mount-utils supports btrfs.
ref https://github.com/kubernetes/mount-utils/blob/0ae0fec1f24fc52f35dabf92c671b8e0ae49333e/resizefs_linux.go#L93-L104

llamerada-jp (approved):
LGTM

satoru-takeuchi (approved):

PR #358, opened 2021-09-04
Fix lvmd is not previleged in deploying with Helm

faruryo:
Helm fails to deploy lvmd under PSP. Fix this.

lvmd
```
  Warning  FailedCreate      29s (x14 over 70s)  daemonset-controller  Error creating: pods "topolvm-lvmd-0-" is forbidden: PodSecurityPolicy: unable to admit pod: [spec.containers[0].securityContext.privileged: Invalid value: true: Privileged containers are not allowed]
```


llamerada-jp (approved):
LGTM

satoru-takeuchi (approved):

PR #372, opened 2021-10-01
[helm] support `existingCertManagerIssuer`

dungdm93:
Support user specify cert-manager issuer.

Currently, `topolvm` helm chart only support managed CA issuer, which is unsatisfactory in some org that has very strict policy of (internal) certificates. So, it's nice to provide an option to end-user provide there own cert-manager issuer instead of helm default.


toshipp (approved):

cupnes (approved):
LGTM

PR #373, opened 2021-10-01
[helm] support `PodMonitor`

dungdm93:
Support `PodMonitor` for `PrometheusOperator` users.

Signed-off-by: Đặng Minh Dũng <dungdm93@live.com>


toshipp:
Hi @dungdm93 
I am concerned about this PR being merged because we do not use PrometheusOperator,
it is hard to review and maintain this for us.

If you maintain the feature, we intend to merge this.
Would you let me know your thought on this idea?

dungdm93:
@toshipp This MR heavy inspire from Bitnami helm charts, therefore, the features are almost complete.
Here is an quick demo:
Screenshot from 2021-10-05 12-18-52
Screenshot from 2021-10-05 12-18-41
If there is any future issues, I'm very happy to help.

toshipp:
@dungdm93 I appreciate your help.
I delegate @satoru-takeuchi and @cupnes to further review, however, my quick thought is `controller.metrics` or `node.metrics` is too generic, it is actually for Prometheus Operator, so it may be better to be `controller.metrics.PrometheusOperator`, IMO.

dungdm93:
@toshipp `PrometheusOperator` has several CRD `PodMonitor`, `ServiceMonitor`, `Probe` and `Prometheus` and  `Alertmanager`. So named as `controller.metrics.PrometheusOperator` make no scene.
Further more, `metrics.podMonitor` and `metrics.serviceMonitor` are de-factor in helm community.

toshipp:
@dungdm93 I mean `controller.metrics.PrometheusOperator` section is for PrometheusOperator, so add podMonitor to `controller.metrics.PrometheusOperator.podMonitor`.
I personally do not agree podMonitor put under metrics.
If other people want to add another metrics operator configuration, how we should do it? put it as `controller.metrics.SomeOtherCR`?

If you know how other projects treat this, let me know.

dungdm93:
@toshipp Sorry, I mean `podMonitor` & `serviceMonitor` are de-factor (not with `metrics` prefix).
And yeah, in most charts it's located under `metrics`, some others is under `prometheus` (e.g `cert-manager`), or in root (e.g `grafana`). But I don't see any chart nested it in `metrics.prometheusOperator`.

`metrics.prometheusOperator.podMonitor` is bit more longer and harder for end-user, but I'm OK to change.
However, we should make final decision first, cc @cupnes, @satoru-takeuchi

satoru-takeuchi:
@dungdm93 About podMonitor related paramer names, I consider metrics is ok. Current chart already use `metrics` as prometheus's metrics, for example, in `node.metrics.enabled`.

Thank you for adding new parameters and making some cleanups. I'll approve this PR after my trivial change request is reflected.

toshipp:
@satoru-takeuchi 
`metrics.enable` is a predicate to add an annotation for prometheus rather than prometheus operator.
(And currently this style seems to be deprecated on upstream, https://github.com/prometheus/prometheus/pull/4131)
So two different settings in the same section confuse uses, IMO.

dungdm93:
Updated. `podMonitor` are now `node.prometheus.podMonitor` and `controller.prometheus.podMonitor`
@toshipp, @@satoru-takeuchi could you take a look

PR #374, opened 2021-10-07
Restart csi-registrar if registration is failed.

toshipp:
Sometimes, csi-registrar failed to register topolvm into kubelet due to lost
connection. In theory, the connection loss can be handled by grpc library
automatically, but it does not work by some kind of bugs. We use
livenessProbe to restart csi-regisrar instead. The functionality to detect
the case is introduced as of v2.3.0, the version depends on kubernetes
1.20 or higher, which is higher than our minimum supported version 1.19.
I think it is ok because the plugin registration system does not change.


llamerada-jp:
MEMO:
This change is based on the following document
https://github.com/kubernetes-csi/node-driver-registrar/tree/release-2.3#health-check-with-an-exec-probe

Issue #167, opened 2020-07-20
Logo


masa213f:
## What

We need a project logo!

## How

- Decide on a theme color.
- Decide on the logo design.
- Prepare svg/png/jpeg format images of the logo.

## Checklist

- [ ] Finish implementation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

PR #376, opened 2021-10-12
Add logo

toshipp:
fixes: #167


toshipp:
Good catch! I'll do that.

PR #385, opened 2021-10-20
add snapshot

little-guy-lxr:


toshipp:
@little-guy-lxr 
I have some consideration.
A Snapshot managed by topolvm is stored on the local node where its original volume sits. So if the node is destroyed, the snapshot will be unusable. For PVC, we delete it when a node is deleted from k8s. I think a similar feature is needed.
Another aspect, would you describe the feature on documents? Unlike another csi driver, topolvm manages local snapshot only, so snapshots are not useful for backup. 
Finally, would you add e2e test for the feature?

little-guy-lxr:
e2e test has added and add some basic docs for snapshot

llamerada-jp:
CI failed. Could you update this PR? [e2e-k8s-daemonset-lvmd (current, true)](https://github.com/topolvm/topolvm/runs/4100044228?check_suite_focus=true) is a known  to be an intermittent failure and is not needed to fix.

little-guy-lxr:
pipeline maybe is not stable, in fork versoin , the pipline all paased. could you rerun all the job ?

satoru-takeuchi:
@little-guy-lxr We'll start to review once we confirm all green.

e2e/snapshot_test.go
```diff
@@ -0,0 +1,214 @@
...
+			}
+
+			var pvc corev1.PersistentVolumeClaim
+			err = json.Unmarshal(result, &pvc)
+			if err != nil {
+				return err
+			}
+			if pvc.Status.Phase != corev1.ClaimBound {
+				return fmt.Errorf("pvc %s phase is %s but expect %s and status: %s", pvc.Name, pvc.Status.Phase, corev1.ClaimBound, pvc.Status.String())
+			}
```
llamerada-jp, line 159: It looks like CI is failing in the same place. I think it is better to output more logs temporary to find out why it is failing.

satoru-takeuchi: @little-guy-lxr We have run the CI several times, but it keeps failing as @llamerada-jp said. Please investigate the problem on your side and make it stable.  Thank you!

satoru-takeuchi:
@little-guy-lxr All CI has passed, thank you. We'll review this PR soon.

satoru-takeuchi:
@little-guy-lxr I left a comment about the design document but the code review is not finished yet. Please update this doc without waiting for the remaining review.

little-guy-lxr:
> @little-guy-lxr I left a comment about the design document but the code review is not finished yet. Please update this doc without waiting for the remaining review.

@satoru-takeuchi  sorry, I didn't see any comment in the design document.

docs/snapshot.md
```diff
@@ -0,0 +1,22 @@
+Snapshot Design
+==================
```
satoru-takeuchi, line 2: In addition to sequence diagrams, could you also write down the design in text like [this](https://github.com/topolvm/topolvm/blob/main/docs/design.md)?

satoru-takeuchi (changes requested):

satoru-takeuchi:
@little-guy-lxr Oops, sorry. I forgot to push "request changes" button. Please see [this comment](https://github.com/topolvm/topolvm/pull/385#discussion_r755877305).

llamerada-jp (changes requested):
First, I checked the documentations. Other files are currently being reviewed.

llamerada-jp:
@little-guy-lxr 
I am checking on the design and it will take some time to reply.

e2e/testdata/snapshot/deploy/controller.yaml
```diff
@@ -0,0 +1,42 @@
+
```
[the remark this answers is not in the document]
little-guy-lxr: > 

ok , I will replace the copy with the URL.  https://github.com/kubernetes-csi/external-snapshotter

satoru-takeuchi:
@little-guy-lxr 

I finished reading this PR. I consider that the design review is necessary before the more detailed review. Could you rewrite your design in the text by referring to the following document?

https://github.com/topolvm/topolvm/blob/main/docs/design.md

Please include the following information in the design document.

* use-case: I guess your use-case is taking snapshots locally and restored volume is also in the local node. Remote backup is out of scope.
* limitation: At least there are the following limitations coming from lvm's specification.
   * both snapshot and the restored volume are in the same node as the original volume
   * snapshot is in the same volume group as the original volume
* volume expansion is prohibited when snapshots exist

In addition, I have some requests.

* Add the way to specify the snapshot method. It will be necessary when topolvm supports another snapshot method like remote restoring.
* Pass the snapshot-related parameters via new fields in LogicalVolume CR than annotations.
* Consider how to handle invalidate state of the original volume. If the snapshot exhausts its data space, the original volume becomes invalidate state. One idea is to add the parameter to specify the size of the snapshot and make its default size 100%.

little-guy-lxr:
> @little-guy-lxr
> 
> I finished reading this PR. I consider that the design review is necessary before the more detailed review. Could you rewrite your design in the text by referring to the following document?
> 
> https://github.com/topolvm/topolvm/blob/main/docs/design.md
> 
> Please include the following information in the design document.
> 
> * use-case: I guess your use-case is taking snapshots locally and restored volume is also in the local node. Remote backup is out of scope.
> * limitation: At least there are the following limitations coming from lvm's specification.
>   
>   * both snapshot and the restored volume are in the same node as the original volume
>   * snapshot is in the same volume group as the original volume
> * volume expansion is prohibited when snapshots exist
> 
> In addition, I have some requests.
> 
> * Add the way to specify the snapshot method. It will be necessary when topolvm supports another snapshot method like remote restoring.
> * Pass the snapshot-related parameters via new fields in LogicalVolume CR than annotations.
> * Consider how to handle invalidate state of the original volume. If the snapshot exhausts its data space, the original volume becomes invalidate state. One idea is to add the parameter to specify the size of the snapshot and make its default size 100%.


@satoru-takeuchi Thank you for your good advice.  

- I will rewrite the design doc using text, but it will take some time.
- This snapshot implementation is all done on the local node, not considering remote backup.
- The flaws of lvm you mentioned are real
- about some requests  you mentioned, can we take some disscussion after I rewrited the design doc

little-guy-lxr:
@satoru-takeuchi  I have rewrite the design doc and update some related document. please help me  check , thanks.

satoru-takeuchi:
@little-guy-lxr I'll read it in next year, thank you!

docs/snapshot.md
```diff
@@ -0,0 +1,154 @@
...
+  name: pvc-42041aee-79a2-4184-91aa-5e1df6068b9f # restore pv name
+  annotations:
+    topolvm.cybozu.com/snapshot: df4b5a3e-8285-4554-b964-8aec23339be1 # snapshotVolID
+    topolvm.cybozu.com/volumemode: Filesystem   # Filesystem or Block
+    topolvm.cybozu.com/fstype: xfs # xfs or ext4
+spec:
+  name: pvc-42041aee-79a2-4184-91aa-5e1df6068b9f
+  nodeName: 192.168.26.40
+  size: 1Gi
+```
```
llamerada-jp, line 112: We'd like to expand LogicalVolume CRD by adding the snapshot-related parameters via new fields instead of annotations as follow.

```suggestion
```yaml
apiVersion: topolvm.cybozu.com/v1
kind: LogicalVolume
metadata:
  name: pvc-42041aee-79a2-4184-91aa-5e1df6068b9f # restore pv name
  annotations:
spec:
  name: pvc-42041aee-79a2-4184-91aa-5e1df6068b9f
  nodeName: 192.168.26.40
  size: 1Gi
  # dataSource field is nil when normal allocation.
  dataSource:
    # It describes how to copy data from data source. It must be `snapshot`.
    copyMode: snapshot
    # LogicalVolume name of data source.
    name: <string>
status:
  volumeID: <Store the volume id of the lvm snapshot as same as existing LV.>
```
```

File: docs/snapshot.md [hunk 5]
```diff
@@ -0,0 +1,154 @@
...
+
```
llamerada-jp, line 113: ### Validate pods having multiple PVCs that are restored from snapshots.

Users can apply a pod having some PVCs that are restored from snapshots. If users specify PVCs with snapshots placed on a different node as the data sources, the restore will fail and the pod will not start. I think it is better to validate it before applying manifests using a validation webhook. This feature is not a requirement, but I would like to implement this feature.

llamerada-jp (changes requested):
I think we are not at the stage of reviewing the implementation yet. So, to ease the review, could you remove the code from this PR for now. Please respond to the comments regarding our ["some requests"](https://github.com/topolvm/topolvm/pull/385#issuecomment-985355038).If you have any objections, please comment on the design we came up with, which reflects all these requests.

llamerada-jp:
@little-guy-lxr 
sp98 is in a hurry to add the proposals about thin volume(#442) and snapshot using thin volume(#446). Therefore, we are dealing with that. To deal with the thin volume, we should update the specification of the snapshot in this PR. And some of the common features of the snapshot may implement in those.

toshipp:
@little-guy-lxr If you still have an interest, please rework it based on the main branch.

Issue #406, opened 2021-12-03
Unable to attach volume in block mode


rkrzewski:
**Describe the bug**
`topolvm-node` reports that `NodePublishVolume(block)` command succeeded, but no device node is created in `/var/snap/microk8s/common/var/lib/kubelet/plugins/kubernetes.io/csi/volumeDevices/publish/<pv>`

**Environments**
- Version: topolvm 0.10.2, microk8s v1.21.7
- OS: Ubuntu 20.04.3 LTS, Linux x86_64 5.4.0-91-generic

**To Reproduce**
While trying to start a Ceph OSD pod, that is using topolvm volume in block mode I see the following events:
```
  Type     Reason           Age                   From     Message
  ----     ------           ----                  ----     -------
  Warning  FailedMount      42m (x3 over 76m)     kubelet  Unable to attach or mount volumes: unmounted volumes=[set1-data-1-7b67p], unattached volumes=[rook-ceph-log rook-ceph-crash set1-data-1-7b67p run-udev set1-data-1-7b67p-bridge kube-api-access-gf9fb rook-data rook-config-override]: timed out waiting for the condition
  Warning  FailedMapVolume  109s (x50 over 87m)   kubelet  MapVolume.EvalHostSymlinks failed for volume "pvc-a0365ed0-9ff9-4b5f-947d-9ec87f6dcee8" : lstat /var/snap/microk8s/common/var/lib/kubelet/plugins/kubernetes.io/csi/volumeDevices/publish/pvc-a0365ed0-9ff9-4b5f-947d-9ec87f6dcee8/0485c803-5831-43d6-b8e6-4cc40363d2c6: no such file or directory
```
Directory `/var/snap/microk8s/common/var/lib/kubelet/plugins/kubernetes.io/csi/volumeDevices/publish/pvc-a0365ed0-9ff9-4b5f-947d-9ec87f6dcee8/` exists but `0485c803-5831-43d6-b8e6-4cc40363d2c6` entry is in fact missing.

When looking for messages mentioning PV `pvc-a0365ed0-9ff9-4b5f-947d-9ec87f6dcee8` in the logs of `topolvm-node` container, I see the following:
```
[...]

PR #407, opened 2021-12-06
Make CSI --kubelet-registration-path configurable in the Helm chart

rkrzewski:


rkrzewski:
As you can see in #406 I had to override a bunch of paths to get the `topolvm-node` component to behave on microk8s.

Maybe introducing a single setting like `node.kubeletWorkDirectory` with default value of `/var/log/kubelet` that could be overriden to `/var/snap/microk8s/common/var/log/kubelet` or even setting `node.microk8s` with default value of `false` that would switch between those paths would be a better idea? Or both? Let me know, I'll update the PR.

toshipp:
As you say, it may be better to introduce `node.kubeletWorkDirectory`, but we chose to introduce parameters for volumes and volumeMounts for flexibility. If introducing the parameter, they conflict with each other and it may confuse users, so keeping your current implementation is better than that, IMO.
For `node.microk8s`, I think it is not needed because your PR is sufficient for the issue, and it breaks simplicity if introducing it.

toshipp:
Could you update `charts/topolvm/README.md` by executing `./bin/helm-docs`?

rkrzewski:
I've updated documentation as requested.

rkrzewski:
I'm pondering the possible configuration options, and it think it would be preferable to:
- introduce `node.kubeletWorkDirectory` parameter
- move the contents of `node.volumes` and `node.volumeMounts.topolvmNode` to `templates/node/deamonset.yaml` and replace `/var/lib/kubelet/` prefix in the paths with `node.kubeletWorkDirectory` parameters
- initialize `node.volumes` and `node.volumeMounts.topolvmNode` with value of `{}` in `values.yaml` and add logic to `templates/node/deamonset.yaml` use them for overriding default configuration them if necessary.

This would allow the following benefits:
- It would do the right thing in the default case of `/var/lib/kubelet` work directory
- Deploying on microk8s would require overriding a single parameter instead of a large chunk yaml (see #406)
- In the case that overriding individual paths is necessary it still can be done by overriding `node.volumes` and `node.volumeMounts.topolvmNode` however my understanding is that it should not be necessary. These paths need to change together for the integration to work: both kubelet and CSI sidecars need to see the same data under the same paths for the commands to be processed correctly.

I realize that it's a bit of an intrusive change, especially that there are more components that have their mounts declared via `values.yaml` besides node deaemonset, so I am waiting for your opinion before proceeding.

toshipp:
It looks good. Anyway, it is not just a refactoring but adding a feature, so would you separate it from this PR if you proceed?

rkrzewski:
I'll open a separate PR, because it'll be an alternative solution to the same problem. If you accept the other one, this PR can be discarded.

daichimukai:
An alternative solution #410 is merged. Close this.

Issue #252, opened 2021-01-26
scheduler extender convert to scheduler framework 


little-guy-lxr:
As we know , kubernetes may deprecrate scheduler extender for which has some shotcommings and is unfriendly for user, so recommend topolvm could take kubernetes scheduler framework as new scheduler extension.

ymmt2005:
We will replace the scheduler extender with Storage Capacity Tracking in #193.
With storage capacity tracking, we can implement an optimal volume scheduling with the standard scheduler.

toshipp:
We have no plan to do this now.

PR #408, opened 2021-12-07
Update links in design doc

lentzi90:
This updates some old links regarding the scheduler design proposal. I also changed `extender` -> `plugin` in the same places since this seems to be the current terminology used. However, I'm sure there are many more places where we have "extender", not sure if it is worth updating all of them since it is still easy to understand I think.

The old link goes here: https://github.com/kubernetes/community/blob/master/contributors/design-proposals/scheduling/scheduler_extender.md
It links to [this archived repo](https://github.com/kubernetes/design-proposals-archive/blob/main/scheduling/scheduler_extender.md). But I think it makes more sense to change the link to [the scheduling framework docs](https://kubernetes.io/docs/concepts/scheduling-eviction/scheduling-framework/) (which is what I did in this PR), or to [this KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-scheduling/624-scheduling-framework/README.md) (which is linked in the scheduler framework docs also).

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ A diagram of components is available in [docs/design.md](docs/design.md#diagram)
 This repository contains these programs:
 
 - `topolvm-controller`: CSI controller service.
-- `topolvm-scheduler`: A [scheduler extender](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/scheduling/scheduler_extender.md) for TopoLVM.
+- `topolvm-scheduler`: A [scheduler plugin](https://kubernetes.io/docs/concepts/scheduling-eviction/scheduling-framework/) for TopoLVM.
 - `topolvm-node`: CSI node service.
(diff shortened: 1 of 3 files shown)


cupnes:
@lentzi90 Thank you for your PR.

But, "Scheduler Extender" and "Scheduling Framework (plugin)" are different. It is a mistake to change "Scheduler Extender" to "Scheduling Framework (plugin)" since it is "Scheduler Extender" that is being referred to in the pointed out section.

cupnes:
@lentzi90 On the other hand, the documentation for Scheduler Extender is archived. This URL will be updated in the #411.

lentzi90:
Ah thank you, now I understand!
I noticed also that the scheduler extender will probably be removed according to [this](https://github.com/topolvm/topolvm/issues/252#issuecomment-767348624) so maybe not so important to update this.

PR #409, opened 2021-12-07
Add topolvm-controller CLI flag to skip node finalize

### LINKED ISSUE #397: Support for disabling pvc auto removal
There should be option to disable PVC deletion when node is deleted.

Certain K8s cluster use node replacement strategy for node upgrades. (CAPI does this) In those cases node is temporary removed from cluster and rebuild. After that it joins back to the cluster. Node name will be same and all data disks will be kept intact. 
So once the node joins back to the cluster it would behave similarly as in case of node reboot. For point of view of Topolvm.

This could also be used in cases where host operating system needs to be reinstalled, but data disks would like to be reused with their data. So that expensive data migration would not be needed.

macaptain:
Adds a flag `--skip-node-finalize` to the topolvm-controller, which skips the doFinalize function in the node controller. When the flag is enabled, PersistentVolumeClaims are not automatically cleaned up when a Node is deleted. An administrator for the Kubernetes cluster should clear up the PVCs/LogicalVolumes manually.

The Helm value controller.nodeFinalize.skipped has been added to support the new flag from the Helm chart.

The e2e tests and documentation are updated as well, along with a new entry in the matrix for the GitHub action, which deploys the chart with the flag turned on/off. The cleanup checks have been amended to cover both cases.

Fixes #397

### DIFF
--- a/docs/topolvm-controller.md
+++ b/docs/topolvm-controller.md
@@ -159,7 +159,11 @@ Controllers
 `topolvm-metrics` adds `topolvm.cybozu.com/node` finalizer.
 
 When a Node is being deleted, the controller deletes all PVCs and LogicalVolumes for TopoLVM
-on the deleting node.
+on the deleting node. 
+
+This node finalize procedure may be skipped with the `--skip-node-finalize` flag. 
+When this is true, the PVCs and the LogicalVolume CRs from a deleted node must be
+deleted manually by a cluster administrator.
 
 ### PVC finalizer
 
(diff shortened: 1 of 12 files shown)


macaptain:
Hi @llamerada-jp. I put together this PR based on your suggestions in #397. I couldn't disable the NodeReconciler completely, because the finalizer is added by topolvm-node, and until the finalizer is removed, node deletion will block forever. Instead I'm skipping the steps in `doFinalize`, and then the finalizer is removed by the NodeReconciler.

I've run the e2e tests in both cases (`SKIP_NODE_FINALIZE=false` and `SKIP_NODE_FINALIZE=true`) and verified that these tests pass in my local environment.

I'm also happy to rename the flag if you think there is a better name for it. Let me know what you think!

e2e/cleanup_test.go
```diff
@@ -53,8 +53,10 @@ func testCleanup() {
 						topolvmFinalize = true
 					}
 				}
+				// Even if node finalize is skipped, the finalizer is still present on the node
+				// The finalizer is added by the metrics-exporter runner from topolvm-node
```
macaptain, line 57: I was surprised to find out that topolvm-node's metrics-exporter is responsible for patching Node finalizers, because I thought it would only do metrics. 

[Link to code](https://github.com/topolvm/topolvm/blob/9eea3dd64b3123a378c86c6bdeb485e9fdbba537/runners/metrics_exporter.go#L131-L150)

Perhaps it would be good to refactor so that patching the Node is the responsibility of a different function in topolvm-node? (Not in this PR, I will raise a separate issue and PR if there's agreement. :slightly_smiling_face: )

docs/topolvm-controller.md
```diff
@@ -159,7 +159,11 @@ Controllers
 `topolvm-metrics` adds `topolvm.cybozu.com/node` finalizer.
 
 When a Node is being deleted, the controller deletes all PVCs and LogicalVolumes for TopoLVM
-on the deleting node.
+on the deleting node. 
+
+This node finalize procedure may be skipped with the `--skip-node-finalize` flag. 
+When this is true, the PVCs from a deleted node must be deleted manually
```
llamerada-jp, line 165:
```suggestion
When this is true, the PVCs and the LogicalVolume CRs from a deleted node must be deleted manually
```

llamerada-jp (changes requested):

docs/topolvm-controller.md
macaptain: Updated.

llamerada-jp (approved):
LGTM

daichimukai (approved):

daichimukai:
Thank you for contribution. LGTM.

Please wait for a while for the merge, as we need to adjust the required checks for the merge due to CI matrix updated.

PR #410, opened 2021-12-08
Make kubelet work directory overridable via single chart parameter

rkrzewski:
/var/lib/kubelet directory appears multiple times in node component
configuration. node.kubeletWorkDirectory parameter allows overriding
them all in concert, which is useful for deplyment on microk8s where
/var/snap/microk8s/common/var/lib/kubelet directory needs to be used
instead.

This is an alternative to #407. Either one can be used to solve the problem of deployment on microk8s cluster described in #406. This PR is larger but provides more streamlined user experience: only 1 string property to override instead of 4, 3 of which require structured values.

### DIFF
--- a/charts/topolvm/templates/lvmd/psp.yaml
+++ b/charts/topolvm/templates/lvmd/psp.yaml
@@ -14,9 +14,13 @@ spec:
     - 'emptyDir'
     - 'hostPath'
     - 'secret'
-  {{- with .Values.lvmd.psp.allowedHostPaths }}
-  allowedHostPaths: {{ toYaml . | nindent 2 }}
-  {{- end }}
+  allowedHostPaths:
+    {{- if .Values.lvmd.psp.allowedHostPaths }}
+    {{- toYaml .Values.lvmd.psp.allowedHostPaths | nindent 4 }}
+    {{- else }}
+    - pathPrefix: {{ dir .Values.lvmd.socketName }}
+      readOnly: false
+    {{- end }}
   runAsUser:
     rule: 'RunAsAny'
   seLinux:
--- a/charts/topolvm/templates/node/psp.yaml
+++ b/charts/topolvm/templates/node/psp.yaml
@@ -14,9 +14,15 @@ spec:
     - 'emptyDir'
     - 'secret'
     - 'hostPath'
-  {{- with .Values.node.psp.allowedHostPaths }}
-  allowedHostPaths: {{ toYaml . | nindent 2 }}
-  {{- end }}
+  allowedHostPaths:
+    {{- if .Values.node.psp.allowedHostPaths }}
+    {{- toYaml .Values.node.psp.allowedHostPaths | nindent 4 }}
+    {{- else }}
+    - pathPrefix: {{ .Values.node.kubeletWorkDirectory }}
+      readOnly: false
+    - pathPrefix: {{ dir .Values.node.lvmdSocket }}
+      readOnly: false
+    {{- end }}
   hostNetwork: false
   runAsUser:
     rule: 'RunAsAny'
--- a/charts/topolvm/values.yaml
+++ b/charts/topolvm/values.yaml
@@ -157,16 +157,16 @@ lvmd:
   nodeSelector: {}
 
   # lvmd.volumes -- Specify volumes.
-  volumes:
-    - name: lvmd-socket-dir
-      hostPath:
-        path: /run/topolvm
-        type: DirectoryOrCreate
+  volumes: []
+  #  - name: lvmd-socket-dir
+  #    hostPath:
+  #      path: /run/topolvm
+  #      type: DirectoryOrCreate
 
   # lvmd.volumeMounts -- Specify volumeMounts.
-  volumeMounts:
-    - name: lvmd-socket-dir
-      mountPath: /run/topolvm
+  volumeMounts: []
+  #  - name: lvmd-socket-dir
+  #    mountPath: /run/topolvm
 
   # lvmd.additionalConfigs -- Define additional LVM Daemon configs if you have additional types of nodes.
   # Please ensure nodeSelectors are non overlapping.
@@ -181,9 +181,9 @@ lvmd:
 
   psp:
     # lvmd.psp.allowedHostPaths -- Specify allowedHostPaths.
-    allowedHostPaths:
-      - pathPrefix: "/run/topolvm"
-        readOnly: false
+    allowedHostPaths: []
+    #  - pathPrefix: "/run/topolvm"
+    #    readOnly: false
 
   # lvmd.updateStrategy -- Specify updateStrategy.
   updateStrategy: {}
@@ -195,7 +195,10 @@ lvmd:
 # CSI node service
 node:
   # node.lvmdSocket -- Specify the socket to be used for communication with lvmd.
-  lvmdSocket: /run/lvmd/lvmd.sock
+  lvmdSocket: /run/topolvm/lvmd.sock
+  # node.kubeletWorkDirectory -- Specify the work directory of Kubelet on the host.
+  # For example, on microk8s it needs to be set to `/var/snap/microk8s/common/var/lib/kubelet`
+  kubeletWorkDirectory: /var/lib/kubelet
 
   # node.securityContext. -- Container securityContext.
   ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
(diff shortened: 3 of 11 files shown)


toshipp:
@rkrzewski 
We took a look at this PR and have a concern about there is necessary to write down all volumes section if the lvmd socket path is changed. This is a similar issue to your motivation for the PR, so I think #407 is better, as well as in terms of simplicity.

rkrzewski:
#407 is indeed simpler in terms of implementation, but more complicated in usage. Helm offers path manipulation functions so generating correct mounts etc for a modified `node.lvmdSocket` can be implemented with little effort.

Before I delve into it I'd like to understand the requirements better

`lvmd.socketName` is `/run/topolvm/lvmd.sock` by default. The predefined volumes / volume mounts agree with the default value, but need to be overridden if the directory part of `lvmd.socketName` is changed. This could be improved.

`node.lvmdSocket` is `/run/lvmd/lvmd.sock` by default. It appears to disagree with default value of `lvmd.socketName`

`lvmd-socket-dir` volume for `node` component refers to `/run/topolvm` so it agrees with default socket location but disagrees with `node.lvmdSocket` location. It is then mounted at `/run/lvmd` location in the `topolvm-node` container. At the same time 
`--lvmd-socket={{ .Values.node.lvmdSocket }}` is passed to the process. As long as directory part of `node.lvmdSocket` is not changed, it works albeit in a roundabout way.

I thought it would be more natural to set `node.lvmdSocket` a default value of `/run/topolvm/lvmd.sock`, use `{{ dir .Values.node.lvmdSocket }}` host path for `lvmd-socket-dir` volume and mount it at the same path in the container, however there is a problem: `/run/topolvm` path in the container is used for `node-plugin-dir` volume mounted from `/var/lib/kubelet/plugins/topolvm.cybozu.com/node` at the host. `--csi-address=/run/topolvm/csi-topolvm.sock` is then passed to CSI sidecar containers. 

Okay, after following through the paths and mounts carefully and writing the above I have a plan for improvement. I'll add a couple more commits to PR for your consideration.

e2e/manifests/values/daemonset-lvmd-storage-capacity.yaml
```diff
@@ -32,59 +32,9 @@ lvmd:
...
-        type: DirectoryOrCreate
-    - name: lvmd-socket-dir
-      hostPath:
-        path: /tmp/topolvm/daemonset_lvmd
-        type: Directory
-    - name: device-dir
-      hostPath:
-        path: /dev
-        type: Directory
+  lvmdSocket: /run/topolvm/daemonset_lvmd/lvmd.sock
```
toshipp, line 37: We set `/tmp/topolvm/daemonset_lvmd/lvmd.sock`  to `lvmd.socketName`, so `node.lvmdSocket` is also set as well, IMO.

e2e/manifests/values/deployment-scheduler.yaml
```diff
@@ -32,47 +32,6 @@ scheduler:
...
-      hostPath:
-        path: /var/lib/kubelet/plugins/kubernetes.io/csi
-        type: DirectoryOrCreate
-    - name: pod-volumes-dir
-      hostPath:
-        path: /var/lib/kubelet/pods/
-        type: DirectoryOrCreate
-    - name: lvmd-socket-dir
-      hostPath:
-        path: /tmp/topolvm
```
toshipp, line 69: We use `/tmp/topolvm` to store a lvmd socket, which is set by the kind configuration.
https://github.com/topolvm/topolvm/blob/1dad6f5a0389163027a0c061cf431ef3cf1412e5/e2e/topolvm-cluster.yaml#L49
In this case, lvmds running on the host use `/tmp/topolvm/lvmd[1-3]/lvmd.sock`, kind nodes mounts these paths on `/tmp/topolvm`, so topolvm-node containers need to mount them.
There are two possible solutions.
The first is to update values.yaml to use `/tmp/topolvm`.
The second is to update the kind configuration to mount the host path on `/run/topolvm` that is a default of helm charts.
I assume the second one can work and reduce values.yaml, but the first one is recommended because it avoid facing  other problems.

toshipp (changes requested):

e2e/manifests/values/deployment-scheduler.yaml
rkrzewski: My mistake! I misread the location `/tmp/topolvm` as `/run/topolovm`. Fixing.

rkrzewski: I've added the change, but I think it's still not right. Earlier manifest mounts `lvmd-socket-dir` at `/run/lvmd` and now it'll end up at `/run/topolvm`. Further, I'm puzzled how `topolvm-node` interacts with `/tmp/topolvm/lvmd[1-3]/lvmd.sock` since I don't see where those locations are being passed to it.

charts/topolvm/templates/lvmd/psp.yaml
```diff
@@ -14,9 +14,13 @@ spec:
     - 'emptyDir'
     - 'hostPath'
     - 'secret'
-  {{- with .Values.lvmd.psp.allowedHostPaths }}
-  allowedHostPaths: {{ toYaml . | nindent 2 }}
-  {{- end }}
+  allowedHostPaths: 
+    {{- if .Values.lvmd.psp.allowedHostPaths }}
+    {{- toYaml .Values.lvmd.psp.allowedHostPaths | nindent 2 }}
```
daichimukai, line 19:
```suggestion
    {{- toYaml .Values.lvmd.psp.allowedHostPaths | nindent 4 }}
```

I want to match the level of indentation with the else clause.

charts/topolvm/templates/node/psp.yaml
```diff
@@ -14,9 +14,15 @@ spec:
     - 'emptyDir'
     - 'secret'
     - 'hostPath'
-  {{- with .Values.node.psp.allowedHostPaths }}
-  allowedHostPaths: {{ toYaml . | nindent 2 }}
-  {{- end }}
+  allowedHostPaths: 
+    {{- if .Values.node.psp.allowedHostPaths }}
+    {{- toYaml .Values.node.psp.allowedHostPaths | nindent 2 }}
```
daichimukai, line 19:
```suggestion
    {{- toYaml .Values.node.psp.allowedHostPaths | nindent 4 }}  
```

ditto

charts/topolvm/values.yaml
```diff
@@ -157,16 +157,16 @@ lvmd:
   nodeSelector: {}
 
   # lvmd.volumes -- Specify volumes.
-  volumes:
-    - name: lvmd-socket-dir
-      hostPath:
-        path: /run/topolvm
-        type: DirectoryOrCreate
+  volumes: {}
```
daichimukai, line 160:
```suggestion
  volumes: []
```

The type of value of `volumes` is array, so `[]` is would be appropriate. It seems there is a similar problem in a few places.

daichimukai (changes requested):

rkrzewski:
Thanks for the review @daichimukai. I have applied the requested changes.

e2e/manifests/values/deployment-scheduler.yaml
toshipp: Socket paths on the host `/tmp/topolvm/lvmd[1-3]/lvmd.sock ` is mapped to `/tmp/topolvm/lvmd.sock` for each node by these settings.
https://github.com/topolvm/topolvm/blob/d64dda32ef7f0d7bcbd5ad9d4e5ee5a0447380b1/e2e/topolvm-cluster.yaml#L49-L50
https://github.com/topolvm/topolvm/blob/d64dda32ef7f0d7bcbd5ad9d4e5ee5a0447380b1/e2e/topolvm-cluster.yaml#L58-L59
https://github.com/topolvm/topolvm/blob/d64dda32ef7f0d7bcbd5ad9d4e5ee5a0447380b1/e2e/topolvm-cluster.yaml#L67-L68
From k8s node perspective, each node has the socket on `/tmp/topolvm/lvmd.sock`.

toshipp (changes requested):
I'd passed tests locally with the following values. Would you change values accordingly?

```
diff --git a/e2e/manifests/values/daemonset-scheduler.yaml b/e2e/manifests/values/daemonset-scheduler.yaml
index 5fa3da84..2267486f 100644
--- a/e2e/manifests/values/daemonset-scheduler.yaml
+++ b/e2e/manifests/values/daemonset-scheduler.yaml
@@ -23,31 +23,7 @@ lvmd:
   managed: false

 node:
-  volumes:
-    - name: registration-dir
-      hostPath:
-        path: /var/lib/kubelet/plugins_registry/
-        type: Directory
-    - name: node-plugin-dir
-      hostPath:
-        path: /var/lib/kubelet/plugins/topolvm.cybozu.com/node
-        type: DirectoryOrCreate
-    - name: csi-plugin-dir
-      hostPath:
-        path: /var/lib/kubelet/plugins/kubernetes.io/csi
-        type: DirectoryOrCreate
-    - name: pod-volumes-dir
-      hostPath:
-        path: /var/lib/kubelet/pods/
-        type: DirectoryOrCreate
-    - name: lvmd-socket-dir
-      hostPath:
-        path: /tmp/topolvm
-        type: Directory
-    - name: device-dir
-      hostPath:
-        path: /dev
-        type: Directory
+  lvmdSocket: /tmp/topolvm/lvmd.sock

 storageClasses:
   - name: topolvm-provisioner
```

toshipp:
@rkrzewski 
Thank you for your update!
I confirmed that `deployment-scheduler.yaml` and `storage-capacity.yaml ` can be also updated as well as `daemonset-scheduler.yaml`. Would you fix them?

e2e/manifests/values/deployment-scheduler.yaml
```diff
@@ -32,21 +32,17 @@ scheduler:
 lvmd:
   managed: false
 
+storageClasses:
```
toshipp, line 35: It looks like you moved it by mistake.. Would you revert this if unnecessary?

rkrzewski: It was indeed a mistake, sorry about that. Fixing.

rkrzewski:
I've submitted the requested changes. Holding fingers crossed for the CI!

Would you like me to squash the changes before merging? There is some back-and-forth in those commits that won't provide much value from historical perspective.

There's one more thing that struck me while working with the CI manifests: there's a discrepancy between parameter names: `lvmd.socketName` vs `node.lvmdSocket`. Shouldn't the latter be `node.lvmdSocketName`? I realize this is a non-backwards compatible change and should be performed in a separate PR (if at all). Just wanted to put it out there.

e2e/manifests/values/deployment-scheduler.yaml
```diff
@@ -81,7 +43,7 @@ storageClasses:
       volumeBindingMode: WaitForFirstConsumer
       allowVolumeExpansion: true
       additionalParameters:
-        topolvm.cybozu.com/device-class: "ssd"
+        topolvm.cybozu.com/device-class: "ssd"
```
toshipp, line 46: Please remove trailing spaces.

toshipp:
> Would you like me to squash the changes before merging? There is some back-and-forth in those commits that won't provide much value from historical perspective.

Ok. I'll notify you to ready to squash.

> There's one more thing that struck me while working with the CI manifests: there's a discrepancy between parameter names: `lvmd.socketName` vs `node.lvmdSocket`. Shouldn't the latter be `node.lvmdSocketName`? I realize this is a non-backwards compatible change and should be performed in a separate PR (if at all). Just wanted to put it out there.

Agreed. BTW, I think this PR already breaks backward compatibility because this changes lvmdSocket value, so if someone tweaks volumeMounts, topolvm-nodes may not work properly. Therefore, I intend to release this feature as major update, I think the renaming can be done in this PR too.

@daichimukai Do you have any concerns about it?

toshipp:
I thought about it for a while, but I think it would be better to separate the PRs to make the intent clearer.

toshipp:
@rkrzewski 
I have no other comments except [this](https://github.com/topolvm/topolvm/pull/410/files#r773947142), please fix this and squash them.

daichimukai:
Now the changes (with deletion of trailing white space) look good to me. I'll approve PR when a squashed commit is pushed.

I have no objection on changing `node.lvmdSocket` to `node.lvmdSocketName` in the new PR :)

e2e/manifests/values/deployment-scheduler.yaml
rkrzewski: Done, I found a couple more in `psp.yaml` files

rkrzewski:
I've squashed the changes to a single commit and amended the commit message. If you have any further suggestions, please let me know.

toshipp (approved):
LGTM. You did a great job!

daichimukai (approved):
LGTM. Thanks @rkrzewski

rkrzewski:
Thank you, I really enjoyed working on this :)

I'll post the PR for renaming `node.lvmdSocket` to `node.lvmdSocketName` as discussed.

PR #412, opened 2021-12-09
change slack invitation link

toshipp:
Because the old form uses legacy API, I replace it with slack official invitation link


ymmt2005 (approved):
LGTM

daichimukai (approved):

Issue #398, opened 2021-11-19
When PSPs are enabled, csi-registrar container crashes due to read-only filesystem 


funkypenguin:
When deploying topolvm-node pods via the helm chart 3.1.2, the csi-registrar container crashes with the following error:

```
I1119 02:32:51.961824       1 main.go:164] Version:
I1119 02:32:51.961876       1 main.go:165] Running node-driver-registrar in mode=registration
I1119 02:32:51.966324       1 node_register.go:52] Starting Registration Server at: /registration/topolvm.cybozu.com-reg.sock
I1119 02:32:51.966798       1 node_register.go:61] Registration Server started at: /registration/topolvm.cybozu.com-reg.sock
I1119 02:32:51.966923       1 node_register.go:91] Skipping healthz server because HTTP endpoint is set to: ""
I1119 02:32:51.977493       1 main.go:100] Received GetInfo call: &InfoRequest{}
E1119 02:32:51.977676       1 main.go:105] "Failed to create registration probe file" err="mkdir /var/lib/kubelet: read-only file system" registrationProbePath="/var/lib/kubelet/plugins/topolvm.cybozu.com/node/registration"
I1119 02:32:51.977736       1 main.go:107] "Kubelet registration probe created" path="/var/lib/kubelet/plugins/topolvm.cybozu.com/node/registration"
I1119 02:32:52.021812       1 main.go:118] Received NotifyRegistrationStatus call: &RegistrationStatus{PluginRegistered:true,Error:,}
```

I've tracked this down to the fact that the [PSP deployed by the helm chart](https://github.com/topolvm/topolvm/blob/main/charts/topolvm/templates/node/psp.yaml#L29) sets `readOnlyRootFilesystem: true`

As a result, my topolvm-node pods are continuously crashlooping.

**To Reproduce**

1. Use a PSP-enabled cluster
2. Deploy topolvm using the helm chart, and set `podSecurityPolicy.create: true`
3. Observe the topolvm-node pods crashlooping
4. Observe the logs from the csi-registrar containers (`kubectl logs -n topolvm-system -l app.kubernetes.io/name=topolvm-node -c csi-registrar`)
5. Edit the PSP (`kubectl edit psp topolvm-node`) and delete the `readOnlyRootFilesystem` line
6. Restart the topolvm-node daemonset rollout (`kubectl rollout restart -n topolvm-system daemonset topolvm-node`)
7. Observe that the pods no longer crashloop, the error about read-only mode is gone 🥳 

[...]

PR #414, opened 2021-12-13
Use emptyDir for kubelet-registration-probe to support PSP enabled cluster

y1r:
I've also faced the same issue reported by https://github.com/topolvm/topolvm/issues/398.
This PR fixes the issue by using emptyDir as suggested by @toshipp .


toshipp:
Thank you for your effort!
We are reviewing #410 which tweaks the paths, and I think it resolves this together.

toshipp:
I've confirmed that the latest main branch works fine. We'll release it soom.

PR #416, opened 2021-12-15
Add proposal for allocation strategies

### LINKED ISSUE #415: Proposal: dedicated disks and allocation policy

It should be possible to ask for dedicated disks for performance sensitive applications that benefit from "owning" the whole disk they use. A "softer" variant of this is also of interest: basically setting the `--alloc` flag to `contiguous` or `cling` in order to keep the data on a single PV. In this case the PV can still be shared with other applications.

I have a proposal ready and will create a PR for it soon so we can discuss it.

We have a use case where we want dedicated disks for certain applications (e.g. ceph). The idea is that these applications should not share the PV/disk with other for performance reasons.
In some cases it may be too expensive to allocate the whole disk for a single application. But it would still be beneficial to pick the allocation policy (e.g. set it to `contiguous`).

lentzi90:
Allocation strategies is a feature to allow choosing how to allocate LVs
from the VG when using TopoLVM. For example, keep the LV on a single PV
or spread it over multiple.
This proposal suggests extending the current implementation that
supports stripes, to also expose the `alloc` flag of `lvcreate`. It also
proposes a way to ask for a "dedicated" disk so that the LV would
consume exactly one PV.

Fixes #415

### DIFF
--- a/docs/proposals/allocation-strategies.md
+++ b/docs/proposals/allocation-strategies.md
@@ -1,0 +1,151 @@
+# Allocation strategies
+
+<!-- toc -->
+
+<!-- /toc -->
+
+## Summary
+
+Allocation strategies is a feature to allow choosing how to allocate LVs from the VG when using TopoLVM.
+For example, keep the LV on a single PV or stripe it over multiple.
+
+## Motivation
+
+Some applications benefit from "owning" the whole disk they use and keeping the data on a single physical disk.
+For optimal performance, this physical disk should not be shared with other applications.
+
+### Goals
+
+- Make it possible to reserve a whole PV for one LV (`dedicated`).
+- Expose the `--alloc` flag of `lvcreate` to the TopoLVM user.
+- Keep backward compatibility.
+
+## Proposal
+
+### Option A) lvmd options
+
+Specify additional options for how to create the LVs in the `lvmd` config.
+This is similar to how striping is currently configured.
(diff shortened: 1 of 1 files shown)


llamerada-jp:
Hi @lentzi90 ,

We would like to discuss the `dedicated` feature and the `alloc` feature separately because they are different features. How about changing this discussion is only for the `dedicated` feature and creating the new discussion to handle the `alloc` feature?

Our first impression is as follows.

- The `dedicated` feature: It would be better to create a different controller than adding this feature to TopoLVM because this feature conflicts with the topolvm concepts a lot. 
- The `alloc` feature: We don't want to adopt the features this proposal suggested if possible. If you want to implement it, We would like to suggest that add a field for optional `lvcreate` options. So that the administrator of each system can add options at their responsibility.

As a side note, our further reply will be delayed because the Japanese new year holiday season will come soon...

lentzi90:
Thank you for taking the time to look into this!

I agree that it makes sense to split this into 2 separate things and will create a new PR for the `lvcreate` options feature.

So then for full disk allocation: Is there any point in discussing this further or is it too much in conflict with topolvm concepts? In that case I guess we can close this PR.

satoru-takeuchi:
> So then for full disk allocation: Is there any point in discussing this further or is it too much in conflict with topolvm concepts? In that case I guess we can close this PR.

Yes. we think this feature does not match topolvm, so I will close this ticket. In any case, thanks for this suggestion!

PR #417, opened 2021-12-17
skip PDB when topolvm-scheduler isn't enabled

### LINKED ISSUE #413: Request: Helm chart should disable PDB for scheduler if scheduler itself is disabled
**What should the feature do:** Helm chart should not template PDB for scheduler if scheduler itself is disabled.

I'm using topolvm with [official Helm chart](https://github.com/topolvm/topolvm/blob/main/charts/topolvm/README.md), and utilizing Storage Capacity Tracking instead of topolvm-scheduler ([doc](https://github.com/topolvm/topolvm/blob/b04d9c4584e7bcf9748ec29559026b65e0eab159/deploy/README.md#use-storage-capacity-tracking)).

Our `values.yaml` is set as below:

```yaml
controller:
  storageCapacityTracking:
    enabled: true
scheduler:
  enabled: false
webhook:
  podMutatingWebhook:
    enabled: false
```

After upgrading to 0.10.3 from 0.9.1, the Chart newly adds PDBs for topolvm-scheduler and topolvm-controller. However because my cluster has no topolvm-scheduler Pod, the PDB for scheduler is meaningless.

I'm getting `NoPods` event from controller manager once a day:

image

Yeah we can disable the PDB with `scheduler.podDisruptionBudget.enabled: false` (I did this as a workaround after this request), but it would be nice if the Chart would not put PDBs for non-existent Pods by default.

llamerada-jp:
issue: topolvm/topolvm#413

### DIFF
--- a/charts/topolvm/templates/scheduler/poddisruptionbudget.yaml
+++ b/charts/topolvm/templates/scheduler/poddisruptionbudget.yaml
@@ -1,5 +1,4 @@
-{{ if .Values.scheduler.enabled }}
-  {{ if .Values.scheduler.podDisruptionBudget.enabled }}
+{{ if and .Values.scheduler.enabled .Values.scheduler.podDisruptionBudget.enabled }}
 ---
 apiVersion: policy/v1beta1
 kind: PodDisruptionBudget
@@ -12,5 +11,4 @@ spec:
     matchLabels:
       app.kubernetes.io/name: topolvm-scheduler
 ---
-  {{ end }}
 {{ end }}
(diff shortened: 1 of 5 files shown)


charts/topolvm/templates/scheduler/poddisruptionbudget.yaml
```diff
@@ -1,4 +1,5 @@
-{{- if .Values.scheduler.podDisruptionBudget.enabled }}
+{{ if .Values.scheduler.enabled }}
+  {{ if .Values.scheduler.podDisruptionBudget.enabled }}
```
toshipp, line 2: I recommend using and function.
https://helm.sh/docs/chart_template_guide/function_list/#and

llamerada-jp: @toshipp 
Other files in templates/scheduler aren't used `and` function. I mimicked those files after thinking about it a bit. Would I rewrite all of the files using `and` function?

toshipp: Oh, I missed it. Other templates except for scheduler use `and` function rather than nesting `if`, I think it is better. So let's rewrite templates to use `and`

toshipp (approved):

satoru-takeuchi (approved):

PR #420, opened 2021-12-23
Add proposal for lvcreate options

lentzi90:
LVcreate options is a feature to allow passing options to `lvcreate`
when using TopoLVM.
For example, set the `alloc` flag or specify a `raid` configuration.
This proposal suggests adding a field `lvcreate-options` that would be
used to pass options directly to `lvcreate`.

This replaces part of #416.

### DIFF
--- a/docs/proposals/lvcreate-options.md
+++ b/docs/proposals/lvcreate-options.md
@@ -1,0 +1,116 @@
+# Allocation strategies
+
+## Summary
+
+LVCreate options is a feature to allow passing options to `lvcreate` when using TopoLVM.
+For example, set the `alloc` flag or specify a `raid` configuration.
+
+## Motivation
+
+LVM offers many options for how to create an LV (e.g. RAID configurations and mirroring).
+These options would be useful to expose through TopoLVM so users can make use of them.
+
+### Goals
+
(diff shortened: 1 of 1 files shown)


llamerada-jp:
Hi @lentzi90,

Thank you for your new proposal. We think Option A is better than B. The reason for this is like below.

- Fewer changes are required.
- Some options of lvcreate are depending on the structures of VG (like `--mirrors 2` you pointed). In option A, options for lvcreate and VG correspond 1:1. It looks easy to understand for administrators and is good to avoid problems.

About `stripe` and `stripe-size options`, We don't think we have to depreciate. I think it is better to annotate for the document of `lvcreate-options`.
About scheduling and multiple device-classes, We agree with your proposal. This means that no changes for these.

lentzi90:
@llamerada-jp great!
How should I move forward here? Do I need to make any changes to the proposal? Or should I wait for more reviews first?

llamerada-jp:
@lentzi90 
The other members have checked this proposal. And they agreed above comment. Would you please fill the `TODO`? I'll merge this PR after filling it. So, you can start implementing it.

toshipp (approved):
LGTM

llamerada-jp (approved):
LGTM

PR #436, opened 2022-02-04
Bump chart version to 4.0.1

peng225:


cupnes (approved):

llamerada-jp (changes requested):
[This PR](https://github.com/topolvm/topolvm/pull/427)'s goal is to stabilize the test but is also including changes for the helm template. So would you add the record of it for the changelog?

llamerada-jp (approved):

cupnes (approved):

PR #445, opened 2022-03-08
Override nodeSelector depending on the kind cluster name.

peng225:

### DIFF
--- a/e2e/Makefile
+++ b/e2e/Makefile
@@ -170,7 +170,7 @@ create-cluster: topolvm.img
 	$(HELM) repo add jetstack https://charts.jetstack.io
 	$(HELM) repo update
 	$(HELM) dependency build ../charts/topolvm/
-	$(HELM) install --namespace=topolvm-system topolvm ../charts/topolvm/ -f $(HELM_VALUES_FILE) $(HELM_SKIP_NODE_FINALIZE_FLAG)
+	$(HELM) install --namespace=topolvm-system topolvm ../charts/topolvm/ -f $(HELM_VALUES_FILE) --set "controller.nodeSelector.kubernetes\.io/hostname"=$(KIND_CLUSTER_NAME)-worker $(HELM_SKIP_NODE_FINALIZE_FLAG)
 
 .PHONY: test
 test: create-cluster


llamerada-jp (approved):

toshipp:
@peng225 Cloud you tell me the reason for the PR? I think the controller can be run anywhere.

toshipp:
Oh, we use nodeSelector, I get it.
https://github.com/topolvm/topolvm/blob/e00d2f66180b9058e3c78268ce095de34805291f/e2e/manifests/values/daemonset-scheduler.yaml#L11

e2e/Makefile
```diff
@@ -170,7 +170,7 @@ create-cluster: topolvm.img
 	$(HELM) repo add jetstack https://charts.jetstack.io
 	$(HELM) repo update
 	$(HELM) dependency build ../charts/topolvm/
-	$(HELM) install --namespace=topolvm-system topolvm ../charts/topolvm/ -f $(HELM_VALUES_FILE) $(HELM_SKIP_NODE_FINALIZE_FLAG)
+	$(HELM) install --namespace=topolvm-system topolvm ../charts/topolvm/ -f $(HELM_VALUES_FILE) --set "controller.nodeSelector.kubernetes\.io/hostname"=$(KIND_CLUSTER_NAME)-worker $(HELM_SKIP_NODE_FINALIZE_FLAG)
```
cupnes, line 173: Why is only the dot in ".io" escaped with a backslash? Doesn't the dot in "controller.nodeSelector.kubernetes" need to be escaped with a backslash?

peng225: @cupnes 
The definition of [yaml file](https://github.com/topolvm/topolvm/blob/main/e2e/manifests/values/deployment-scheduler.yaml) is as shown below.
```
controller:
  replicaCount: 1
  securityContext:
    enabled: false
  nodeSelector:
    kubernetes.io/hostname: topolvm-e2e-worker
```
First two dots are used to traverse the yaml tree, but the last dot is just a part of string "kubernetes.io/hostname".

cf. https://helm.sh/docs/intro/using_helm/#the-format-and-limitations-of---set

cupnes (approved):

PR #446, opened 2022-03-08
doc: Add design doc for thin-snapshots

Yuggupta27:
Add design document to allow the creation of VolumeSnapshots
for PVCs backed by thinly-provisioned logical volumes.

Based on: https://github.com/topolvm/topolvm/pull/442

Signed-off-by: Yug Gupta <yuggupta27@gmail.com>

### DIFF
--- a/docs/proposals/thin-snapshots-restore.md
+++ b/docs/proposals/thin-snapshots-restore.md
@@ -1,0 +1,248 @@
+# Snapshot & Restore
+
+> Note: Based on thin-provisioning of volumes. Please refer to [thin-provisioning proposal doc](thin-volumes.md).
+
+> Warning: Support for creation of VolumeSnapshots is available only for `thinly-provisioned` volumes.
+
+## Goals
+
+Allow users to create thin-snapshots from thinly-provisioned logical volumes.
+These thin-snapshots, then set the foundation for allowing creation of snapshots of persistent volume claims.
+
+* **Allow Snapshot Creation**: Enable users to take snapshots of PVs that are provisioned by TopoLVM.
+
+* **Restoring Snapshots**: Based on a volumesnapshot, a new PVC can be created that contains the contents of the snapshot.
+
+## Introduction
+
+### Thin-Snapshots
+
+* Snapshots of thin LVs are efficient because the data blocks common to a thin LV and its snapshot are shared.
+
+* As common data blocks are shared, A thin snapshot volume can reduce disk usage when there are multiple snapshots of the same origin volume.
+
+* The thin snapshot volume is independent of source volume deletion; and can be used independently just like a standalone volume.
+
+* These thin logical volumes can be activated as `Read-Only` copies of the source, and provided to kubernetes users as a VolumeSnapshotContent.
+
+### Restoring thin-snapshots
+
+* Thin snapshot volumes can be used as a logical volume origin for another snapshot. This allows for an arbitrary depth of recursive snapshots (snapshots of snapshots of snapshots ... )
+
+* Blocks common to recursive snapshots are also shared in the thin pool. There is no limit to or degradation from sequences of snapshots.
+
+* To provide a restored PVC to the user, we create another thin-provisioned snapshot of the available snapshot; and since lvm snapshots,  by design, are writable, so they can be provided to users to use as a volume for a new PVC.
+
+* This way, we also don't lose the identity of the snapshot after restoring; so we can restore a snapshot multiple times, to create multiple copies.
+
+* These thin snapshot volumes are activated as `Read-Write` copies of the source, and provided to kubernetes users as a PersistentVolume object.
+
+![](https://i.imgur.com/bwsK1Dl.png)
+
+> Note: The Snapshots, Restored volumes and Clones are all independent of their source deletion.
+
+## Flow of Operations for Snapshot Creation
+
+* A user creates a `VolumeSnapshot` custom resource; with the source set to the source PVC name.
+* The `snapshot-controller` which was watching the `VolumeSnapshot` object, sends a `Create VolumesnapshotContent` req to the api-server.
+* The `external-snapshotter` kubernetes sidecar receives this request, and then sends a `CreateSnapshot` gRPC call to TopoLVM CSI driver.
+* This `CreateSnapshot` request is received by the `topolvm-controller` which then creates a `LogicalVolume` CR with the source set to the name of Logical Volume of the source PVC.
+* `topolvm-node` on the target node reconciles the `LogicalVolume` CR created for the snapshot request.
+* `topolvm-node` sends a volume create request to lvmd to create a thin-snapshot from the thin-volume datasource.
+* `lvmd` creates an LVM snapshot logical volume as requested.
+* `topolvm-node` updates the status of `LogicalVolume`.
+* `topolvm-controller` finds the updated status of `LogicalVolume`.
+* `topolvm-controller` sends the success (or failure) to `external-snapshotter` sidecar.
+
+## LogicalVolume CR for Snapshots
+
+```yaml
+apiVersion: topolvm.cybozu.com/v1
+kind: LogicalVolume
+metadata:
+  name: snapcontent-b083470e-8293-47cc-810d-9561bd1754e6 # restore pv name
+  annotations:
+spec:
+  deviceClass: <string>
+  name: snapcontent-b083470e-8293-47cc-810d-9561bd1754e6
+  nodeName: 192.168.26.40
+  # 'source' specifies the `LogicalVolume` name of the source; if present.
+  # This field is populated only when `LogicalVolume` has a source.
+  source: <string>
+  # 'accessType' specifies how the user intents to consume the
+  # thin-snapshot logical volume.
+  # Set to "ro" when creating a snapshot and to "rw" when restoring a snapshot.
+  # This field is populated only when `LogicalVolume` has a source.
+  accessType: ro
+status:
+  volumeID: <The volume id of the lvm snapshot>
+```
+
+## Flow of Operations for Restore
+
+* An user creates a `PersistentVolumeClaim` custom resource with data source as a `VolumeSnapshot`
+* The `external-provisioner` kubernetes sidecar receives this request, and then sends a `CreateVolume` gRPC call to TopoLVM CSI driver.
+* This `CreateVolume` request containing a dataSource is received by the `topolvm-controller`.
+* From the `CreateVolume` req, validate if the volume creation request has a dataSource
+  * If `req.VolumeContentSource` is nil, proceed with normal volume creation
+  * Or, the Volume creation request might have a data source which is of type Snapshot or Persistent Volume.
+
+    * i.e., `volumeSource.Type.(type)` can be
+      * `VolumeContentSource_Snapshot`: for PVC-Restore from snapshot.
+      * `VolumeContentSource_Volume`: in case of PVC-PVC clone request.
+
+* Fetch the Logical Volume corresponding to the source.
+* Since we have the source volume now, pass it to the `lvservice.CreateVolume`; so that we can populate the `LogicalVolume` CRD with the same parameters as source:
+   <!-- snapshot add -->
+
+  ```yaml
+  Spec: topolvmv1.LogicalVolumeSpec{
+            Name:        pvcName,
+            # The restored lv must be created on the same node as source.
+            NodeName:    sourceVolume.Spec.NodeName,
+            # restored lv created on the same deviceClass as the source.
+            DeviceClass: sourceVolume.Spec.DeviceClass,
+            # the size of the restored logical volume must be the same as the source.
+            Size:        sourceVolume.Spec.Size,
+            # 'source' specifies the logicalvolume name of the source; if present.
+            Source:      <string>
+            # Set to "ro" when creating a snapshot and to "rw" when restoring a snapshot.
+            accessType:  rw
+        },
+  ```
+  This will make sure that the **Restored volumes are placed on the same node and the same DeviceClass as the source**.
+
+* `topolvm-node` on the target node reconciles the `LogicalVolume` CR.
+* `topolvm-node` sends a volume create request to lvmd to create a thin-snapshot from the thin-volume datasource.
+* `lvmd` creates an LVM logical volume as requested.
+* `lvmd` activates the thin logical volume with `Read-Write` permissions.
+* `topolvm-node` updates the status of `LogicalVolume`.
+* `topolvm-controller` finds the updated status of `LogicalVolume`.
+* `topolvm-controller` sends the success (or failure) to `provisioner` sidecar.
+
+
+## LogicalVolume CR for Restore operation
+
+```yaml
+apiVersion: topolvm.cybozu.com/v1
+kind: LogicalVolume
+metadata:
+  name: pvc-42041aee-79a2-4184-91aa-5e1df6068b9f # restore pv name
+  annotations:
+spec:
+  deviceClass: <string>
+  name: pvc-42041aee-79a2-4184-91aa-5e1df6068b9f
+  nodeName: 192.168.26.40
+  # 'source' specifies the `LogicalVolume` name of the source; if present.
+  # This field is populated only when `LogicalVolume` has a source.
+  source: <string>
+  # 'accessType' specifies how the user intents to consume the
+  # thin-snapshot logical volume.
+  # Set to "ro" when creating a snapshot and to "rw" when restoring a snapshot.
+  # This field is populated only when `LogicalVolume` has a source.
+  accessType: rw
+status:
+  volumeID: <Store the volume id of the lvm snapshot as same as existing LV.>
+```
+
+## LV Operations
+
+* Snapshot creation from a thinly-provisioned volume
+    ```bash
+    $ lvcreate -s --name thinsnap VG0/thinvolume
+    ```
+
+* Activation of a logical volume
+
+  A thinly-provisioned snapshot will be activated as Read-Only for snapshots and Read-Write for Restore scenarios.
+
+  For e.g: Activating a logical volume for Read-Write operations:
+
+  ```bash
+  $ lvchange -ay -K VG0/thinsnap
+  ```
+
+* Restoring a Snapshot
+
+
+  ```bash
+  $ lvcreate -s --name thin-restored VG0/thinsnap
+  ```
+
+## VolumeSnapshotClass and VolumeSnapshot
+
+### VolumeSnapshotClass
+
+```yaml
+apiVersion: snapshot.storage.k8s.io/v1
+kind: VolumeSnapshotClass
+metadata:
+  name: topolvm-snapclass
+driver: topolvm.cybozu.com
+deletionPolicy: Delete
+```
+
+### VolumeSnapshot
+
+```yaml
+apiVersion: snapshot.storage.k8s.io/v1
+kind: VolumeSnapshot
+metadata:
+  name: new-snapshot
+spec:
+  volumeSnapshotClassName: topolvm-snapclass
+  source:
+    persistentVolumeClaimName: snapshot-pvc
+```
+
+## Restoring the Snapshot to a new PVC
+
+When creating a snapshot or restoring it to a new volume; both the snapshot and the restored volume need to be provisioned
+on the same node as the source volume.
+
+To make sure the restored volume is provisioned on the same node as that of the source, we use node selectors to select the same
+node for provisioning.
+
+To schedule the restored pod on the same node, the restored pod has the node affinity set for the node on which source PVC is provisioned.
+
+`pod-restore.yaml`
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: pod-restore
+spec:
+# The Restored PVC needs to be provisioned on the same node as the source volume.
+  affinity:
+    nodeAffinity:
+      requiredDuringSchedulingIgnoredDuringExecution:
+        nodeSelectorTerms:
+        - matchExpressions:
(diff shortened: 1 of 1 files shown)


llamerada-jp:
Thank you for your proposal. We will take a moment for reviewing it.

llamerada-jp:
@Yuggupta27 cc @sp98 

Could you please confirm the following issues?

- The `snapshot` and `clone` are different features I think. I want to handle those features in separate proposals if possible. Is the `clone` feature included in the dev freeze mentioned below?
https://topolvm.slack.com/archives/CT0AS6UQ1/p1647251609636729?thread_ts=1646817356.890029&cid=CT0AS6UQ1
- I think the scope of this proposal is only the `local snapshot/clone` feature. If this is the case, could you please write this?
- We can only create and restore `snapshot` and `clone` in the same node of the source volume without the remote data copy. Could you please additionally write that we need to change the scheduler to realize it.
- There is [the proposal of snapshot not using thin volume](https://github.com/topolvm/topolvm/pull/385). And remote copy feature may be considered in the future. So I want to avoid making those impossible in this proposal.

Yuggupta27:
Thanks for the review @llamerada-jp !

> @Yuggupta27 cc @sp98
> 
> Could you please confirm the following issues?
> 
> * The snapshot and clone are different features I think. I want to handle those features in separate proposals if possible. Is the clone feature included in the dev freeze mentioned below?

Yes. The snapshot and clone are separate features. Added a separate design doc for the [PVC-PVC clone feature](https://github.com/topolvm/topolvm/pull/447). 

> Is the clone feature included in the dev freeze mentioned below?
>   https://topolvm.slack.com/archives/CT0AS6UQ1/p1647251609636729?thread_ts=1646817356.890029&cid=CT0AS6UQ1

Yes, both Snapshots and Clones features are also targeted for the same deadlines as thin-lv feature.

> * I think the scope of this proposal is only the local snapshot/clone feature. If this is the case, could you please write this?
> * We can only create and restore snapshot and clone in the same node of the source volume without the remote data copy. Could you please additionally write that we need to change the scheduler to realize it.

The Snapshots, Restored volumes from the snapshots, and Volume Clones(PVC-PVC clones); will all be present in the same node as the parent volume. Will mention the same in the doc as well :+1: 

> * There is [the proposal of snapshot not using thin volume](https://github.com/topolvm/topolvm/pull/385). And remote copy feature may be considered in the future. So I want to avoid making those impossible in this proposal.

Sure! The thin-lv snapshot design was proposed with keeping thick snapshot support also in mind by adding `type` parameter in the LogicalVolume CR for snapshots. Will love to have suggestions on improving the design and making it future-ready!

satoru-takeuchi (changes requested):
FYI, @llamerada-jp will also comment later.

llamerada-jp (changes requested):
- Could you please create a new section to write that how PV/PVC using snapshot as a datasource to place at the same node to the original PV/PVC.
- Do you know can we restore the snapshot using different StorageClass of source PV/PVC? I counld't find it in KEP or other documents. If it is allowed, the user may point to a different device-class from source PV/PVC. I believe that this issue has a significant impact on this design.
- Do you want to select the device-class to which the snapshot will be created in this proposal? If so, would you please add an example of `VolumeSnapshotClass` to describe how this should be specified?

docs/proposals/thin-snapshots-restore.md
```diff
@@ -0,0 +1,202 @@
...
+
+## Proposed LogicalVolume CR for Snapshots
+
+```yaml
+apiVersion: topolvm.cybozu.com/v1
+kind: LogicalVolume
+metadata:
+  name: snapcontent-b083470e-8293-47cc-810d-9561bd1754e6 # restore pv name
+  annotations:
+spec:
```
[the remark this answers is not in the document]
Yuggupta27: Nice catch! that was an overlook from my side, updated now :+1:

docs/proposals/thin-snapshots-restore.md
```diff
@@ -0,0 +1,202 @@
...
+
+## Additional Notes
+
+* When creating a thin snapshot volume, you do not specify the size of the volume.
+
+* Thin snapshot volumes are not supported across the nodes in a cluster. The snapshot volume must be exclusively activated on only one cluster node.
+## Open Questions
+
+- Should we only allow thick snapshot creation for thick volumes and thin-snapshots for thinly-provsioned volumes?
+  - If yes, then we don't need to specify `type` parameter for snapshot volumes in the LogicalVolume CR; as we can assume thin-snapshots if the parent is a thinly-provisioned volume.
```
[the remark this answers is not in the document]
Yuggupta27: Sure, we can keep the `type` parameter

Yuggupta27:
> * Could you please create a new section to write that how PV/PVC using snapshot as a datasource to place at the same node to the original PV/PVC.

Thanks for the suggestion! Added a section for the flow of operations in case of a Restore Volume operation [here](https://github.com/topolvm/topolvm/pull/446/files#diff-552a2fdc08796a742045a5300895bdad82bd500a67da333c995a528fa193b872R106).

> * Do you know can we restore the snapshot using different StorageClass of source PV/PVC? I counld't find it in KEP or other documents. If it is allowed, the user may point to a different device-class from source PV/PVC. I believe that this issue has a significant impact on this design.

Restoring to different storageclass should not be allowed from our side as we intend to create a copy of the parent pvc, which involves having the same parameters as the parent.
Although it is something we can look into in the future, but for now, we should allow the clone/restored pvc to be of the same storageclass as the parent.

> * Do you want to select the device-class to which the snapshot will be created in this proposal? If so, would you please add an example of VolumeSnapshotClass to describe how this should be specified?

Added examples of both VolumeSnapshotClass and VolumeSnapshot. As suggested by you, we can have a field in VolumeSnapshotClass `"topolvm.cybozu.com/restore-mode": "local"` which specifies the Snapshots here is created on the local node, and it also opens up the possibility of having remote snapshots(in different node) in future!

llamerada-jp:
> Restoring to different storageclass should not be allowed from our side as we intend to create a copy of the parent pvc, which involves having the same parameters as the parent.
Although it is something we can look into in the future, but for now, we should allow the clone/restored pvc to be of the same storageclass as the parent.

OK. Would you please write this limitation to this proposal?

> Added examples of both VolumeSnapshotClass and VolumeSnapshot. As suggested by you, we can have a field in VolumeSnapshotClass "topolvm.cybozu.com/restore-mode": "local" which specifies the Snapshots here is created on the local node, and it also opens up the possibility of having remote snapshots(in different node) in future!

Is it correct that in this proposal, snapshots will be created only for "the same device-class" on the same node? If it is, I would like to write that to be clear.

Yuggupta27:
> > Restoring to different storageclass should not be allowed from our side as we intend to create a copy of the parent pvc, which involves having the same parameters as the parent.
> > Although it is something we can look into in the future, but for now, we should allow the clone/restored pvc to be of the same storageclass as the parent.
> 
> OK. Would you please write this limitation to this proposal?
> 
> > Added examples of both VolumeSnapshotClass and VolumeSnapshot. As suggested by you, we can have a field in VolumeSnapshotClass "topolvm.cybozu.com/restore-mode": "local" which specifies the Snapshots here is created on the local node, and it also opens up the possibility of having remote snapshots(in different node) in future!
> 
> Is it correct that in this proposal, snapshots will be created only for "the same device-class" on the same node? If it is, I would like to write that to be clear.

Thanks, will mention that :+1:

Yuggupta27:
@llamerada-jp @satoru-takeuchi @cupnes which of the proposed approaches mentioned in the doc would you guys suggest for the LogicalVolume CRs in case of snapshots?

llamerada-jp:
@Yuggupta27 
About scheduler, the following methods were listed internally by us. Would you please consider which options are better in your case? Or could you let me know if you get any other ideas?
1. The users select the same node as the original PV by the node selector. It is easiest to implement but it takes labor for users.
2. Set node selector by mutating webhook. It is useable on public k8s services like EKS/GKE that are not supported scheduler extenders. It may conflict with CD software such as ArgoCD and other operators. So I think requires validation.
3. Update topolvm-scheduler. It may have the highest discuss and  implementation costs. And it is not usable EKS/GKE etc. We want to avoid this option.

Yuggupta27:
> @Yuggupta27 About scheduler, the following methods were listed internally by us. Would you please consider which options are better in your case? Or could you let me know if you get any other ideas?

Hi @llamerada-jp , So setting the Logical Volume CR values to the same as that of parent(NodeName and deviceClass) [here](https://github.com/topolvm/topolvm/pull/446/files#diff-552a2fdc08796a742045a5300895bdad82bd500a67da333c995a528fa193b872R128-R131) will not be enough for asking for lv creation on a particular node(so that it can be on same node as parent)?
For e.g 
```yaml
  },
  Spec: topolvmv1.LogicalVolumeSpec{
            Name:        pvcName,
            # The cloned lv created on the same node as parent.
            NodeName:    sourceVolume.Spec.NodeName,
            # cloned lv created on the same deviceClass as the parent.
            DeviceClass: sourceVolume.Spec.DeviceClass,
            # the size of the restored logical volume is same as the parent.
            Snapshot: {
                type: thin
                dataSource: <string>
                accessType: rw
            }
        },
  ```

Here, `sourceVol` refers to the parent Logical Volume

**Comment by llamerada-jp** [comment]
When `CreateVolume` gRPC method of topolvm-controller called, 
`accessibility_requirements` parameter is passed and it contains the selectable nodes topology list. LogicalVolume CR will be created after `CreateVolume` is called. So I guess we may consider scheduling.
https://github.com/topolvm/topolvm/blob/f76b553b2d0bdf2ac71d009a4cd0ec36cd9aea5d/driver/controller.go#L86-L122

I am not confident in this parameter and will survey it a little.

**Comment by Yuggupta27** [comment]
As per my current understanding, when TopoLVM receives a CreateVolume request, we can check if the request has a `Datasource`.
Any CreateVolume request, that is regarding Restoring a snapshot or Creating a clone, has a `dataSource` parameter set in the request; which we can access via:   ` req.VolumeContentSource`. 

In case a volume has a `dataSource`( i.e. `req.VolumeContentSource != nil` ), we can ignore the scheduling intelligence and directly attempt to create the volume on the same node as the parent(as it will fail otherwise).
Also, if the node does not have enough space, we can let the request fail with appropriate error messages.

Apart from Restored or Cloned volumes, scheduling intelligence will still behave as intended for all the other volumes.

Thoughts?

**Comment by llamerada-jp** [comment]
The CSI driver's behavior is specified on CSI spec. When `accessible_topology` is specified, TopoLVM must obey this parameter. I think your assumed flow is correct if `accessible_topology` contains the target node or there are no constraints. We should control the `accessible_topology` parameter before the `CreateVolume` method will be called using some scheduler. I am concerned about the mechanisms for realizing it.

Would you please refer to the note of `TopologyRequirement` definitions?
https://github.com/container-storage-interface/spec/blob/b09ca9e903260a0d715c6cb7766417f47b0057c6/spec.md#createvolume

File: docs/proposals/thin-snapshots-restore.md [hunk 3]
```diff
@@ -0,0 +1,232 @@
...
+
+## Deployment Changes
+
+* Add `external-snapshotter` sidecar.
+
+* Perform `snapshot-controller` deployment.
+
+## Metrics and Monitoring
+
+Should be already covered during adding support for `thin` logical volumes. See [thin-provisioning proposal doc](https://github.com/topolvm/topolvm/pull/442).
```
cupnes, line 222: PR #442 has been merged, please change the URL.
```suggestion
Should be already covered during adding support for `thin` logical volumes. See [thin-provisioning proposal doc](docs/proposals/thin-volumes.md).
```

cupnes (changes requested):
Please correct just a few details. At the moment, I don't think there are any other problems.

docs/proposals/thin-snapshots-restore.md
Yuggupta27: Thanks, updated :+1:

docs/proposals/thin-snapshots-restore.md
```diff
@@ -0,0 +1,232 @@
...
+![](https://i.imgur.com/bwsK1Dl.png)
+
+> Note: The Snapshots, Restored volumes and Clones are all independent of their parent deletion.
+
+## Flow of Operations for Snapshot Creation
+
+* An user creates a `VolumeSnapshot` custom resource.
+* The `snapshot-controller` which was watching the `VolumeSnapshot` object, sends a `Create VolumesnapshotContent` req to the api-server.
+* The `external-snapshotter` kubernetes sidecar receives this request, and then sends a `CreateSnapshot` gRPC call to TopoLVM CSI driver.
+* This `CreateSnapshot` request is received by the `topolovm-controller` which then create a `LogicalVolume` CR.
```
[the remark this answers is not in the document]
Yuggupta27: Nice catch! Updated now :+1:

Yuggupta27:
> Please correct just a few details. At the moment, I don't think there are any other problems.

Thanks for the suggestions! Updated now :+1:

docs/proposals/thin-snapshots-restore.md
```diff
@@ -0,0 +1,231 @@
+# Snapshot & Restore
+
+> Note: Based on thin-provisioning of volumes. Please refer to [thin-provisioning proposal doc](https://github.com/topolvm/topolvm/pull/442).
```
cupnes, line 3: Please correct this as well.
```suggestion
> Note: Based on thin-provisioning of volumes. Please refer to [thin-provisioning proposal doc](docs/proposals/thin-volumes.md).
```

cupnes (changes requested):
Thank you. But please make one more correction.

docs/proposals/thin-snapshots-restore.md
Yuggupta27: Great catch! Updated :+1:

cupnes:
@Yuggupta27 Thank you. LGTM. Please confirm @llamerada-jp's comment.

llamerada-jp:
@Yuggupta27 
Would you add a limitation or node that the user needs to specify the appropriate node like below when restoring PVC, as we have talked about before?
https://topolvm.slack.com/archives/CT0AS6UQ1/p1648447741324199?thread_ts=1646817356.890029&cid=CT0AS6UQ1

llamerada-jp:
@Yuggupta27 
> We can later perform improvements on the same in follow-up PRs if deemed necessary.

I agree with your comment. For this reason, I would like you to update the proposal as follows point.
- The scheduler does not update and the user needs to specify the node when creating the pod currently.
- An example of how to specify the node as you commented.

Yuggupta27:
> * The scheduler does not update and the user needs to specify the node when creating the pod currently.

Thanks for the suggestion! Updated the point and added the example for the same :+1:

nbalacha (changes requested):

Yuggupta27:
> I agree with your comment. For this reason, I would like you to update the proposal as follows point.
> 
> * The scheduler does not update and the user needs to specify the node when creating the pod currently.
> * An example of how to specify the node as you commented.

Thanks for confirming @llamerada-jp ! Updated the PR with the details. 

>@Yuggupta27 Thank you. LGTM. Please confirm @llamerada-jp's comment.

@cupnes  :+1:

llamerada-jp:
@Yuggupta27 
LGTM
I'm going to wait to merge this PR until merging the PR for implementation(#463), is that ok?

Yuggupta27:
> @Yuggupta27 LGTM I'm going to wait to merge this PR until merging the PR for implementation(#463), is that ok?

Sure! Thanks!

docs/proposals/thin-snapshots-restore.md
```diff
@@ -0,0 +1,291 @@
...
+In the `pod-restore.yaml`, we can schedule the restore pod on the same node as the source using the source pod name as the label for pod affinity.
+Here, before creating the pod, an user just has to mention the name of the source pvc pod.
+
+For example:
+
+Source application pod uses the label
+`[app.kubernetes.io/name](http://app.kubernetes.io/name): my-pod`.
+
+To schedule the restored pod on the same node,
+the restored pod has the pod affinity set for te source PVC label.
```
cupnes, line 231: typo
```suggestion
the restored pod has the pod affinity set for the source PVC label.
```

cupnes (changes requested):
Please make one correction.

docs/proposals/thin-snapshots-restore.md
Yuggupta27: Great catch! Updated :+1:

docs/proposals/thin-snapshots-restore.md
```diff
@@ -0,0 +1,274 @@
...
+
+Allow users to create thin-snapshots from thinly-provisioned logical volumes.
+These thin-snapshots, then set the foundation for allowing creation of snapshots of persistent volume claims.
+
+* **Allow Snapshot Creation**: Enable users to take snapshots of PVs that are provisioned by TopoLVM.
+
+* **Restoring Snapshots**: Based on a volumesnapshot, a new PVC can be created that contains the contents of the snapshot.
+
+## Introduction
+
```
nbalacha, line 15: Please add  a section for Limitations.

nbalacha (changes requested):

docs/proposals/thin-snapshots-restore.md
```diff
@@ -0,0 +1,274 @@
+# Snapshot & Restore
+
+> Note: Based on thin-provisioning of volumes. Please refer to [thin-provisioning proposal doc](thin-volumes.md).
```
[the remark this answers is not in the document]
Yuggupta27: Added a warning at the beginning of the doc and a limitation section

docs/proposals/thin-snapshots-restore.md
```diff
@@ -0,0 +1,274 @@
...
+> Note: The Snapshots, Restored volumes and Clones are all independent of their parent deletion.
+
+## Flow of Operations for Snapshot Creation
+
+* A user creates a `VolumeSnapshot` custom resource.
+* The `snapshot-controller` which was watching the `VolumeSnapshot` object, sends a `Create VolumesnapshotContent` req to the api-server.
+* The `external-snapshotter` kubernetes sidecar receives this request, and then sends a `CreateSnapshot` gRPC call to TopoLVM CSI driver.
+* This `CreateSnapshot` request is received by the `topolvm-controller` which then create a `LogicalVolume` CR.
+* `topolvm-node` on the target node finds the earlier created `LogicalVolume` CR.
+* `topolvm-node` sends a volume create request to lvmd to create a thin-snapshot from the thin-volume datasource.
```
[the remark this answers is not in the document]
Yuggupta27: I believe we can leave this one as is as it already mentions ` from the thin-volume datasource`.

satoru-takeuchi (changes requested):

docs/proposals/thin-snapshots-restore.md
```diff
@@ -0,0 +1,276 @@
...
+apiVersion: v1
+kind: Pod
+metadata:
+  name: pod-restore
+spec:
+# The Restored PVC needs to be provisioned on the same node as the source volume.
+  affinity:
+    podAffinity:
+      requiredDuringSchedulingIgnoredDuringExecution:
+    * labelSelector:
```
cupnes, line 246:
```suggestion
      - labelSelector:
```

docs/proposals/thin-snapshots-restore.md
```diff
@@ -0,0 +1,276 @@
...
+          matchExpressions:
+        * key:  [app.kubernetes.io/name](http://app.kubernetes.io/name)
```
cupnes, line 248:
```suggestion
            - key: app.kubernetes.io/name
```

docs/proposals/thin-snapshots-restore.md
```diff
@@ -0,0 +1,276 @@
...
+```
+
+## Restoring the Snapshot to a new PVC
+
+When creating a snapshot or restoring it to a new volume; both the snapshot and the restored volume need to be provisioned
+on the same node as the source volume.
+
+Since the a snapshots can only be taken of a BOUND PVC, and volumeBinding is `waitforfirstconsumer`; it is safe to
+assume that the source PVC already has an application POD using it. So while creating a restored PVC from the snapshot of source, we need
+to schedule the pod/app on the same node. To do that, we can make use of pod affinity.
```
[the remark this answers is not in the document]
Yuggupta27: Updated :+1:

docs/proposals/thin-snapshots-restore.md
```diff
@@ -0,0 +1,274 @@
...
```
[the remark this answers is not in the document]
Yuggupta27: Thanks for the suggestion. True, although POD Affinity works is convenient if the source PVC has an application using it, but Node Affinity covers all the use cases.
Updated to use node affinity now

docs/proposals/thin-snapshots-restore.md
```diff
@@ -0,0 +1,274 @@
...
+    persistentVolumeClaimName: snapshot-pvc
+```
+
+## Restoring the Snapshot to a new PVC
+
+When creating a snapshot or restoring it to a new volume; both the snapshot and the restored volume need to be provisioned
+on the same node as the source volume.
+
+Since the a snapshots can only be taken of a BOUND PVC, and volumeBinding is `waitforfirstconsumer`; it is safe to
+assume that the source PVC already has an application POD using it. So while creating a restored PVC from the snapshot of source, we need
```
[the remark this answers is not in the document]
Yuggupta27: Thanks for the suggestion! Updated to use Node selectors

cupnes (approved):

satoru-takeuchi (approved):

nbalacha (approved):

npu21:
hi @Yuggupta27 ,Is there a release plan for this feature?

PR #448, opened 2022-03-17
thinp: Implementation of Thin Logical Volumes support

leelavg:
Pending:

from proposal:
~- https://github.com/topolvm/topolvm/pull/442#discussion_r828765950 inputs about the implementation is needed~ (finalised in same thread)
~- https://github.com/topolvm/topolvm/pull/442#discussion_r828706602 inputs about the final decision~ (finalised in https://github.com/topolvm/topolvm/pull/442#issuecomment-1081595993)

general (below will be part of another PR):
- no regressions as per unit tests and need to look into e2e runs from CI and add new e2e tests wrt thinpool
- document updates about thinpool
~- decisions/inputs on TODO, most of those doesn't effect the functionality though~

Minor Tracker:
- Revisit https://github.com/topolvm/topolvm/pull/448#discussion_r838056977 after Snapshot feature is implemented

ps: pls review by commit bases, I tried to segregate them as a logical unit

ref: https://github.com/topolvm/topolvm/pull/442


leelavg:
@llamerada-jp 
- regarding the movement of cards for this PR, this is actually re-viewable and only reason I moved it to draft is to make sure that there are no regressions in e2e
- if the team doesn't review draft PR I'll remove that status.

llamerada-jp:
@leelavg 
We check progress and move the cards in our daily meetings. Would you please change the state to `Ready for review` after passing the CI? We will assign reviewers.

leelavg:
- ack, facing some issues with running CI in local environment
- many complaints from kind and minikube while running e2e in fedora machines
- if we create a new ubuntu machine and install golang & docker in it, rest of the env setup will be taken care by make targets right?

llamerada-jp:
> if we create a new ubuntu machine and install golang & docker in it, rest of the env setup will be taken care by make targets right?
Yes, it is. The document and configs for CI will help you, I hope.
https://github.com/topolvm/topolvm/tree/main/e2e#how-to-run-tests
https://github.com/topolvm/topolvm/blob/main/.github/workflows/e2e-k8s.yaml

leelavg:
- created a new ubuntu VM
- installed golang and docker via snap
- facing below issue
```
 ✗ Preparing nodes 📦 📦 📦 📦  
ERROR: failed to create cluster: docker run error: command "docker run --hostname topolvm-e2e-worker --name topolvm-e2e-worker --label io.x-k8s.kind.role=worker --privileged --security-opt seccomp=unconfined --security-opt apparmor=unconfined --tmpfs /tmp --tmpfs /run --volume /var --volume /lib/modules:/lib/modules:ro --detach --tty --label io.x-k8s.kind.cluster=topolvm-e2e --net kind --restart=on-failure:1 --init=false --volume=/dev:/dev --volume=/tmp/topolvm/lvmd1:/tmp/topolvm --volume=/tmp/topolvm/worker1:/var/lib/kubelet:rshared kindest/node:v1.23.3" failed with error: exit status 125
Command Output: 3d7e4ca1b3056d4489d973d450e9a95a47c15911ac2e2648573bff9173bffeab
docker: Error response from daemon: path /tmp/topolvm/worker1 is mounted on /tmp but it is not a shared mount.
Makefile:89: recipe for target 'launch-kind' failed
make[1]: *** [launch-kind] Error 1
make[1]: Leaving directory '/home/ubuntu/topolvm/e2e'
Makefile:163: recipe for target 'create-cluster' failed
make: *** [create-cluster] Error 2
```

leelavg:
@llamerada-jp I ran the failed test locally and could see it's passing locally. Picked k8s version 1.22.2 as it has the most passed matrix in https://github.com/topolvm/topolvm/actions/runs/2016111597
```
KUBERNETES_VERSION='1.22.2' TEST_SCHEDULER_MANIFEST='daemonset' STORAGE_CAPACITY='false' SKIP_NODE_FINALIZE='false' make test
Ran 72 of 112 Specs in 2198.174 seconds
SUCCESS! -- 72 Passed | 0 Failed | 1 Pending | 39 Skipped
PASS

Ginkgo ran 1 suite in 41m20.211613736s```

**Comment by llamerada-jp** [comment]
@leelavg 
I tried to reproduce this case like below locally, but I couldn't.

```
KUBERNETES_VERSION='1.21.2' TEST_SCHEDULER_MANIFEST='daemonset' STORAGE_CAPACITY='false' SKIP_NODE_FINALIZE='false' make test
```

I re-ran the CI.

File: lvmd/device_class_manager.go [hunk 1]
```diff
@@ -11,7 +11,13 @@ import (
 // ErrNotFound is returned when a VG or LV is not found.
 var ErrNotFound = errors.New("device-class not found")
 
-const defaultSpareGB = 10
+type LVType string
```
llamerada-jp, line 14: This type is used to specify the type of VG or pool, it is not for LV, I guess. Is `DeviceType` better?

lvmd/device_class_manager.go
```diff
@@ -36,6 +51,10 @@ type DeviceClass struct {
 	StripeSize string `json:"stripe-size"`
 	// LVCreateOptions are extra arguments to pass to lvcreate
 	LVCreateOptions []string `json:"lvcreate-options"`
+	// Type is the name of logical volume target, supports 'thick' (default) or 'thin' currently
+	Type LVType `json:"type,omitempty"`
+	// ThinPoolConfig holds the configuration for thinpool in this volume group corresponding to the device-class
+	ThinPoolConfig *ThinPoolConfig `json:"thin-pool,omitempty"`
```
llamerada-jp, line 57:
```suggestion
	// Type is the name of logical volume target, supports 'thick' (default) or 'thin' currently
	Type LVType `json:"type"`
	// ThinPoolConfig holds the configuration for thinpool in this volume group corresponding to the device-class
	ThinPoolConfig *ThinPoolConfig `json:"thin-pool"`
```

lvmd/lvservice.go
```diff
@@ -41,7 +42,28 @@ func (s *lvService) CreateLV(_ context.Context, req *proto.CreateLVRequest) (*pr
 		return nil, err
 	}
 	requested := req.GetSizeGb() << 30
-	free, err := vg.Free()
+	free := uint64(0)
+	pool := &command.ThinPool{}
```
llamerada-jp, line 46:
```suggestion
	var pool *command.ThinPool
```

llamerada-jp (commented):
I reviewed excluding metrics and test.
I noticed that `VolumeGroup.ListVolumes` seeks volumes including in some pools. I guess this logic list up extra volumes. Would you plaease update that function?

lvmd/command/lvm.go
```diff
@@ -120,7 +121,8 @@ func parseOutput(cmd, fields string, args ...string) ([]LVInfo, error) {
 
 // VolumeGroup represents a volume group of linux lvm.
 type VolumeGroup struct {
-	name string
+	name  string
+	ratio *float64 // overprovision ratio of thinpool
```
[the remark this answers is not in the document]
leelavg: - before performing any thin pool operations we set this value based on the thinpoolconfig in deviceClass
- afaik no rpc in vg/lv service mutates/uses this in parallel

lvmd/command/lvm.go
```diff
@@ -415,14 +438,88 @@ func (t *ThinPool) ListVolumes() ([]*LogicalVolume, error) {
...
+			return volume, nil
+		}
+	}
+	return nil, ErrNotFound
+}
+
 // CreateVolume creates a thin volume from this pool.
-func (t *ThinPool) CreateVolume(name string, size uint64) (*LogicalVolume, error) {
-	if err := CallLVM("lvcreate", "-T", t.fullname, "-n", name, "-V", fmt.Sprintf("%vg", size>>30)); err != nil {
+func (t *ThinPool) CreateVolume(name string, size uint64, tags []string, stripe uint, stripeSize string, lvcreateOptions []string) (*LogicalVolume, error) {
```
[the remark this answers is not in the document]
leelavg: - sry, didn't get this, by tagging it `memo`, you are just noting down this info or want me to change anything?

lvmd/command/lvm.go
```diff
@@ -415,14 +438,88 @@ func (t *ThinPool) ListVolumes() ([]*LogicalVolume, error) {
...
+
+	if err := CallLVM("lvcreate", lvcreateArgs...); err != nil {
 		return nil, err
 	}
 	return t.vg.FindVolume(name)
 }
 
+// Free on a thinpool returns used data, metadata percent and free space
+// with and without overprovisioning
+func (t *ThinPool) Free() (ThinPoolUsage, error) {
```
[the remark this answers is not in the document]
leelavg: - initially, I also thought of returning pointer however felt the data is small and creating a struct in the function and passing it out of the function is less ideal
- while returning pointer, the backing data might skip the stack
- however let me know if that's not the case or `return nil, err` is still better

lvmd/command/lvm.go
```diff
@@ -415,14 +438,88 @@ func (t *ThinPool) ListVolumes() ([]*LogicalVolume, error) {
...
+			if err != nil {
+				return tpu, err
+			}
+			virtualSize += lvSize
+		}
+	}
+
+	free := 1.0 - (tpu.DataPercent+tpu.MetadataPercent)/100
+	tpu.FreeBytes = uint64(math.Ceil(float64(free) * float64(tpu.SizeBytes)))
+	tpu.OverprovisionBytes = t.sizeLimit - virtualSize
```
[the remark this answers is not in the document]
leelavg: - ah! makes sense.
- will it be fine if I could just do this operation in the call site itself? In that way, setting the ratio on VolumeGroup struct can also be removed i think.

lvmd/device_class_manager.go
leelavg: ack, sounds good.

lvmd/device_class_manager.go
leelavg: - ack, we might get null values, is that fine?

lvmd/lvservice.go
leelavg: ack

leelavg (commented):
- addressed comments, should I wait for one more review or can I go ahead and fix the code?

leelavg:
> I guess this logic list up extra volumes. 
@llamerada-jp yes, it does list thin logical volumes as well, so filtering them out in call site (`GetLVList`) if the request is on volumegroup by below logic (for thinlv poolname will be set).
``` go
		if dc.Type == TypeThick && lv.GetPoolName() != "" {
			// do not send thin lvs if request is on TypeThick
			continue
		}
```

lvmd/command/lvm.go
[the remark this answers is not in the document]
llamerada-jp: I think so.
What do you think about it? @cupnes @satoru-takeuchi

lvmd/command/lvm.go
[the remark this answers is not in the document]
llamerada-jp: It is just only memo for me. No changes are required.

lvmd/device_class_manager.go
llamerada-jp: IMO, `omitempty` option effects when converting this struct to string. I guess there is no such code.

lvmd/command/lvm.go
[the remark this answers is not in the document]
cupnes: I agree with @llamerada-jp.

lvmd/command/lvm.go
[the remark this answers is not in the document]
llamerada-jp: Strictly, your way better performance can be expected. But I think there is no performance problem in this case. And we alined to not return a halfway value in an error.

llamerada-jp:
It's OK to fix code, exclude about `overprovision-ratio`.

> so filtering them out in call site (GetLVList) if the request is on volumegroup by below logic (for thinlv poolname will be set).

I'll check it again.

leelavg:
Cool! Now the e2e also has cleanly passed with no regressions, so the existing functionality isn't effected 😀

lvmd/device_class_manager.go
```diff
@@ -72,11 +91,34 @@ func ValidateDeviceClasses(deviceClasses []*DeviceClass) error {
...
+				return fmt.Errorf("thinpool name should not be empty: %s", dc.Name)
+			}
+
+			if dc.ThinPoolConfig.OverprovisionRatio == nil || *dc.ThinPoolConfig.OverprovisionRatio < 1.0 {
+				return fmt.Errorf("overprovision for thin pool %s in device class %s should be greater than 1.0", dc.ThinPoolConfig.Name, dc.Name)
+			}
+			// combination of volumegroup and thinpool should be unique across device classes
+			// so the key 'name' shouldn't appear twice to verify it's uniqueness
+			name = name + "/" + dc.ThinPoolConfig.Name
+		}
```
llamerada-jp, line 114: Could you please add a check for dc.Type. It must be `TypeThin`, `TypeThick`, or empty.

lvmd/device_class_manager.go
```diff
@@ -20,7 +26,16 @@ var qualifiedNameRegexp = regexp.MustCompile("^([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Z
 // This regexp is used to check StripeSize format
 var stripeSizeRegexp = regexp.MustCompile("(?i)^([0-9]*)(k|m|g|t|p|e|b|s)?$")
 
-// DeviceClass maps between device-classes and volume groups.
+// ThinPoolConfig holds the configuration of thin pool in a volume group
+type ThinPoolConfig struct {
+	// Name of thinpool
+	Name string `json:"name"`
+	// OverprovisionRatio signifies the upper bound multiplier for allowing logical volume creation in this pool
+	OverprovisionRatio *float64 `json:"overprovision-ratio"`
```
llamerada-jp, line 34:
```suggestion
	OverprovisionRatio float64 `json:"overprovision-ratio"`
```
This value looks not nullable.

llamerada-jp (changes requested):

leelavg (commented):
addressed comments.

lvmd/command/lvm.go
```diff
@@ -505,6 +589,14 @@ func (l *LogicalVolume) Pool() (*ThinPool, error) {
 	return l.vg.FindPool(*l.pool)
 }
 
+// GetPoolName returns thin pool name that the logical volume is part of or nil
```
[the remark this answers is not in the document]
leelavg: - on another thought, better to return `*string` as we are only checking the presence in call site
- will update this in next review from others

lvmd/lvservice.go
```diff
@@ -41,7 +43,27 @@ func (s *lvService) CreateLV(_ context.Context, req *proto.CreateLVRequest) (*pr
...
+	case TypeThin:
+		pool, err = vg.FindPool(dc.ThinPoolConfig.Name)
+		if err != nil {
+			return nil, err
+		}
+		tpu, err := pool.Free()
+		if err != nil {
+			return nil, err
+		}
+		free = uint64(math.Ceil(dc.ThinPoolConfig.OverprovisionRatio*float64(tpu.SizeBytes))) - tpu.VirtualBytes
```
llamerada-jp, line 60:
```suggestion
		free = uint64(math.Ceil(dc.ThinPoolConfig.OverprovisionRatio*float64(tpu.FreeBytes)))
```
`OverprovisionRatio * SizeBytes - VirtualBytes` is not the virtual free size. `FreeBytes` is near the actual free size, it is suitable to calculate free size.

lvmd/vgservice.go
```diff
@@ -111,14 +155,53 @@ func (s *vgService) send(server proto.VGService_WatchServer) error {
...
+			tpi := &proto.ThinPoolItem{}
+			pool, err := vg.FindPool(dc.ThinPoolConfig.Name)
+			if err != nil {
+				return status.Error(codes.Internal, err.Error())
+			}
+			tpu, err := pool.Free()
+			if err != nil {
+				return status.Error(codes.Internal, err.Error())
+			}
+			opb := uint64(math.Ceil(dc.ThinPoolConfig.OverprovisionRatio*float64(tpu.SizeBytes))) - tpu.VirtualBytes
```
llamerada-jp, line 179:
```suggestion
			opb := uint64(math.Ceil(dc.ThinPoolConfig.OverprovisionRatio*float64(tpu.FreeBytes)))
```

llamerada-jp (commented):
I (re)reviewed `watch` and `metrics`. I'm going to review the tests. Do you have a plan to add e2e tests?

lvmd/command/lvm.go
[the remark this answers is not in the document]
leelavg: - linking to other comments seems to be hard
- so, it's decided to keep the overprovision ratio out of this file and calculations are performed at call site

lvmd/command/lvm.go
```diff
@@ -415,14 +438,88 @@ func (t *ThinPool) ListVolumes() ([]*LogicalVolume, error) {
...
+	}
+	return nil, ErrNotFound
+}
+
 // CreateVolume creates a thin volume from this pool.
-func (t *ThinPool) CreateVolume(name string, size uint64) (*LogicalVolume, error) {
-	if err := CallLVM("lvcreate", "-T", t.fullname, "-n", name, "-V", fmt.Sprintf("%vg", size>>30)); err != nil {
+func (t *ThinPool) CreateVolume(name string, size uint64, tags []string, stripe uint, stripeSize string, lvcreateOptions []string) (*LogicalVolume, error) {
+
+	lvcreateArgs := []string{"-T", t.FullName(), "-n", name, "-V", fmt.Sprintf("%vg", size>>30), "-W", "y", "-y"}
```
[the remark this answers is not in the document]
leelavg: - for others who might be referring this thread, `-Wy` will wipe the signatures from the logical volume if it had any before creating the logical volumes
---
- this was used in thick vols creation as well and I didn't think it should be configurable and so used this option
- could you pls provide an usecase for making it's configurable?
---
@llamerada-jp any past experience in using this option?

lvmd/device_class_manager.go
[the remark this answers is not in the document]
leelavg: - per pool correct and as mentioned earlier, removed this from the file
- at vg level we'll not have this option

lvmd/device_class_manager.go
```diff
@@ -36,6 +51,10 @@ type DeviceClass struct {
 	StripeSize string `json:"stripe-size"`
 	// LVCreateOptions are extra arguments to pass to lvcreate
 	LVCreateOptions []string `json:"lvcreate-options"`
+	// Type is the name of logical volume target, supports 'thick' (default) or 'thin' currently
+	Type LVType `json:"type,omitempty"`
```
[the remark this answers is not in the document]
leelavg: - in previous discussions it was made to respect the ThinPoolConfig if Type is `thick` or else ThinPoolConfig will not be used
- I also feel it's better to take decision based on Type, let me know otherwise on where this needs some more thought

leelavg:
@llamerada-jp
- thanks a lot for your reviews, they are helpful in simplifying the code in some places
> I (re)reviewed `watch` and `metrics`. I'm going to review the tests. Do you have a plan to add e2e tests?
- updated the PR description, they'll be part of another PR
- my goal for this PR is, thin provisioning implementation and changes in vg/lv services for proto changes, unit tests to verify these and no regressions in e2e
- then we'll have a build which can be used in the operator to do admin tasks and so e2e will be raised later however surely they'll be in before next release
---
- you can proceed with reviewing the unit-tests as I don't see your recent comments will have less effect unit-tests, thanks.

lvmd/lvservice.go
leelavg: - I need to disagree with this statement
- SizeBytes is a constant unless thinpool is extended and we should be calculating the overprovisioned size based on actual thinpool size but not on the moving target FreeBytes which depends on data & meta percent used
- our intention is over-provisioned size should reduce when there are multiple thin lvs created in the thinpool
- however, if we use freebytes in the calculation, over-provisioned size will only be reduced based on actual usage of those thin lvs

lvmd/vgservice.go
leelavg: - will be tracked under https://github.com/topolvm/topolvm/pull/448#discussion_r838056977

leelavg:
- don't trigger the CI yet, it seems I missed out a commit while force pushing 😢 
- let me fix that

lvmd/command/lvm.go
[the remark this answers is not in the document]
llamerada-jp: @nbalacha @leelavg 
It is more problematic when the old data is in the newly allocated area. `-W` option cause no problem, IMO.

llamerada-jp:
@leelavg 

> ci was failing during `make check-uncommited`

```
git diff --exit-code --name-only
csi/csi.pb.go
lvmd/proto/lvmd.pb.go
```

Would you try to re-generate some file like below when nect commit?

```
rm csi/csi.pb.go lvmd/proto/lvmd.pb.go
make csi/csi.pb.go lvmd/proto/lvmd.pb.go
```

llamerada-jp:
@leelavg 

> ci was failing during make check-uncommited

I made a mistake. This diff caused by a known bug. I have just been told by a colleague. Could you rebase after merging the PR below?

leelavg:
> Could you rebase after merging the PR below?
- np, I always rebase before each force push.
- last commit to main was 20 days ago, or you want me to rebase on a PR after that specific PR is merged? If yes, pls let me know the PR with fix.

lvmd/lvservice.go
llamerada-jp: @leelavg 
I understood the intent of the formula, but I have a question. Is this formula works normally when snapshots exists?

leelavg: - As a gist, overprovision size is being used for only pod scheduling and better we take snapshot size as well into consideration
- there's a possibility that some snapshots will be activated (~clones) for read/write and they'll behave as a normal thin lv at some point
- having all said, data and metadata percent on thinpool is the key for monitoring and we are exporting those metrics
- so I propose we don't change the current formula and revisit if there need to be some changes after snapshots is merged, sounds good?

llamerada-jp: I think `OverprovisionRatio * SizeBytes - VirtualBytes ` - `data size consumed by the snapshot` is better. About snapshot size, could't you get the value by following option?

```
$ lvs -o help
  Logical Volume Fields
  ---------------------
...
    origin_size            - For snapshots, the size of the origin device of this LV.
...
```

> I want to handle this after looking at snapshot implementation

I agree with you.

leelavg: - yes, that's the idea after snapshot gets it, cause existing one doesn't effect the operations in any way
- will definitely get back after snapshot PR

leelavg (commented):
- uh oh! I responded to the comments yesterday and missed submitting it
- @llamerada-jp

lvmd/lvservice_test.go
```diff
@@ -39,10 +39,43 @@ func TestLVService(t *testing.T) {
 	notifier := func() {
 		count++
 	}
-	lvService := NewLVService(NewDeviceClassManager([]*DeviceClass{{Name: vgName, VolumeGroup: vgName}}), notifier)
```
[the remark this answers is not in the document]
leelavg: @llamerada-jp awaiting your response in this thread to update PR for current set of reviews.

Just clarifing not to think one is waiting for another 😅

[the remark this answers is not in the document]
llamerada-jp: My explanation was insufficient. After creating `LVService` instance, `CreateLV` method called for `test_pool`(thin pool) device class replaced from `test_lvservice`(VG) device class. So the tests `CreateLV` method for VG is looks removed.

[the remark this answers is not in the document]
leelavg: - ntg is removed
- earlier the deviceClass name in `LVService` was the vgName itself but now only the name is changed, everything else stayed same
- tests for thinpool is added at the end, I didn't use a func/loop as at some places sizes if different and overprovisioning should work for thinpool but not on volumegroup
- pls let me know what change you expect in vgservice_test.go and lvservice_test.go

lvmd/vgservice_test.go
```diff
@@ -47,89 +50,170 @@ func (s *mockWatchServer) RecvMsg(m interface{}) error {
...
+	// create thick volume
 	testtag := "testtag"
 	_, err = vg.CreateVolume("test1", 1<<30, []string{testtag}, 0, "", nil)
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	res, err = vgService.GetLVList(context.Background(), &proto.GetLVListRequest{DeviceClass: vg.Name()})
+	// thick lv validation
+	res, err = vgService.GetLVList(context.Background(), &proto.GetLVListRequest{DeviceClass: thickdev})
```
llamerada-jp, line 216: The tests for VG are replaced for thin-pool and tests for VG are looks removed. Would you keep the tests for VG? It is similar to [this comment](https://github.com/topolvm/topolvm/pull/448#discussion_r839364202).

llamerada-jp (commented):
I reviewed `lvm/vg service test.go`. Would you check some comments?

lvmd/lvservice_test.go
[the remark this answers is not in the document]
llamerada-jp: Both of `lvService.CreateLV(... DeviceClass: thickdev, ...)` and `lvService.CreateLV(... DeviceClass: vgName, ...)` are need.  If only `DeviceClass: thickdev`, this sections([1](https://github.com/topolvm/topolvm/blob/976bd0bc6fd3a5bccff4edce63465f886dcd617d/lvmd/lvservice.go#L49-L56), [2](https://github.com/topolvm/topolvm/blob/976bd0bc6fd3a5bccff4edce63465f886dcd617d/lvmd/lvservice.go#L99)) will be not test.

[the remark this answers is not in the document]
llamerada-jp: @leelavg 
My mistake! The test was not removed.

lvmd/vgservice_test.go
llamerada-jp: This comment is my mistake. `thickdev` is not `thindev`.

lvmd/vgservice_test.go
```diff
@@ -47,89 +50,170 @@ func (s *mockWatchServer) RecvMsg(m interface{}) error {
...
+				VolumeGroup: "test_vgservice",
+			}},
+		},
+		{"thinpool", []*DeviceClass{
+			{
+				Name:        "ssd",
+				VolumeGroup: "test_vgservice",
+				Type:        TypeThin,
+				ThinPoolConfig: &ThinPoolConfig{
+					Name:               "test_pool",
```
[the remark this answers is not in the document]
leelavg: 1. `func TestVGService(t *testing.T) ` creates volume group before running `testVGService(t, vg)`
2. Inside `func testVGService(t *testing.T, vg *command.VolumeGroup)` thinpool in the supplied volume group is created and after `testVGService(t, vg)` completion `testWatch(t)` is called
3. Inside `func testWatch(t *testing.T)` volumegroup created in step 1 and thinpool created in step 2 will be used to verify the metrics flow

maybe it'd help to directly see whole files 
https://github.com/leelavg/topolvm/blob/thin-lv-3/lvmd/vgservice_test.go
https://github.com/leelavg/topolvm/blob/thin-lv-3/lvmd/lvservice_test.go

lvmd/vgservice_test.go
leelavg: - yes, as mentioned no existing unit test is removed.

lvmd/vgservice_test.go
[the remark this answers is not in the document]
llamerada-jp: Thank you. I understood.

lvmd/device_class_manager_test.go
```diff
@@ -74,6 +76,61 @@ func TestValidateDeviceClasses(t *testing.T) {
...
+					VolumeGroup: "vg1",
+					Type:        TypeThin,
+					ThinPoolConfig: &ThinPoolConfig{
+						Name:               "pool0",
+						OverprovisionRatio: opRatio,
+					},
+				},
+			},
+			valid: true,
+		},
```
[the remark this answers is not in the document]
leelavg: done.

leelavg:
- rebased, added extra unit test, addressed comments.

lvmd/vgservice_test.go
```diff
@@ -47,89 +50,170 @@ func (s *mockWatchServer) RecvMsg(m interface{}) error {
...
 	if err != nil {
 		t.Fatal(err)
 	}
 	numVols1 := len(res.GetVolumes())
 	if numVols1 != 0 {
 		t.Errorf("numVolumes must be 0: %d", numVols1)
 	}
+
+	// thin lvs
+	res, err = vgService.GetLVList(context.Background(), &proto.GetLVListRequest{DeviceClass: thickdev})
```
cupnes, line 198:
```suggestion
	res, err = vgService.GetLVList(context.Background(), &proto.GetLVListRequest{DeviceClass: thindev})
```

lvmd/vgservice_test.go
```diff
@@ -183,6 +324,23 @@ func testVGService(t *testing.T, vg *command.VolumeGroup) {
...
+	// thin lv size validations
+	res2, err = vgService.GetFreeBytes(context.Background(), &proto.GetFreeBytesRequest{DeviceClass: thindev})
+	if err != nil {
+		t.Fatal(err)
+	}
+	tpu, err := pool.Free()
+	if err != nil {
+		t.Fatal(err)
+	}
+	opb := uint64(math.Ceil(overprovisionRatio*float64(tpu.SizeBytes))) - tpu.VirtualBytes
```
cupnes, line 336:
```suggestion
	opb := uint64(math.Floor(overprovisionRatio*float64(tpu.SizeBytes))) - tpu.VirtualBytes
```

cupnes (changes requested):

docs/lvmd-protocol.md
```diff
@@ -193,17 +194,39 @@ The volume size will be set to exactly &#34;size_gb&#34;.
...
+
+### ThinPoolItem
+Represents the details of thinpool.
+
+
+| Field | Type | Label | Description |
+| ----- | ---- | ----- | ----------- |
+| data_percent | [double](#double) |  | Data and Metadata fields are used for monitoring thinpool
+
+Data percent occupied on the thinpool |
```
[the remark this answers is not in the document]
leelavg: good catch, thanks, updated.

docs/lvmd-protocol.md
```diff
@@ -193,17 +194,39 @@ The volume size will be set to exactly &#34;size_gb&#34;.
...
+| metadata_percent | [double](#double) |  | Metadata percent occupied on the thinpool |
+| overprovision_bytes | [uint64](#uint64) |  | Used for scheduling decisions
+
+Free space on the thinpool with overprovision. |
```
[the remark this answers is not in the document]
leelavg: updated.

leelavg (commented):
addressed comments.

cupnes (dismissed):
Thank you for your updates. LGTM

lvmd/vgservice.go
```diff
@@ -111,14 +163,53 @@ func (s *vgService) send(server proto.VGService_WatchServer) error {
...
+			}
+
+			// used for updating prometheus metrics
+			tpi.DataPercent = tpu.DataPercent
+			tpi.MetadataPercent = tpu.MetadataPercent
+
+			// used for annotating the node for capacity aware scheduling
+			tpi.OverprovisionBytes = opb
+
+			// TODO (leelavg): one can calculate freebytes by using data, metdata and size values,
```
llamerada-jp, line 200: @leelavg 
There are some `TODO` in your code. Would you please check those comments and remove them if aren't needed.

lvmd/vgservice.go
```diff
@@ -111,14 +163,53 @@ func (s *vgService) send(server proto.VGService_WatchServer) error {
...
+			if err != nil {
+				return status.Error(codes.Internal, err.Error())
+			}
+			tpu, err := pool.Free()
+			if err != nil {
+				return status.Error(codes.Internal, err.Error())
+			}
+			opb := uint64(math.Floor(dc.ThinPoolConfig.OverprovisionRatio*float64(tpu.SizeBytes))) - tpu.VirtualBytes
+			if dc.Default {
+				// TODO (leeavg): remove this after removal of support for inline ephemeral volumes
```
llamerada-jp, line 189: I would like you to leave this comment.

llamerada-jp (changes requested):

lvmd/vgservice.go
leelavg: done.

lvmd/vgservice.go
leelavg: done, left out 2 TODOs (one here and another in test) and removed the rest.

leelavg:
@llamerada-jp removed obsolete TODO's

lvmd/command/lvm.go
```diff
@@ -415,14 +425,88 @@ func (t *ThinPool) ListVolumes() ([]*LogicalVolume, error) {
...
+			// thin lv
+			lvSize, err := strconv.ParseUint(info["lv_size"], 10, 64)
+			if err != nil {
+				return nil, err
+			}
+			virtualSize += lvSize
+		}
+	}
+
+	free := 1.0 - (tpu.DataPercent+tpu.MetadataPercent)/100
```
nbalacha, line 504: Why is the metadata percent part of the calculation? It should only use the datapercent.  
Consider if a metadata pool is 15% full but the data is only 1% full.
Here the calculated free values are incorrect.

leelavg: - it depends on the context where this `free` is used in the rest of the code
1. this isn't used in scheduling decisions
2. used in monitoring however the user for thinpool should rely on data and metadata percent that are exposed when the metrics is for thinpool subsystem
3. in general topolvm freebytes on a volume group is also part of the metrics using the same notion, freebytes on thinpool is calculated based on data and metadata percent occupied in the thinpool
4. why will the free values here be incorrect as far as the calculation is concerned, as for the context user isn't going to create alerts based on this field rather the user will use data and metadata percentages

llamerada-jp: I misunderstand too. `DataPercent` and `MetadataPercent`  may describe occupancy in different areas in an LV. In the following situation, `free` is incorrect. If you write data to the thin volume fully, you will get the following outputs.

```
$ sudo lvs
  LV        VG        Attr       LSize   Pool   Origin Data%  Meta%  Move Log Cpy%Sync Convert
  mylv      loop      Vwi-aotz--   1.00t mypool        0.01
  mypool    loop      twi-aotzD- 100.00m               100.00 12.11
  ubuntu-lv ubuntu-vg -wi-ao---- 110.00g
```

nbalacha: > * it depends on the context where this `free` is used in the rest of the code
So where exactly is it used?
> 
> 1. this isn't used in scheduling decisions
> 2. used in monitoring however the user for thinpool should rely on data and metadata percent that are exposed when the metrics is for thinpool subsystem
> 3. in general topolvm freebytes on a volume group is also part of the metrics using the same notion, freebytes on thinpool is calculated based on data and metadata percent occupied in the thinpool
> 4. why will the free values here be incorrect as far as the calculation is concerned, as for the context user isn't going to create alerts based on this field rather the user will use data and metadata percentages

nbalacha: The tpu.FreeBytes is using this value for its calculation. When is this used?

leelavg: - understood! thanks for pointing this out
- I'll remove this field as there was a TODO on this field which was asking whether this can be dropped altogether

leelavg: @nbalacha @llamerada-jp @cupnes since rest of the PR is reviewed multiple times, providing below diff for the changes done for the force push after removing tpu.FreeBytes, hope it'll be helpful

``` diff
diff --git a/lvmd/command/lvm.go b/lvmd/command/lvm.go
index 85a4d5ca..0d030191 100644
--- a/lvmd/command/lvm.go
+++ b/lvmd/command/lvm.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"math"
 	"os"
 	"os/exec"
 	"path"
@@ -360,7 +359,6 @@ type ThinPoolUsage struct {
 	DataPercent     float64
 	MetadataPercent float64
 	VirtualBytes    uint64
-	FreeBytes       uint64
 	SizeBytes       uint64
 }
 
@@ -462,8 +460,8 @@ func (t *ThinPool) CreateVolume(name string, size uint64, tags []string, stripe
 	return t.vg.FindVolume(name)
 }
 
-// Free on a thinpool returns used data, metadata percentages, calculated freebytes
-// using percentages, sum of virtualsizes of all thinlvs and size of thinpool
+// Free on a thinpool returns used data, metadata percentages,
+// sum of virtualsizes of all thinlvs and size of thinpool
 func (t *ThinPool) Free() (*ThinPoolUsage, error) {
 	tpu := &ThinPoolUsage{}
 	infoList, err := parseOutput("lvs", "lv_size,data_percent,metadata_percent,pool_lv", "-S", fmt.Sprintf("lv_name=%s||pool_lv=%s", t.Name(), t.Name()))
@@ -501,8 +499,6 @@ func (t *ThinPool) Free() (*ThinPoolUsage, error) {
 		}
 	}
 
-	free := 1.0 - (tpu.DataPercent+tpu.MetadataPercent)/100
-	tpu.FreeBytes = uint64(math.Floor(float64(free) * float64(tpu.SizeBytes)))
 	tpu.VirtualBytes = virtualSize
 	return tpu, nil
 }
diff --git a/lvmd/vgservice.go b/lvmd/vgservice.go
index 457644e9..da2ca18c 100644
--- a/lvmd/vgservice.go
+++ b/lvmd/vgservice.go
@@ -200,7 +200,6 @@ func (s *vgService) send(server proto.VGService_WatchServer) error {
 			// include thinpoolitem in the response
 			res.Items = append(res.Items, &proto.WatchItem{
 				DeviceClass: dc.Name,
-				FreeBytes:   tpu.FreeBytes,
 				SizeBytes:   tpu.SizeBytes,
 				ThinPool:    tpi,
 			})
diff --git a/runners/metrics_exporter.go b/runners/metrics_exporter.go
index 8888bb4e..9d6e9b97 100644
--- a/runners/metrics_exporter.go
+++ b/runners/metrics_exporter.go
@@ -41,10 +41,9 @@ type NodeMetrics struct {
 
 // thinPoolMetricsExporter is the subset of metricsExporter corresponding to the deviceclass target
 type thinPoolMetricsExporter struct {
-	tpAvailableBytes *prometheus.GaugeVec
-	tpSizeBytes      *prometheus.GaugeVec
-	dataPercent      *prometheus.GaugeVec
-	metadataPercent  *prometheus.GaugeVec
+	tpSizeBytes     *prometheus.GaugeVec
+	dataPercent     *prometheus.GaugeVec
+	metadataPercent *prometheus.GaugeVec
 }
 
 type metricsExporter struct {
@@ -82,15 +81,6 @@ func NewMetricsExporter(conn *grpc.ClientConn, mgr manager.Manager, nodeName str
 	metrics.Registry.MustRegister(sizeBytes)
 
 	// metrics available under thinpool subsystem
-	tpAvailableBytes := prometheus.NewGaugeVec(prometheus.GaugeOpts{
-		Namespace:   metricsNamespace,
-		Subsystem:   "thinpool",
-		Name:        "available_bytes",
-		Help:        "LVM VG Thin Pool raw available bytes",
-		ConstLabels: prometheus.Labels{"node": nodeName},
-	}, []string{"device_class"})
-	metrics.Registry.MustRegister(tpAvailableBytes)
-
 	tpSizeBytes := prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace:   metricsNamespace,
 		Subsystem:   "thinpool",
@@ -125,10 +115,9 @@ func NewMetricsExporter(conn *grpc.ClientConn, mgr manager.Manager, nodeName str
 		availableBytes: availableBytes,
 		sizeBytes:      sizeBytes,
 		thinPool: &thinPoolMetricsExporter{
-			tpAvailableBytes: tpAvailableBytes,
-			tpSizeBytes:      tpSizeBytes,
-			dataPercent:      dataPercent,
-			metadataPercent:  metadataPercent,
+			tpSizeBytes:     tpSizeBytes,
+			dataPercent:     dataPercent,
+			metadataPercent: metadataPercent,
 		},
 	}
 }
@@ -148,7 +137,6 @@ func (m *metricsExporter) Start(ctx context.Context) error {
 					m.sizeBytes.WithLabelValues(met.DeviceClass).Set(float64(met.SizeBytes))
 				} else if met.DeviceClassType == TypeThin {
 					// metrics for thinpool subsystem
-					m.thinPool.tpAvailableBytes.WithLabelValues(met.DeviceClass).Set(float64(met.FreeBytes))
 					m.thinPool.tpSizeBytes.WithLabelValues(met.DeviceClass).Set(float64(met.SizeBytes))
 					m.thinPool.dataPercent.WithLabelValues(met.DeviceClass).Set(float64(met.DataPercent))
 					m.thinPool.metadataPercent.WithLabelValues(met.DeviceClass).Set(float64(met.MetadataPercent))
@@ -186,7 +174,6 @@ func (m *metricsExporter) updateNode(ctx context.Context, wc proto.VGService_Wat
 			if item.ThinPool != nil {
 				ch <- NodeMetrics{
 					DeviceClass:     item.DeviceClass,
-					FreeBytes:       item.FreeBytes,
 					SizeBytes:       item.SizeBytes,
 					DataPercent:     item.ThinPool.DataPercent,
 					MetadataPercent: item.ThinPool.MetadataPercent,
@@ -227,9 +214,11 @@ func (m *metricsExporter) updateNode(ctx context.Context, wc proto.VGService_Wat
 
 		node2.Annotations[topolvm.CapacityKeyPrefix+topolvm.DefaultDeviceClassAnnotationName] = strconv.FormatUint(res.FreeBytes, 10)
 		for _, item := range res.Items {
-			freeSize := item.FreeBytes
+			var freeSize uint64
 			if item.ThinPool != nil {
 				freeSize = item.ThinPool.OverprovisionBytes
+			} else {
+				freeSize = item.FreeBytes
 			}
 			node2.Annotations[topolvm.CapacityKeyPrefix+item.DeviceClass] = strconv.FormatUint(freeSize, 10)
 		}


```

lvmd/command/lvm.go
```diff
@@ -415,14 +423,86 @@ func (t *ThinPool) ListVolumes() ([]*LogicalVolume, error) {
 	return ret, nil
 }
 
+// FindVolume finds a named logical volume in this thin pool
+func (t *ThinPool) FindVolume(name string) (*LogicalVolume, error) {
```
nbalacha, line 427: This function calls ListVolumes which loops over all the volumes in the vg and returns those in the thinpool. This function then loops over those volumes to find the matching name.
It would be more efficient to write a function which will list the volumes in the vg and check pool pool and name in the same loop.

```
func (t *ThinPool) FindVolume(name string) (*LogicalVolume, error) { {
	volumes, err := t.vg.ListVolumes()
	if err != nil {
		return nil, err
	}

	for _, volume := range volumes {
		if volume.name == name && volume.pool != nil && *volume.pool == t.name {
			return volume, nil
		}
	}
	return nil, ErrNotFound
}
```

nbalacha (commented):
Just a suggestion.

lvmd/command/lvm.go
leelavg: - ack, makes sense, since the lv names are unique the condition will be evaluated only once and reduces looping as well compared to existing implementation
- again, pasting the diff as not all code needs to be re-reviwed
``` diff
# gd thin-lv-3..temp
diff --git a/lvmd/command/lvm.go b/lvmd/command/lvm.go
index 0d030191..24aff2b9 100644
--- a/lvmd/command/lvm.go
+++ b/lvmd/command/lvm.go
@@ -425,12 +425,13 @@ func (t *ThinPool) ListVolumes() ([]*LogicalVolume, error) {

 // FindVolume finds a named logical volume in this thin pool
 func (t *ThinPool) FindVolume(name string) (*LogicalVolume, error) {
-       volumes, err := t.ListVolumes()
+       volumes, err := t.vg.ListVolumes()
        if err != nil {
                return nil, err
        }
+
        for _, volume := range volumes {
-               if volume.Name() == name {
+               if volume.name == name && volume.pool != nil && *volume.pool == t.name {
                        return volume, nil
                }
        }
@@ -585,11 +586,6 @@ func (l *LogicalVolume) Pool() (*ThinPool, error) {
        return l.vg.FindPool(*l.pool)
 }

-// GetPoolName returns a pointer to name of the logical volumes
-func (l *LogicalVolume) GetPoolName() *string {
-       return l.pool
-}
-
 // MajorNumber returns the device major number.
 func (l *LogicalVolume) MajorNumber() uint32 {
        return l.devMajor
diff --git a/lvmd/vgservice.go b/lvmd/vgservice.go
index da2ca18c..ded3aa49 100644
--- a/lvmd/vgservice.go
+++ b/lvmd/vgservice.go
@@ -71,7 +71,7 @@ func (s *vgService) GetLVList(_ context.Context, req *proto.GetLVListRequest) (*

        vols := make([]*proto.LogicalVolume, 0, len(lvs))
        for _, lv := range lvs {
-               if dc.Type == TypeThick && lv.GetPoolName() != nil {
+               if dc.Type == TypeThick && lv.IsThin() {
                        // do not send thin lvs if request is on TypeThick
                        continue
                }
```
- thanks @Yuggupta27 for pointing out existing `IsThin()` method.

llamerada-jp (approved):
LGTM
@nbalacha 
I'll wait for the merge if necessary to review. Is it OK?

nbalacha (commented):
Looks good to me.

runners/metrics_exporter.go
```diff
@@ -79,8 +131,16 @@ func (m *metricsExporter) Start(ctx context.Context) error {
...
 			case met := <-metricsCh:
-				m.availableBytes.WithLabelValues(met.DeviceClass).Set(float64(met.FreeBytes))
-				m.sizeBytes.WithLabelValues(met.DeviceClass).Set(float64(met.SizeBytes))
+				if met.DeviceClassType == TypeThick {
+					// metrics for volumegroup subsystem
+					m.availableBytes.WithLabelValues(met.DeviceClass).Set(float64(met.FreeBytes))
+					m.sizeBytes.WithLabelValues(met.DeviceClass).Set(float64(met.SizeBytes))
+				} else if met.DeviceClassType == TypeThin {
+					// metrics for thinpool subsystem
+					m.thinPool.tpSizeBytes.WithLabelValues(met.DeviceClass).Set(float64(met.SizeBytes))
```
sp98, line 140: Can you also include the thin pool name in the labels?

llamerada-jp: It looks better to use other metrics like `kube_persistentvolume_info` to describe VG/pool name and another property. If this feature is not an urgent requirement, Would you discuss it later?

sp98: Not an urgent requirement.  We can discuss this later.

cupnes (approved):
I reviewed again. LGTM

PR #453, opened 2022-03-25
make CI faster

toshipp:
- cache go dependencies
- cache e2e sidecar binaries
- do not build container images fully. they are just variants, so one can use another as a base.
- skip tests if md only change.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,11 @@
-build
+.git
 bin
-include
-example
+build
+charts
+config
+deploy
+docs
 e2e
+example
+include
+testbin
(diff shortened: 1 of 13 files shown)


leelavg:
- hi, just a suggestion, it's better to add paths-ignore for docs as well https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths?
- As there were instances only docs addition/update triggered whole CI suite iirc

toshipp:
neat. I'll improve it to skip CI if only docs change.

leelavg (approved):

peng225 (approved):

cupnes (approved):

PR #462, opened 2022-04-07
revise CODEOWNERS

toshipp:


cupnes (approved):

peng225 (dismissed):

.github/workflows/main.yaml
```diff
@@ -3,9 +3,11 @@ on:
   pull_request:
     paths-ignore:
       - "**/*.md"
+      - "CODEOWNERS"
```
daichimukai, line 6: If I understand correctly, we'll get stuck when we change CODEOWNERS file next time since main workflow must pass to be merged by branch protection rule. This problem also exists for `**/*.md` rule.

toshipp: I think you are right. I'll remove paths-ignore from the workflow. This runs just two jobs, almost fine.

cupnes (approved):

daichimukai (approved):

PR #465, opened 2022-04-13
Modified to use k8s.io/utils/io.ConsistentRead

### LINKED ISSUE #371: Avoid getting incorrect information from `/proc/mounts`.
## What

We rarely get wrong information from `/proc/mounts` if read it simply according to the following post.
ref. https://github.com/kubernetes/kubernetes/issues/31272#issuecomment-300209720

## How

Read and parse `/proc/mounts` repeatedly until we get 2 in a row that match.

## Checklist

- [ ] Test all functions

bells17:
fix #371

### DIFF
--- a/filesystem/util.go
+++ b/filesystem/util.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"golang.org/x/sys/unix"
+	"k8s.io/utils/io"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


toshipp (approved):
LGTM

llamerada-jp (approved):
LGTM

PR #467, opened 2022-04-15
fix: send correct proto.WatchResponse for thin and thick lvs

leelavg:
- a vg can host both thin and thick vols at same time
- this pr makes that distinction clear and send the WatchResponse correctly
- existing code will either send WatchResponse for thick or thin vols but not for both
- add thin vol steps in example build

Signed-off-by: Leela Venkaiah G <lgangava@redhat.com>

### DIFF
--- a/example/Makefile
+++ b/example/Makefile
@@ -44,10 +44,11 @@ run:
 	$(HELM) install --namespace=topolvm-system topolvm ../charts/topolvm/ -f ./values.yaml
 	$(KUBECTL) wait --for=condition=available --timeout=120s -n topolvm-system deployments/topolvm-controller
 	$(KUBECTL) apply -f podpvc.yaml
+	$(KUBECTL) wait --for=condition=ready --timeout=60s -n default pod -l app=example
 
 setup: $(KUBECTL)
 	$(SUDO) apt-get update
-	$(SUDO) apt-get install -y lvm2 xfsprogs
+	$(SUDO) apt-get install -y lvm2 xfsprogs thin-provisioning-tools
 	cd ..; $(MAKE) install-kind
 	mkdir -p $(BINDIR)
 	mkdir -p build
@@ -87,6 +88,7 @@ start-lvmd: $(TMPDIR)/lvmd/lvmd.yaml
 	truncate --size=20G $(BACKING_STORE)/backing_store; \
 	$(SUDO) losetup -f $(BACKING_STORE)/backing_store; \
 	$(SUDO) vgcreate -f -y myvg1 $$($(SUDO) losetup -j $(BACKING_STORE)/backing_store | cut -d: -f1); \
+	$(SUDO) lvcreate -T myvg1/thinpool -L 2G; \
 	$(SUDO) systemd-run --unit=lvmd.service $(shell pwd)/build/lvmd --config=$(TMPDIR)/lvmd/lvmd.yaml; \
 
 stop-lvmd:
--- a/lvmd/device_class_manager.go
+++ b/lvmd/device_class_manager.go
@@ -138,9 +138,10 @@ func ValidateDeviceClasses(deviceClasses []*DeviceClass) error {
 
 // DeviceClassManager maps between device-classes and volume groups.
 type DeviceClassManager struct {
-	defaultDeviceClass  *DeviceClass
-	deviceClassByName   map[string]*DeviceClass
-	deviceClassByVGName map[string]*DeviceClass
+	defaultDeviceClass        *DeviceClass
+	deviceClassByName         map[string]*DeviceClass
+	deviceClassByVGName       map[string]*DeviceClass
+	deviceClassByThinPoolName map[string]*DeviceClass
 }
 
 // NewDeviceClassManager creates a new DeviceClassManager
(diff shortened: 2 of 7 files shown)


leelavg:
- The timing of this PR is so bad due to below changes I had to spend considerable amount of time to figure out the issue
a. thinp not being part of 0.10.6 (chart appVersion)
b. change of image repo from quay to ghcr
c. bump version of only app but not chart with example always pointing to appVersion in chart 
d. lvmd as systemd unit in example with different version of other components in kind
- If it had been a week before or after, above issues wouldn't be hit

----

- All in all, this PR proves the point that thinp is working fine alongside thick vols
- Preparation of doc is still in progress for thinp e2e and pls expect a delay on that

```
# at the end of `make -C example run BUILD_IMAGE="true"`
# k get no -ojson|jq -r '.items[].metadata.annotations | with_entries(if (.key|test("topolvm")) then ({key: .key, value: .value }) else empty end)'
{}
{
  "capacity.topolvm.cybozu.com/00default": "18241028096",
  "capacity.topolvm.cybozu.com/ssd": "18241028096",
  "capacity.topolvm.cybozu.com/ssd-thin": "10737418240"
}


# curl -s localhost:8080/metrics | grep -Ei 'topolvm_' | grep -Ev 'HELP|TYPE'
topolvm_thinpool_data_percent{device_class="ssd-thin",node="topolvm-example-worker"} 0.54
topolvm_thinpool_metadata_percent{device_class="ssd-thin",node="topolvm-example-worker"} 11.52
topolvm_thinpool_size_bytes{device_class="ssd-thin",node="topolvm-example-worker"} 2.147483648e+09
topolvm_volumegroup_available_bytes{device_class="ssd",node="topolvm-example-worker"} 1.8241028096e+10
topolvm_volumegroup_size_bytes{device_class="ssd",node="topolvm-example-worker"} 2.1470642176e+10
```

example/Makefile
```diff
@@ -41,13 +41,17 @@ run:
 	$(HELM) repo add jetstack https://charts.jetstack.io
 	$(HELM) repo update
 	$(HELM) dependency build ../charts/topolvm/
-	$(HELM) install --namespace=topolvm-system topolvm ../charts/topolvm/ -f ./values.yaml
+	$(HELM) install --namespace=topolvm-system topolvm ../charts/topolvm/ -f ./values.yaml --set image.tag=$(TOPOLVM_VERSION)
```
toshipp, line 44: We test the chart with the default image version to ensure it works as-is. We will release a new chart soon. Please wait a moment.

example/Makefile
```diff
@@ -41,13 +41,17 @@ run:
...
+	$(KUBECTL) apply -f ./thin-sc.yaml
```
toshipp, line 45: We define SCs by values.yaml. Could you update `example/values.yaml`?

example/Makefile
```diff
@@ -41,13 +41,17 @@ run:
...
 	$(KUBECTL) wait --for=condition=available --timeout=120s -n topolvm-system deployments/topolvm-controller
 	$(KUBECTL) apply -f podpvc.yaml
+	$(KUBECTL) wait --for=condition=ready --timeout=60s -n default pod -l app=example
 
 setup: $(KUBECTL)
-	$(SUDO) apt-get update
-	$(SUDO) apt-get install -y lvm2 xfsprogs
+	if ! grep fedora /etc/os-release > /dev/null; then \
```
toshipp, line 51: It is more reliable to check command directly (e.g. `which apt-get`), IMO.
BTW, we use ubuntu for development, it is a bit noisy.
In top-level Makefile, we separate tool and command setup, would you do it in the same way if you need?

example/Makefile
```diff
@@ -87,7 +91,14 @@ start-lvmd: $(TMPDIR)/lvmd/lvmd.yaml
 	truncate --size=20G $(BACKING_STORE)/backing_store; \
 	$(SUDO) losetup -f $(BACKING_STORE)/backing_store; \
 	$(SUDO) vgcreate -f -y myvg1 $$($(SUDO) losetup -j $(BACKING_STORE)/backing_store | cut -d: -f1); \
-	$(SUDO) systemd-run --unit=lvmd.service $(shell pwd)/build/lvmd --config=$(TMPDIR)/lvmd/lvmd.yaml; \
+	$(SUDO) lvcreate -T myvg1/thinpool -L 2G; \
+	if grep fedora /etc/os-release > /dev/null; then \
+		mv -f $(shell pwd)/build/lvmd /usr/local/bin/lvmd-$(TOPOLVM_VERSION); \
+		$(SUDO) restorecon -rv /usr/local/bin/lvmd-$(TOPOLVM_VERSION); \
+		$(SUDO) systemd-run --unit=lvmd.service /usr/local/bin/lvmd-$(TOPOLVM_VERSION) --config=$(TMPDIR)/lvmd/lvmd.yaml; \
```
toshipp, line 98: Sorry, this is not acceptable.

lvmd/vgservice.go
```diff
@@ -148,33 +148,19 @@ func (s *vgService) send(server proto.VGService_WatchServer) error {
...
-		if err != nil {
-			return err
-		}
 
-		switch dc.Type {
-		case TypeThick:
-			if dc.Default {
-				res.FreeBytes = vgFree
+		for _, pool := range pools {
+			dc, err := s.dcManager.FindDeviceClassByPoolName(vg.Name() + "/" + pool.Name())
```
toshipp, line 157: If arguments are required to be concatenated, it is better to express the restriction as a function parameter.
So, why not define the function as  `FindDeviceClassByPoolName(vg string, pool string)` ?

toshipp (changes requested):

example/Makefile
leelavg: This is a new storage class, let me see if charts can create multiple storage classes via values.yaml

example/Makefile
leelavg: Ack, this tripped me off so badly. Will wait then.

example/Makefile
leelavg: Ack, I'll probably not change anything then.

example/Makefile
leelavg: Ack 😅

example/Makefile
toshipp: Please refer this file.
https://github.com/cybozu-go/neco-apps/blob/7a43554da869c2aa06c562541402fb16a360aa33/topolvm/base/values.yaml#L80-L134

lvmd/vgservice.go
leelavg: Ah! Makes sense.

example/Makefile
leelavg: - if I'm updating the new storageclass in `example/values.yaml` this overwrites existing `topolvm-provisioner` sc
- is it fine if I add `topolvm-provisioner` to `example/values.yaml` as well?

toshipp: Yes. Defining `topolvm-provisioner` is needed. This is a bit ugly, though.

leelavg:
- addressed comments

toshipp (approved):
LGTM

lvmd/device_class_manager.go
```diff
@@ -138,26 +138,33 @@ func ValidateDeviceClasses(deviceClasses []*DeviceClass) error {
 
 // DeviceClassManager maps between device-classes and volume groups.
 type DeviceClassManager struct {
-	defaultDeviceClass  *DeviceClass
-	deviceClassByName   map[string]*DeviceClass
-	deviceClassByVGName map[string]*DeviceClass
+	defaultDeviceClass    *DeviceClass
+	deviceClassByName     map[string]*DeviceClass
+	deviceClassByVGName   map[string]*DeviceClass
+	deviceClassByPoolName map[string]*DeviceClass
```
satoru-takeuchi, line 144: In this case, pool name always means thin pool name. So, how about rename this field to `deviceClassByThinPoolName`?  In addition, it's nice to add a comment to show a device class is in either deviceClassByVGName or deviceClassByPoolName. It would make the code more descriptive.

satoru-takeuchi (changes requested):

lvmd/device_class_manager.go
leelavg: makes sense, will change the var name and add comments.

satoru-takeuchi (approved):

toshipp (approved):

PR #472, opened 2022-04-16
Set fail-fast option to false

bells17:
It is now possible to re-run only failed jobs by themselves with the update of Github Actions. Therefore, even if some jobs fail, I want to change them so that other jobs will continue.
This can reduce the cost of re-run jobs.

- https://github.blog/changelog/2022-03-17-github-actions-re-run-only-failed-or-individual-jobs/
- https://docs.github.com/ja/actions/using-workflows/workflow-syntax-for-github-actions
- https://github.community/t/error-the-operation-was-canceled-in-ci/166506/4
- https://stackoverflow.com/questions/61070925/github-actions-disable-auto-cancel-when-job-fails


toshipp (approved):
LGTM

cupnes (approved):
LGTM

PR #498, opened 2022-06-20
Add support to create PVC-PVC Clones for thin volumes

Yuggupta27:
For PVCs backed by thin-volumes, users will now be able to create
PVC-PVC clones.
Also, added deployment of PVC-Clone to the example setup.

### DIFF
--- a/driver/controller.go
+++ b/driver/controller.go
@@ -460,6 +460,7 @@ func (s controllerService) GetCapacity(ctx context.Context, req *csi.GetCapacity
 func (s controllerService) ControllerGetCapabilities(context.Context, *csi.ControllerGetCapabilitiesRequest) (*csi.ControllerGetCapabilitiesResponse, error) {
 	capabilities := []csi.ControllerServiceCapability_RPC_Type{
 		csi.ControllerServiceCapability_RPC_CREATE_DELETE_VOLUME,
+		csi.ControllerServiceCapability_RPC_CLONE_VOLUME,
 		csi.ControllerServiceCapability_RPC_GET_CAPACITY,
 		csi.ControllerServiceCapability_RPC_EXPAND_VOLUME,
 		csi.ControllerServiceCapability_RPC_CREATE_DELETE_SNAPSHOT,
--- a/e2e/Makefile
+++ b/e2e/Makefile
@@ -173,7 +173,7 @@ create-cluster: topolvm.img
 .PHONY: test
 test: create-cluster
 	$(KUBECTL) apply -f manifests/common/
-	$(SUDO) -E env PATH=${PATH} E2ETEST=1 BINDIR=$(BINDIR) STORAGE_CAPACITY=$(STORAGE_CAPACITY) SKIP_NODE_FINALIZE=$(SKIP_NODE_FINALIZE) READ_WRITE_ONCE_POD=$(READ_WRITE_ONCE_POD) GOLANG_PROTOBUF_REGISTRATION_CONFLICT=warn $(GINKGO) -skip='snapshot' -skip='CreateSnapshot' -skip='DeleteSnapshot' --failFast -v .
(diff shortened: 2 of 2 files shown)


Yuggupta27:
> We want to keep the example as simple as possible, so would you please remove the example of clone feature? Instead of that, would you add an example of datasource in `user-manual.md`?

@llamerada-jp Sure, I agree. Instead of adding it to the example deployment, we can make it a part of the already existing [design doc](https://github.com/topolvm/topolvm/pull/447) where we can explain its usage.

> Clone feature can be used with only a thin pool, aren't you? example environment is using a thick pool, so the example of this  PR isn't runnable. If you can run it, it may make a thin snapshot in a thick pool. Would you check it because that is unexpected behavior?

In a previous PR, we added support to [create a thin pool](https://github.com/topolvm/topolvm/pull/467/files#diff-a7f0749bdafe7d4247c8435bbb139a3ca42c20bf58a96868ac286419afb764d1R91) in the example cluster, and also updated the `podpvc.yaml` to create a `thin` PVC along with a thick one([See](https://github.com/topolvm/topolvm/pull/467/files#diff-8b57ea8aa84aceada091ad93b91c3acc786aeb3d5eea6f7013d05325b6880cfdR33-R66)). With the added example deployment for [PVC-Clone](https://github.com/topolvm/topolvm/pull/498/commits/3352ff782b331582e5a218cdd2dcc3db3162426e), we are creating a PVC-PVC Clone, from already existing `thin` PVC.
Thus, as a result, we are creating a thin-snapshot (internally for clone) out of a thin parent volume.

> About the e2e test, are you planning to update it in another PR the same as a snapshot?

Since e2e for snapshot is also pending, keeping in mind the strict timelines, we have planned to add all the e2e in a separate PR (As majority of helper functions will be same for restore as for clone; so easier to add everything at once,)
In meantime, just added the deployment of PVC-Clone from a thin-parent-PVC in the example cluster, to demonstrate the successful working of the feature(Which we can get rid of if you suggest).

llamerada-jp:
> we can make it a part of the already existing https://github.com/topolvm/topolvm/pull/447 where we can explain its usage.

Sure. Let's go with that idea.

> In a previous PR, we added support to [create a thin pool](https://github.com/topolvm/topolvm/pull/467/files#diff-a7f0749bdafe7d4247c8435bbb139a3ca42c20bf58a96868ac286419afb764d1R91) in the example cluster

I missed it. The behavior of example in this PR is normal. Thank you.

> Since e2e for snapshot is also pending, keeping in mind the strict timelines, we have planned to add all the e2e in a separate PR (As majority of helper functions will be same for restore as for clone; so easier to add everything at once,)

I understand. Let's go ahead with that plan.

> In meantime, just added the deployment of PVC-Clone from a thin-parent-PVC in the example cluster, to demonstrate the successful working of the feature(Which we can get rid of if you suggest).

We would like to ask you to delete the changes to the example.

And I'm going to check several behaviors today.

Yuggupta27:
> We would like to ask you to delete the changes to the example.

Thanks for the review! Updated :+1:

llamerada-jp (approved):
I checked serval behavior. The behavior is in accordance with the specification, but the following patterns need attention.
- Make origin PVC and Pod.
- Make clone PVC and pod with podAffinity.
- Delete origin PVC and Pod.
- After deleting the origin Pod, the cloned pod can not restart because the cloned pod can not satisfy podAffinity.

To avoid this issue, we shouldn't delete origin pod or should use nodeAffinity instead of podAffinity. There is no problem with this PR, but it's better to add a note to the documentation.

Yuggupta27:
> I checked serval behavior. The behavior is in accordance with the specification, but the following patterns need attention.
> 
> * Make origin PVC and Pod.
> * Make clone PVC and pod with podAffinity.
> * Delete origin PVC and Pod.
> * After deleting the origin Pod, the cloned pod can not restart because the cloned pod can not satisfy podAffinity.
> 
> To avoid this issue, we shouldn't delete origin pod or should use nodeAffinity instead of podAffinity. There is no problem with this PR, but it's better to add a note to the documentation.

Thanks @llamerada-jp for verifying! Will add the notes in the document PR https://github.com/topolvm/topolvm/pull/447 once this PR is merged.

cupnes (approved):
LGTM

PR #500, opened 2022-06-21
example: wait for topolvm controller mutating webhook to become ready

daichimukai:
The run target of example Makefile setup topolvm and then deploy sample pods and pvcs. Deploying pods and pvcs requires that mutating webhook of topolvm controller is ready to process requests. Currently we are waiting for the controller to be available condition, but the certificate is also required for  webhook to work. This patch reduces flakiness by waiting for the certificate to become ready.


Yuggupta27 (approved):
LGTM

toshipp (approved):

satoru-takeuchi (approved):

PR #522, opened 2022-07-05
mount: remove new UUID generation

Yuggupta27:
On generating a new UUID(using xfs_admin here), the mount_linux
re-formats the file system; due to which all
the data in the PVC-Clone and PVC-Restore
volume is removed.

To address this issue, we use the `nouuid` mountoption.
Adding 'nouuid' mount option to all XFS mount so that
we will be able to mount a volume and its restored snapshot on
the same node. Without the option, such a mount fails, because 
XFS detects that two different volumes with the same
filesystem UUID are being mounted

On the other hand, ext4, by nature, has the `nouuid` enabled due to which
we don't need to generate a new UUID there too.

Referenced Issue: https://github.com/container-storage-interface/spec/issues/482

Signed-off-by: Yug Gupta <yuggupta27@gmail.com>

### DIFF
--- a/driver/node.go
+++ b/driver/node.go
@@ -170,9 +170,6 @@ func (s *nodeService) nodePublishFilesystemVolume(req *csi.NodePublishVolumeRequ
 	// avoid duplicate UUIDs
 	if mountOption.FsType == "xfs" {
 		mountOptions = append(mountOptions, "nouuid")
-		s.mounter.Exec.Command("xfs_admin", "-U", "generate", device)
-	} else if mountOption.FsType == "ext4" {
-		s.mounter.Exec.Command("tune2fs", "-U", "random", device)
 	}
 
 	mounted, err := filesystem.IsMounted(device, req.GetTargetPath())


Yuggupta27:
Other CSI drivers also handle this issue in a similar way:
Reference:
1. [aws-ebs-csi-driver](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/913/files#diff4f4d9c2e0ec5c4d3b4ac2d79abd70ee21bacff65ba05c059396921996dbf6607R671-R677)
2. [alibaba-cloud-csi-driver](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pull/570/files#diff-38755914ed4fc5af4f1f8d4bc26f54054642bd93458ba611bb78d264e98f25e5R110-R111)
3. [cinder-csi](https://github.com/kubernetes/cloud-provider-openstack/pull/1614/files)
4. [vsphere-csi-driver](https://github.com/kubernetes-sigs/vsphere-csi-driver/pull/1204/files)

nbalacha (approved):

toshipp (approved):

Issue #502, opened 2022-06-22
mutating webhook denies empty storageclass name


rmartinez3:
**Describe the bug**
Current topolvm version denies creation of pvc with empty storage class or binding of pvc if was already created.

Overall have been running topolvm and recently upgraded to 10.6
Noticed that code restructure was done as when this was merged
https://github.com/topolvm/topolvm/pull/182

did not see issues.

**Environments**
- Version: 10.6

**To Reproduce**
1.  Create pv  with empty storageclass

```yaml
apiVersion: v1
kind: PersistentVolume
metadata:
  name: foo-pv
spec:
  storageClassName: ""
  capacity:
    storage: 1Gi
  accessModes:
  - ReadWriteMany
  persistentVolumeReclaimPolicy: Retain
  hostPath:
    path: /tmm/dir
````
    
2. Create pvc to use pv 
```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: test-vol
  namespace: default
spec:
  storageClassName: ""
  volumeName:  foo-pv
  accessModes:
    - ReadWriteMany
  resources:
    requests:
      storage: 250Mi
```
3. Error 
"Error from server: error when creating "test-vol.yaml": admission webhook "pvc-hook.topolvm.cybozu.com" denied the request: resource name may not be empty"

**Expected behavior**
persistentvolumeclaim/test-vol created

**Additional context**
This as well fails if a pvc was already create previously and pod is trying to mount to pod . Same error as well.
" Error creating: admission webhook "pod-hook.topolvm.cybozu.com" denied the request: resource name may not be empty"

Overall issue is that 
https://github.com/topolvm/topolvm/blob/main/hook/mutate_persistentvolumeclaim.go#L47
https://github.com/topolvm/topolvm/blob/main/hook/mutate_pod.go#L191

checks for nil though in this case storagclass string is there but as empty .
Have done a patch on my end to check for nil as well then check for ""

Overall if this is fine can do MR to fix overall issue as without webhook. k8s allows of creation of these type of pv/pvc
https://kubernetes.io/docs/concepts/storage/persistent-volumes/#reserving-a-persistentvolume

peng225:
Thank you for your bug report. Could you send a PR?

daichimukai:
I'll take over this.

PR #525, opened 2022-07-06
webhook: allow PVCs with storageClassName set to ""

daichimukai:
A PVC is allowed to set `.spec.storageClassName` equal to "" (empty
string) to bound it to a PV with no storage class. In the current
implementation, for such volumes admission webhook tries to get storage
class with name of "", which fails with the error like "resource name
may not be empty". This error is different from "not found" error and
not well handled, which results in the admission webhook returns an
internal server error.

This patch changes the admission webhook to allow PVCs with
`.spec.storageClassName` set to "" since topolvm does not need to care
about such PVCs.

Fixes: #502

### DIFF
--- a/hook/suite_test.go
+++ b/hook/suite_test.go
@@ -36,15 +36,14 @@ var testEnv *envtest.Environment
 var testCtx, testCancel = context.WithCancel(context.Background())
 
 const (
+	emptyStorageClassName                       = ""
 	topolvmProvisionerStorageClassName          = "topolvm-provisioner"
 	topolvmProvisioner2StorageClassName         = "topolvm-provisioner2"
 	topolvmProvisioner3StorageClassName         = "topolvm-provisioner3"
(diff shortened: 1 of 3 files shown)


hook/suite_test.go
```diff
@@ -35,16 +35,15 @@ var k8sClient client.Client
 var testEnv *envtest.Environment
 var testCtx, testCancel = context.WithCancel(context.Background())
 
-const (
+var (
```
toshipp, line 38: Can we use const instead?

toshipp: I see. This is a test code, so I think using var is acceptable.

daichimukai: After thinking about this for a while, I changed my mind. `const` reduces the chance of programming errors and makes test more robust. It is more important than it looks. I've pushed revised version.

peng225 (approved):

llamerada-jp (approved):
LGTM

PR #540, opened 2022-07-26
fix: ResizeLV does not check thinpool overprovisioned size

nbalacha:
The ResizeLV function did not use the available overprovisioned
capacity in the size checks for thin-provisioned devices. This caused
resizing a thin LV to fail when the VG did not have sufficient
space but the overprovisioned capacity did.

Signed-off-by: N Balachandran <nibalach@redhat.com>

### DIFF
--- a/lvmd/lvservice.go
+++ b/lvmd/lvservice.go
@@ -306,14 +306,39 @@ func (s *lvService) ResizeLV(_ context.Context, req *proto.ResizeLVRequest) (*pr
 		return nil, status.Error(codes.OutOfRange, "shrinking volume size is not allowed")
 	}
 
-	free, err := vg.Free()
-	if err != nil {
-		log.Error("failed to free VG", map[string]interface{}{
-			log.FnError: err,
-			"name":      req.GetName(),
-		})
-		return nil, status.Error(codes.Internal, err.Error())
+	free := uint64(0)
+	var pool *command.ThinPool
+	switch dc.Type {
+	case TypeThick:
+		free, err = vg.Free()
+		if err != nil {
+			log.Error("failed to get free bytes", map[string]interface{}{
+				log.FnError: err,
+			})
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+	case TypeThin:
+		pool, err = vg.FindPool(dc.ThinPoolConfig.Name)
+		if err != nil {
+			log.Error("failed to get thinpool", map[string]interface{}{
+				log.FnError: err,
+			})
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+		tpu, err := pool.Free()
+		if err != nil {
+			log.Error("failed to get free bytes", map[string]interface{}{
+				log.FnError: err,
+			})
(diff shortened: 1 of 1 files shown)


lvmd/lvservice.go
```diff
@@ -306,14 +306,39 @@ func (s *lvService) ResizeLV(_ context.Context, req *proto.ResizeLVRequest) (*pr
...
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+		tpu, err := pool.Free()
+		if err != nil {
+			log.Error("failed to get free bytes", map[string]interface{}{
+				log.FnError: err,
+			})
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+		free = uint64(math.Floor(dc.ThinPoolConfig.OverprovisionRatio*float64(tpu.SizeBytes))) - tpu.VirtualBytes
```
nbalacha, line 335: I would like to move this into a separate function as it is being called multiple times. What is the best file/pkg for this?

satoru-takeuchi: It's OK to place the new function in `lvmd/lvservice.go` since this code is only called from this file.

satoru-takeuchi:
@nbalacha I'll approve this PR after you add a new function described in the above comment.

lvmd/lvservice.go
nbalacha: For some reason the tests fail when I move this to a function. Looking into it further but if I cannot find out why, can we continue with this PR for now?

satoru-takeuchi: OK, no problem.

satoru-takeuchi (approved):

lvmd/lvservice.go
leelavg: - iirc, this is used in vgservice.go as well.
- earlier the calculation used was plannrd to be part of thinpool struct
- after a comment that command.go should only contain struct info the idea was dropped

lvmd/lvservice.go
```diff
@@ -306,14 +306,39 @@ func (s *lvService) ResizeLV(_ context.Context, req *proto.ResizeLVRequest) (*pr
...
-		})
-		return nil, status.Error(codes.Internal, err.Error())
+	free := uint64(0)
+	var pool *command.ThinPool
+	switch dc.Type {
+	case TypeThick:
+		free, err = vg.Free()
+		if err != nil {
+			log.Error("failed to get free bytes", map[string]interface{}{
+				log.FnError: err,
```
daichimukai, line 316: Is `"name": req.GetName()` intentionally dropped?

nbalacha: I copied the code block from https://github.com/topolvm/topolvm/blob/5df9e7cede433d41c9c77827b0f26dacfa8a2a59/lvmd/lvservice.go#L52

I can add the "name" field if required.

daichimukai: Understood. Thanks.

daichimukai (approved):

PR #545, opened 2022-08-04
Update e2e tests to handle snapshot csi tests

nbalacha:
This commit provides a way to configure the e2e tests to run csi snapshot
tests only for thin device classes.

Signed-off-by: N Balachandran <nibalach@redhat.com>

### DIFF
--- a/.github/workflows/e2e-k8s-workflow.yaml
+++ b/.github/workflows/e2e-k8s-workflow.yaml
@@ -7,6 +7,9 @@ on:
       skip_node_finalize:
         default: "false"
         type: string
+      thin_csi_sanity:
+        default: "false"
+        type: string
 
 jobs:
   e2e-k8s:
@@ -20,6 +23,7 @@ jobs:
       KUBERNETES_VERSION: ${{ matrix.kubernetes_versions }}
       TEST_SCHEDULER_MANIFEST: ${{ inputs.test_scheduler_manifest }}
       SKIP_NODE_FINALIZE: ${{ inputs.skip_node_finalize }}
+      TEST_THIN_DEVICECLASS: ${{ inputs.thin_csi_sanity }}
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v3
--- a/e2e/sanity_test.go
+++ b/e2e/sanity_test.go
@@ -3,6 +3,8 @@ package e2e
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
+	"os"
 	"path"
 
 	"github.com/kubernetes-csi/csi-test/v4/pkg/sanity"
@@ -32,6 +34,7 @@ func testSanity() {
 			if err != nil {
 				return err
 			}
+			fmt.Println("topolvm-node", "ds.Status.NumberAvailable", ds.Status.NumberAvailable)
 			if ds.Status.NumberAvailable != 1 {
 				return errors.New("node daemonset is not ready")
 			}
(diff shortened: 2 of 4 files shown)


nbalacha:
@llamerada-jp , this is the initial set of changes for enabling the csi sanity snapshot tests for thin deviceclasses. Please let me know if this is what you had in mind. 
I can then update the workflow like this:
`--- a/.github/workflows/e2e-k8s-workflow.yaml
+++ b/.github/workflows/e2e-k8s-workflow.yaml
@@ -16,10 +16,12 @@ jobs:
       fail-fast: false
       matrix:
         kubernetes_versions: ["1.24.2", "1.23.6", "1.22.9"]
+        snapshot_tests: ["true", "false"]
     env:
       KUBERNETES_VERSION: ${{ matrix.kubernetes_versions }}
       TEST_SCHEDULER_MANIFEST: ${{ inputs.test_scheduler_manifest }}
       SKIP_NODE_FINALIZE: ${{ inputs.skip_node_finalize }}
+      TEST_THIN_DEVICECLASS: ${{ matrix.snapshot_tests }}
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v3
`

e2e/sanity_test.go
```diff
@@ -62,5 +65,13 @@ func testSanity() {
 		_, _, err = kubectl("exec", "-n", "topolvm-system", "daemonset/topolvm-node", "--", "test", "-e", path)
 		return sanity.PathIsOther, err
 	}
+
+	if os.Getenv("TEST_THIN_DEVICECLASS") == "true" {
+		// csi.storage.k8s.io/fstype=xfs,topolvm.cybozu.com/device-class=sbvglvm
```
peng225, line 70: What is `sbvglvm`? (The device set is actually set to `thin` at line 73.)

e2e/sanity_test.go
```diff
@@ -62,5 +65,13 @@ func testSanity() {
...
+		volParams := make(map[string]string)
+		volParams["csi.storage.k8s.io/fstype"] = "xfs"
+		volParams["topolvm.cybozu.com/device-class"] = "thin"
+		tc.TestVolumeParameters = volParams
```
peng225, line 74: I'm afraid I could not find any documents about `TestVolumeParameters`.
I would be happy if you could share the document you referred to when you write that code.

llamerada-jp:
@nbalacha 
The way I requested it before was incorrect. Would you please add a new test parameter like below?
the test of snapshot will be skipped in default (equal to the current patern) and will be executed in the new case.

https://github.com/topolvm/topolvm/blob/a00f86f7183260982988af771c487b8678822b0f/.github/workflows/e2e-k8s.yaml#L14-L26

https://github.com/topolvm/topolvm/blob/a00f86f7183260982988af771c487b8678822b0f/.github/workflows/e2e-k8s-workflow.yaml#L2-L9

e2e/sanity_test.go
nbalacha: I found it in the csi-test code:

https://github.com/kubernetes-csi/csi-test/blob/862e81af7f4b61e965714f8c528ee8ca91780dd7/pkg/sanity/sanity.go#L53

This describes the various parameters that can be specified.

e2e/sanity_test.go
nbalacha: I'll fix this.

nbalacha:
> @nbalacha The way I requested it before was incorrect. Would you please add a new test parameter like below? the test of snapshot will be skipped in default (equal to the current patern) and will be executed in the new case.
> 
> https://github.com/topolvm/topolvm/blob/a00f86f7183260982988af771c487b8678822b0f/.github/workflows/e2e-k8s.yaml#L14-L26
> 
> https://github.com/topolvm/topolvm/blob/a00f86f7183260982988af771c487b8678822b0f/.github/workflows/e2e-k8s-workflow.yaml#L2-L9

Thank you @llamerada-jp . I will make the changes.

e2e/sanity_test.go
nbalacha: Fixed.

.github/workflows/e2e-k8s-workflow.yaml
```diff
@@ -16,10 +19,12 @@ jobs:
       fail-fast: false
       matrix:
         kubernetes_versions: ["1.24.2", "1.23.6", "1.22.9"]
+        snapshot_tests: ["true", "false"]
```
peng225, line 22: This parameter is not used anywhere.

nbalacha: Thank you. Fixed.

peng225 (approved):

llamerada-jp (approved):

Issue #559, opened 2022-09-02
lvmd pod can't start


alanyanglong:
2022-09-02T02:41:16.344639Z topolvm-lvmd-0-m85zp lvmd info: "configuration file loaded: " device_classes="[0xc00034fc00]" file_name="/etc/topolvm/lvmd.yaml" socket_name="/run/topolvm/lvmd.sock"

2022-09-02T02:41:16.370328Z topolvm-lvmd-0-m85zp lvmd error: "Error while retrieving volume groups"
Error: strconv.ParseUint: parsing "-1": invalid syntax

strconv.ParseUint: parsing "-1": invalid syntax

What's the issue? and How to fix it ?    Thanks

llamerada-jp:
According to your message, lvmd.yaml containing an wrong configuration value. It seems to be set to -1 where it should be a positive integer.

alanyanglong:
> According to your message, lvmd.yaml containing an wrong configuration value. It seems to be set to -1 where it should be a positive integer.

root@topolvm-lvmd-0-6sfbx:/# /lvmd --container
2022-09-02T09:29:13.435193Z topolvm-lvmd-0-6sfbx lvmd info: "configuration file loaded: " device_classes="[0xc000125f00]" file_name="/etc/topolvm/lvmd.yaml" socket_name="/run/topolvm/lvmd.sock"
2022-09-02T09:29:13.464737Z topolvm-lvmd-0-6sfbx lvmd error: "Error while retrieving volume groups"
Error: strconv.ParseUint: parsing "-1": invalid syntax
strconv.ParseUint: parsing "-1": invalid syntax
root@topolvm-lvmd-0-6sfbx:/# cat /etc/topolvm/lvmd.yaml
socket-name: /run/topolvm/lvmd.sock
device-classes: 
  - default: true
    name: ssd
    spare-gb: 10
    volume-group: vg1

daichimukai:
I suspect that the output of lvm command is not what lvmd expect. Could you share the output of the following command?
```
[...]

PR #561, opened 2022-09-07
Correct LV JSON major/minor

tasleson:
Resolves: https://github.com/topolvm/topolvm/issues/559

### DIFF
--- a/lvmd/command/lvm_json.go
+++ b/lvmd/command/lvm_json.go
@@ -95,16 +95,12 @@ func (u *lv) UnmarshalJSON(data []byte) error {
 	u.uuid = temp.UUID
 	u.path = temp.Path
 
-	var convErr error
-	u.major, convErr = strconv.ParseUint(temp.Major, 10, 32)
-	if convErr != nil {
-		return convErr
-	}
-	u.minor, convErr = strconv.ParseUint(temp.Minor, 10, 32)
-	if convErr != nil {
-		return convErr
-	}
+	// If LV is not active, major/minor numbers will be -1, ignore conversion
+	// errors which results in 0 for values in this case.
+	u.major, _ = strconv.ParseUint(temp.Major, 10, 32)
+	u.minor, _ = strconv.ParseUint(temp.Minor, 10, 32)
 
+	var convErr error
 	u.origin = temp.Origin
 	if len(temp.OriginSize) > 0 {
 		u.originSize, convErr = strconv.ParseUint(temp.OriginSize, 10, 64)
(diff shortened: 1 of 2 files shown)


lvmd/command/lvm_json.go
```diff
@@ -95,16 +95,12 @@ func (u *lv) UnmarshalJSON(data []byte) error {
...
-		return convErr
-	}
-	u.minor, convErr = strconv.ParseUint(temp.Minor, 10, 32)
-	if convErr != nil {
-		return convErr
-	}
+	// If LV is not active, major/minor numbers will be -1, ignore conversion
+	// errors which results in 0 for values in this case.
+	u.major, _ = strconv.ParseUint(temp.Major, 10, 32)
+	u.minor, _ = strconv.ParseUint(temp.Minor, 10, 32)
```
llamerada-jp, line 101: If only one of the numbers is -1, it is a different case from the issue. Would you please terminate the program in this case? And please add outputting a log if both of the numbers are -1.

llamerada-jp (changes requested):

lvmd/command/lvm_json.go
tasleson: Do you really want to take down the service in this case?  Maybe just log it?  This change mimics previous behavior.

nbalacha: I agree with @tasleson . Logging this would be better than having lvmd go into CrashLoopBackOff. 
We can also check with the LVM team if a scenario where only one number set to -1 can occur.

nbalacha (approved):

lvmd/command/lvm_json.go
llamerada-jp: @tasleson @nbalacha 
I'm aware that the current PR behaves the same as the version before using JSON. I have asked for the following reason.
I could not check the case is not exist where (major, minor)=(n, -1). In that case, I think we have a risk to access another LV because (major, minor)=(n, 0) is an acceptable device. If this PR is not appropriate for this change, I will make another PR. What do you think?

nbalacha: Can you please make the change in another PR?

tasleson: @llamerada-jp @nbalacha From chatting with one of the lvm developers, the consensus is they will both be `-1` or both be valid major/minor numbers.

llamerada-jp: @nbalacha @tasleson 
I see. Thank you for checking.

llamerada-jp (approved):

PR #563, opened 2022-09-13
fix: add lvmd env option to helm chart

### LINKED ISSUE #562: Read only file system fix; `LVM_SYSTEM_DIR=/tmp`
On a read-only filesystem, where `/etc/lvm` is read-only, such as in [Talos Linux](https://www.talos.dev/), the `lvcreate` operations will fail with exit status 5. To fix this, you can set the env `LVM_SYSTEM_DIR=/tmp`, on the lvmd daemon pod. 

Currently there is no easy way to set this in the deployment, as the helm chart doesn't take custom envs. However, perhaps Topolvm could autodetect a read only filesystem and automatically set the env when necessary

Davincible:
fix #562


llamerada-jp (changes requested):
Would you please regenerate documents by `make generate`?

Davincible:
Like this ?

llamerada-jp (approved):

Davincible:
@pluser

pluser (approved):

Davincible:
Could you also update the helm chart?

cupnes:
@Davincible helm chart will be updated on beginning of next month.

PR #539, opened 2022-07-23
Rename the group name: topolvm.io

bells17:
- #168

PR #567, opened 2022-09-23
fix: rbac rule per release

fengxsong:
Signed-off-by: fengxsong <fengxsong@outlook.com>

Rename RBAC rules to avoid conflicts when trying to deploy more than one topolvm instances in single cluster.

### DIFF
--- a/charts/topolvm/templates/controller/clusterrolebindings.yaml
+++ b/charts/topolvm/templates/controller/clusterrolebindings.yaml
@@ -1,59 +1,61 @@
+{{- $rlsNamespace := .Release.Namespace -}}
+
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: {{ .Release.Namespace }}:controller
+  name: {{ printf "%s-controller" $rlsNamespace }}
+  labels:
+    {{- include "topolvm.labels" . | nindent 4 }}
+subjects:
+  - kind: ServiceAccount
+    namespace: {{ $rlsNamespace }}
+    name: {{ printf "%s-controller" $rlsNamespace }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ printf "%s-controller" $rlsNamespace }}
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: {{ printf "%s-external-provisioner" $rlsNamespace }}
   labels:
     {{- include "topolvm.labels" . | nindent 4 }}
 subjects:
   - kind: ServiceAccount
     namespace: {{ .Release.Namespace }}
-    name: {{ template "topolvm.fullname" . }}-controller
+    name: {{ printf "%s-controller" $rlsNamespace }}
 roleRef:
-  apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: {{ .Release.Namespace }}:controller
+  name: {{ printf "%s-external-provisioner" $rlsNamespace }}
+  apiGroup: rbac.authorization.k8s.io
 ---
 kind: ClusterRoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: topolvm-csi-provisioner-role
+  name: {{ printf "%s-csi-resizer" $rlsNamespace }}
   labels:
     {{- include "topolvm.labels" . | nindent 4 }}
 subjects:
   - kind: ServiceAccount
     namespace: {{ .Release.Namespace }}
-    name: {{ template "topolvm.fullname" . }}-controller
+    name: {{ printf "%s-controller" $rlsNamespace }}
 roleRef:
   kind: ClusterRole
-  name: topolvm-external-provisioner-runner
+  name: {{ printf "%s-csi-resizer" $rlsNamespace }}
   apiGroup: rbac.authorization.k8s.io
 ---
 kind: ClusterRoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: topolvm-csi-resizer-role
-  labels:
-    {{- include "topolvm.labels" . | nindent 4 }}
-subjects:
-  - kind: ServiceAccount
-    namespace: {{ .Release.Namespace }}
-    name: {{ template "topolvm.fullname" . }}-controller
-roleRef:
-  kind: ClusterRole
-  name: topolvm-external-resizer-runner
-  apiGroup: rbac.authorization.k8s.io
----
-kind: ClusterRoleBinding
-apiVersion: rbac.authorization.k8s.io/v1
-metadata:
-  name: topolvm-csi-snapshotter-role
+  name: {{ printf "%s-external-snapshotter" $rlsNamespace }}
   labels:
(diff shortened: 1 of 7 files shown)


toshipp:
Hi @fengxsong,
I confirmed the problem. But we do not support multiple topolvm controllers on a single namespace, I'd like to use a namespace name for object names because the namespace name is unique on a kubernetes cluster.

fengxsong:
> Hi @fengxsong, I confirmed the problem. But we do not support multiple topolvm controllers on a single namespace, I'd like to use a namespace name for object names because the namespace name is unique on a kubernetes cluster.

even if using different namespaces, clusterrole/clusterrolebinding will still conflicts.

toshipp:
This ClusterRole name contains a namespace name, two ClusterRoles will have different names, IMO.
https://github.com/topolvm/topolvm/blob/fcb270f38c1342a03b5850755935792471220326/charts/topolvm/templates/controller/clusterrolebindings.yaml#L4

charts/topolvm/templates/controller/clusterrolebindings.yaml
```diff
@@ -1,59 +1,61 @@
+{{- $rlsName := (include "topolvm.fullname" .) -}}
+
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: {{ .Release.Namespace }}:controller
+  name: {{ printf "%s-controller" $rlsName }}
```
fengxsong, line 6: @toshipp  do u mean revert all those `{{ .Release.Namespace }}:controller`?

toshipp: Yes, and also other cluster resource names will be `{{ .Release.Namespace }}-xxx`.
BTW, we should use DNS acceptable name due to [this document](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/).
So, the line avobe should be `{{ .Release.Namespace }}-controller`.

How do you think? @cupnes

cupnes: @toshipp I think so, too. There seem to be no other good ideas.

fengxsong: > Yes, and also other cluster resource names will be `{{ .Release.Namespace }}-xxx`. BTW, we should use DNS acceptable name due to [this document](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/). So, the line avobe should be `{{ .Release.Namespace }}-controller`.
> 
> How do you think? @cupnes

done. btw maybe documented `But we do not support multiple topolvm controllers on a single namespace` in README or somewhere else?

toshipp: I realized topolvm is a CSI plugin and the plugin name is unique, therefore, we can not install multiple topolvm in a normal situation. (we now try to change the plugin name on #539, if it is done, we can deploy the old and the new one simultaneously.)

Why do you try this? and is it really needed?

fengxsong: > I realized topolvm is a CSI plugin and the plugin name is unique, therefore, we can not install multiple topolvm in a normal situation. (we now try to change the plugin name on #539, if it is done, we can deploy the old and the new one simultaneously.)
> 
> 
> 
> Why do you try this? and is it really needed?

We have different types(size and read/write speed) of local nvme workers. And want to run some of middlewares that require local storage on a set of nodes. Meanwhile don't want to add traints and tolerations. So using different storageclass to meet the topology requirement. 🙃

toshipp: Thank you for the explanation.
Using one topolvm deployment, We support multiple SCs with device classes. So, IIUC, your case can already be achieved.
https://github.com/topolvm/topolvm/blob/511319a8e73a0c05993f41461d019a945f5a91fc/charts/topolvm/values.yaml#L447-L464
Could you show me your values.yamls?

fengxsong: here's out values.yaml

```
USER-SUPPLIED VALUES:
controller:
  storageCapacityTracking:
    enabled: true
lvmd:
  deviceClasses:
  - default: true
    name: topolvm-default
    spare-gb: 10
    volume-group: topolvm-default
  - name: topolvm-ssd
    spare-gb: 1
    volume-group: topolvm-ssd
  managed: true
node:
  kubeletWorkDirectory: /var/lib/kubelet
  metrics:
    annotations:
      prometheus.io/port: metrics
    enabled: true
scheduler:
  enabled: false
storageClasses:
- name: topolvm-default
  storageClass:
    allowVolumeExpansion: true
    annotations: {}
    fsType: xfs
    isDefaultClass: true
    reclaimPolicy: null
    volumeBindingMode: WaitForFirstConsumer
- name: topolvm-ssd
  storageClass:
    additionalParameters:
      topolvm.cybozu.com/device-class: topolvm-ssd
    allowVolumeExpansion: true
    fsType: xfs
    isDefaultClass: false
    reclaimPolicy: null
    volumeBindingMode: WaitForFirstConsumer
webhook:
  podMutatingWebhook:
    enabled: false
```

toshipp: Thank you.
IIUC, you have another values.yaml for another topolvm release, Could you show that?

fengxsong: let me explain more details, we use `topolvm-default` to provision PVs for workloads like kafka/minio etc, and `topolvm-ssd` for influxdb. The first set of workers has those two device-classes. And we now need to add another set of machines that only run tdengine instances, more specified, only provision PVs for tdengine(**requirement**), using one topolvm deployment can not guarantee that requirement.

fengxsong: Although `taint` and `toleration` is the best solution, running micro services(that only require computer resource) on the second set is allowed, we don't want to add tolerations for every micro service deployments.

fengxsong: > Thank you. IIUC, you have another values.yaml for another topolvm release, Could you show that?

since two deployment don't works as we expected, we've already uninstall it. only `storageClasses` and `lvmd. deviceClasses` changed.

toshipp: I understand what you want, thanks.

>since two deployment don't works as we expected, we've already uninstall it. 
Does this mean it does not work even if this PR is accepted?

To solve your problem, add a new device class for tdengine instances and configure lvmds on the instances to use the device class, then pods using new SC specifying the device class will be deployed on the target instances, IMO.
It is not able to do by using the current helm chart, so we configure and deploy lvmd manually.
Could you try it?

fengxsong: "Does this mean it does not work even if this PR is accepted?", I'm afraid not. but in order to support more than one topolvm deployment in one cluster, those names of clusterrole/clusterrolebinding must not use constants.

I didn't have the chance to try again， cause it's already in production stage. -。-

fengxsong: Yes! feel free to do so. :)

toshipp: Thank you so much!

toshipp:
This PR is merged into #539

PR #582, opened 2022-10-23
helm: Update templates for Container specific resources

laxmanvallandas:


charts/topolvm/values.yaml
```diff
@@ -290,16 +300,6 @@ node:
       #   cpu: "200m"
       #   memory: "200Mi"
 
-  # node.resources -- Specify resources.
-  ## ref: https://kubernetes.io/docs/user-guide/compute-resources/
-  resources: {}
```
toshipp, line 295: We think adding the `containers` field looks wordy, would you assemble all container resources under this section as follows?
```
resources:
  topolvm_node:
    requests:
     ...
  csi_registrar
```

laxmanvallandas: Done 62f221d08bed2121f9b576fef7ff17fb7b0fcbea

charts/topolvm/templates/controller/deployment.yaml
```diff
@@ -74,8 +74,9 @@ spec:
           {{- with .Values.image.pullPolicy }}
           imagePullPolicy: {{ . }}
           {{- end }}
-          {{- with .Values.controller.resources }}
-          resources: {{ toYaml . | nindent 12 }}
+          {{- if .Values.resources.topolvm_controller }}
+          resources:
+            {{- toYaml .Values.resources.topolvm_controller | nindent 12 -}}
```
toshipp, line 79: We use `with` command for this `if` pattern. Could you alter this and others you changed?

toshipp (changes requested):

charts/topolvm/templates/controller/deployment.yaml
laxmanvallandas: Ahh. That's right. Seems like I pushed some trial changes. Thanks. I re-pushed. 
Can you take a look again?

toshipp (approved):
LGTM

llamerada-jp (approved):
LGTM

PR #618, opened 2022-12-05
improve livenessProbe get csi-node-driver-registrar healthz by httpGet

huangzynn:
Use health-check-with-an-exec-probe vs use health-check-with-an-http-get
https://github.com/kubernetes-csi/node-driver-registrar/tree/release-2.5#health-check-with-an-exec-probe
when use health-check-with-an-exec-probe, csi-node-driver-registrar lost connection,health-check-with-an-exec-probe cannot determine whether the socket exists. but when use http-get, it can determine whether the socket exists. code is different.
health-check-with-an-exec-probe:
https://github.com/kubernetes-csi/node-driver-registrar/blob/release-2.5/cmd/csi-node-driver-registrar/main.go#L152
health-check-with-an-http-get:
https://github.com/kubernetes-csi/node-driver-registrar/blob/release-2.5/cmd/csi-node-driver-registrar/node_register.go#L99
when node-driver-registrar is running , exec it remove /registration/{{ include "topolvm.pluginName" . }}-reg.sock. You'll find out

### DIFF
--- a/charts/topolvm/templates/node/daemonset.yaml
+++ b/charts/topolvm/templates/node/daemonset.yaml
@@ -98,13 +98,16 @@ spec:
             - /csi-node-driver-registrar
             - --csi-address={{ .Values.node.kubeletWorkDirectory }}/plugins/{{ include "topolvm.pluginName" . }}/node/csi-topolvm.sock
             - --kubelet-registration-path={{ .Values.node.kubeletWorkDirectory }}/plugins/{{ include "topolvm.pluginName" . }}/node/csi-topolvm.sock
+            - --health-port=9809
+          ports:
+            - containerPort: 9809
+              name: healthz
           livenessProbe:
-            exec:
-              command:
-              - /csi-node-driver-registrar
-              - --kubelet-registration-path={{ .Values.node.kubeletWorkDirectory }}/plugins/{{ include "topolvm.pluginName" . }}/node/csi-topolvm.sock
-              - --mode=kubelet-registration-probe
-            initialDelaySeconds: 3
+            httpGet:
+              path: /healthz
+              port: healthz
+            initialDelaySeconds: 5
+            timeoutSeconds: 5
           lifecycle:
             preStop:
               exec:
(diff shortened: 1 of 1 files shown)


charts/topolvm/templates/node/daemonset.yaml
```diff
@@ -98,13 +98,16 @@ spec:
             - /csi-node-driver-registrar
             - --csi-address={{ .Values.node.kubeletWorkDirectory }}/plugins/{{ include "topolvm.pluginName" . }}/node/csi-topolvm.sock
             - --kubelet-registration-path={{ .Values.node.kubeletWorkDirectory }}/plugins/{{ include "topolvm.pluginName" . }}/node/csi-topolvm.sock
+            - --health-port=9809
```
peng225, line 101: `--health-port` is deprecated. Please use `--http-endpoint` instead.
https://github.com/kubernetes-csi/node-driver-registrar/tree/release-2.5#optional-arguments

charts/topolvm/templates/node/daemonset.yaml
```diff
@@ -98,13 +98,16 @@ spec:
...
-              command:
-              - /csi-node-driver-registrar
-              - --kubelet-registration-path={{ .Values.node.kubeletWorkDirectory }}/plugins/{{ include "topolvm.pluginName" . }}/node/csi-topolvm.sock
-              - --mode=kubelet-registration-probe
-            initialDelaySeconds: 3
+            httpGet:
+              path: /healthz
+              port: healthz
+            initialDelaySeconds: 5
+            timeoutSeconds: 5
```
peng225, line 110: Do you have any reasons why you choosed 5 seconds for `timeoutSeconds`?
If you do not have strong reasons, it might be good to set it to 3 because we often set it to 3 in TopoLVM repository.
https://github.com/topolvm/topolvm/search?q=timeoutSeconds

huangzynn: Thank you for your advice

PR #619, opened 2022-12-07
improve livenessProbe get csi-node-driver-registrar healthz by httpGet 

huangzynn:
Use health-check-with-an-exec-probe vs use health-check-with-an-http-get
https://github.com/kubernetes-csi/node-driver-registrar/tree/release-2.5#health-check-with-an-exec-probe
when use health-check-with-an-exec-probe, csi-node-driver-registrar lost connection,health-check-with-an-exec-probe cannot determine whether the socket exists. but when use http-get, it can determine whether the socket exists. code is different.
health-check-with-an-exec-probe:
https://github.com/kubernetes-csi/node-driver-registrar/blob/release-2.5/cmd/csi-node-driver-registrar/main.go#L152
health-check-with-an-http-get:
https://github.com/kubernetes-csi/node-driver-registrar/blob/release-2.5/cmd/csi-node-driver-registrar/node_register.go#L99
when node-driver-registrar is running , exec it remove /registration/{{ include "topolvm.pluginName" . }}-reg.sock. You'll find out


peng225 (approved):

peng225:
cf. old PR.
https://github.com/topolvm/topolvm/pull/618

cupnes (approved):

PR #625, opened 2022-12-15
Allow multiple device-classes with same VG

glorpen:
Device-classes with repeated VG are now accepted, so you can create classes with different lv raid types (raid0, raid1) to use same VG.

Example:

```yaml
device-classes:
  - default: false
    name: "ssd-raid0"
    spare-gb: 10
    volume-group: "vg0"
    lvcreate-options: [ "--type=raid0" ]

  - default: true
    name: "ssd-raid1"
    spare-gb: 10
    volume-group: "vg0"
    lvcreate-options: [ "--type=raid1" ]
```

### DIFF
--- a/lvmd/device_class_manager.go
+++ b/lvmd/device_class_manager.go
@@ -72,7 +72,7 @@ func ValidateDeviceClasses(deviceClasses []*DeviceClass) error {
 	}
 	var countDefault = 0
 	dcNames := make(map[string]bool)
-	vgNames := make(map[string]bool)
+	thinNames := make(map[string]bool)
 	for _, dc := range deviceClasses {
 		if len(dc.Name) == 0 {
 			return errors.New("device-class name should not be empty")
@@ -118,14 +118,15 @@ func ValidateDeviceClasses(deviceClasses []*DeviceClass) error {
 			// combination of volumegroup and thinpool should be unique across device classes
 			// so the key 'name' shouldn't appear twice to verify it's uniqueness
 			name = name + "/" + dc.ThinPoolConfig.Name
-		}
 
-		if vgNames[name] {
-			return fmt.Errorf("duplicate volumegroup/thinpool name: %s, %s", dc.Name, name)
+			if thinNames[name] {
+				return fmt.Errorf("duplicate volumegroup/thinpool name: %s, %s", dc.Name, name)
+			}
+
+			thinNames[name] = true
 		}
 
 		dcNames[dc.Name] = true
-		vgNames[name] = true
 		if dc.StripeSize != "" && !stripeSizeRegexp.MatchString(dc.StripeSize) {
 			return fmt.Errorf("stripe-size format is \"Size[k|UNIT]\": %s", dc.Name)
 		}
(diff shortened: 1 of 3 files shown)


toshipp:
I understand your motivation, however, I am concerned about the calculation of capacity.
For example, if a VG has 100G capacity and device-class A which uses the VG has `spare-vg` 10 and B which also uses the VG has `spare-vg` 20, we are going to create a 90G PV from device-class A, it will succeed. But, this breaks invariant from device-class B which retains 20G capacity for the VG.

glorpen:
I don't know about internals of topolvm nor CSI drivers (and too much about writing in golang if we are at it), so bear with me :)

In your example, what will happen if `vgService.send` suddenly reports device-class B having 0 free space, or it will not be updated at all in k8s/CSI because device-class B was not touched? Wouldn't be it same case as for when admin creates LV 100%FREE outside topolvm since VG can be shared with system?

Maybe simplest solution could be to make all drive-classes with shared VG have `spare-gb=max(spare-gb's from this VG)` and maybe even emit warning about it. But (currently) it seems to me that either way free space will be reported as 0, even without this change.

toshipp:
> I don't know about internals of topolvm nor CSI drivers (and too much about writing in golang if we are at it), so bear with me :)
> 
> In your example, what will happen if `vgService.send` suddenly reports device-class B having 0 free space, or it will not be updated at all in k8s/CSI because device-class B was not touched?

TopoLVM reports 0 free space.

> Wouldn't be it same case as for when admin creates LV 100%FREE outside topolvm since VG can be shared with system?

It has the same effect, but the administrator has the responsibility to maintain the reservation, if it happens, it means the administrator makes a mistake.
Otherwise, if it happens by topolvm, it is a bug of topolvm.

> Maybe simplest solution could be to make all drive-classes with shared VG have `spare-gb=max(spare-gb's from this VG)` and maybe even emit warning about it. But (currently) it seems to me that either way free space will be reported as 0, even without this change.

I describe retaining spare-gb for all device-classes are invariant above, it is not a solution.

BTW, If you want to use one VG with many lvcreate options, is it sufficient to add a new parameter to specify the options on SCs?

glorpen:
So... without virtually splitting VG in half and allocating each half to each entry it will always allow exhausting other one.

If by SC you mean StorageClass on k8s side, then adding param like `topolvm.io/lvcreate-options` to be later used in lvmd device-class, would solve it too. If you think it is a way to go I'll rewrite this PR.

toshipp:
>If by SC you mean StorageClass on k8s side, then adding param like topolvm.io/lvcreate-options to be later used in lvmd device-class, would solve it too. If you think it is a way to go I'll rewrite this PR.

It is what I think. 

BTW, I discussed other maintainers internally, it would be better to add a mapping layer to designate options rather than just setting options on a StorageClass like device-class. It can hide details of lvm options from kubernetes and change actual options on each node.

E.g.

lvmd:
```
device-class:
  - name: dc-a
     volume-group: vga
  - name: dc-b
     volume-group: vga
lvcreate-option-class:
  - name: opt-a
     options: ["--type=raid1"]
  - name: opt-b
     options: ["--type=raid2"]
```

Then specify the option class name on a StorageClass.
```
kind: StorageClass
...
  metadata:
    annotations:
      topolvm.io/lvcreate-option-class: opt-a
```

If you attempt to work, would you write a proposal first? or we can take over this.

glorpen:
You can take over - with my current knowledge it looks too complicated to be completed by me in sensible time :)

toshipp:
Ok. Thank you for your effort!

Issue #622, opened 2022-12-11
Incorrect default capacity reported on nodes for thin pool


ppmathis:
**Describe the bug**
The current TopoLVM release is reporting an incorrect default capacity (`capacity.topolvm.cybozu.com/00default`) on Kubernetes nodes when the default device class is a thin pool and the storage class does not explicitly specify it. The actual capacity that is being reported is the amount of free bytes of the volume group containing the thin pool, but not the amount of available space in the thin pool itself. When deploying a configuration where a thin pool spans 100% of the volume group, this ends up with no PVCs being deployed.

I am not really familiar with the TopoLVM code base, but after a quick glance I think that the capacity annotations are being done in [metrics_exporter.go](https://github.com/topolvm/topolvm/blob/f411453b49ac32e8157809b18602cf0f6d2e10bf/runners/metrics_exporter.go#L232-L241), where first the default annotation is applied followed by device-class specific information. As can be clearly seen, unlike the per-device-class annotation there is no special treatment for thin pools and `WatchResponse.FreeBytes` is always being used.

This message seems to be constructed within [vgservice.go](https://github.com/topolvm/topolvm/blob/main/lvmd/vgservice.go#L137-L215), where `FreeBytes` is always set to the amount of free space in the volume group. As in my case the volume group has no free space and its whole capacity has been dedicated to the thin pool, this ends up with TopoLVM reporting an incorrect default capacity:

```yaml
# This should be the same as ssd-thin, there are no other device classes configured
capacity.topolvm.cybozu.com/00default: 0

# This contains the expected value, due to the special treatment
# within metrics_exporter.go for per-device-class info
capacity.topolvm.cybozu.com/ssd-thin: 397242531840
```

**Environments**
- Version: 0.16.0 installed from Helm Chart 11.0.0
- OS: Debian 11.5 with 5.10.0-19-amd64

**To Reproduce**
1. Create a new volume group named `vg-ssd` on your test system
2. Create a new thin pool on `vg-ssd` named `tp-ssd` and assign 100% of free space to it
3. Deploy TopoLVM using Helm with these value overrides:
```yaml
controller:
  storageCapacityTracking:
    enabled: true

lvmd:
  managed: true
  socketName: /run/topolvm/lvmd.sock
  deviceClasses:
    - name: ssd-thin
      volume-group: vg-ssd
      default: true
      type: thin
      thin-pool:
        name: tp-ssd
        overprovision-ratio: 10.0

scheduler:
  enabled: false

[...]

PR #632, opened 2023-01-03
fix: set default capacity to thin class capacity when default

suleymanakbas91:
VGService.send should set the default capacity (`capacity.topolvm.cybozu.com/00default`) to the capacity of the thin pool device class when it is the default.

Fixes https://github.com/topolvm/topolvm/issues/622

Signed-off-by: Suleyman Akbas <sakbas@redhat.com>


cupnes:
Please update the comments on this line. Because, in the case of thin pools, overprovisioned bytes are set.
https://github.com/topolvm/topolvm/blob/b3602ddc5b3643b2370df40f124ba57c75bff2a8/lvmd/proto/lvmd.pb.go#L721

suleymanakbas91:
> Please update the comments on this line. Because, in the case of thin pools, overprovisioned bytes are set.
> 
> https://github.com/topolvm/topolvm/blob/b3602ddc5b3643b2370df40f124ba57c75bff2a8/lvmd/proto/lvmd.pb.go#L721

Thanks for the review @cupnes! I've updated the comment and re-generated the files.

toshipp (approved):

cupnes (approved):
Thank you. LGTM

PR #652, opened 2023-02-08
try to update go 1.19 to fix ci

cupnes:
Signed-off-by: Yuma Ogami <yuma-ogami@cybozu.co.jp>
Co-authored-by: ESASHIKA Kaoru <kaoru-esashika@cybozu.co.jp>

### DIFF
--- a/api/legacy/v1/groupversion_info.go
+++ b/api/legacy/v1/groupversion_info.go
@@ -1,6 +1,6 @@
 // Package v1 contains API Schema definitions for the topolvm v1 API group
-//+kubebuilder:object:generate=true
-//+groupName=topolvm.cybozu.com
+// +kubebuilder:object:generate=true
+// +groupName=topolvm.cybozu.com
 package v1
 
 import (
--- a/api/v1/groupversion_info.go
+++ b/api/v1/groupversion_info.go
@@ -1,6 +1,6 @@
 // Package v1 contains API Schema definitions for the topolvm v1 API group
-//+kubebuilder:object:generate=true
-//+groupName=topolvm.io
+// +kubebuilder:object:generate=true
+// +groupName=topolvm.io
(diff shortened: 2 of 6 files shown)


api/legacy/v1/groupversion_info.go
```diff
@@ -1,6 +1,6 @@
 // Package v1 contains API Schema definitions for the topolvm v1 API group
-//+kubebuilder:object:generate=true
-//+groupName=topolvm.cybozu.com
+// +kubebuilder:object:generate=true
```
llamerada-jp, line 2: These comments are used by kubebuilder. I'm not sure if it will be recognized correctly, so please do not change the spaces.

api/v1/groupversion_info.go
```diff
@@ -1,6 +1,6 @@
 // Package v1 contains API Schema definitions for the topolvm v1 API group
-//+kubebuilder:object:generate=true
-//+groupName=topolvm.io
+// +kubebuilder:object:generate=true
```
llamerada-jp, line 2: ditto

llamerada-jp (changes requested):

api/legacy/v1/groupversion_info.go
pluser: The above code generation results from adjustments made by `go fmt`. It does not seem to affect the behavior of kubebuild even if it contains blanks (ref: https://github.com/golang/go/issues/54349). The difference between the result of `go fmt` and that of kubebuilder will be solved by fixing [kubebuilder](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v4/scaffolds/internal/templates/api/group.go#L57).

llamerada-jp (approved):

Issue #650, opened 2023-02-06
Add metric for number of PVs in volume group


Thor77:
**What should the feature do:**
Add a metric for number of PVs in volume group managed by lvmd

**What is use case behind this feature:**
Check whether all disks of a server are in use by comparing it with count of drives visible to OS or HBA controller

I'm open to implementing this myself if you would accept a PR for this. This would probably require changes to the lvmd protocol as it currently doesn't propagate that information (which should already be available in the output of the lvm commands run for finding information like vg size).

daichimukai:
"PV" stands for "Physical Disk" of LVM, right?

As a CSI plugin, TopoLVM does not need to be aware of the PVs that make up the VG. So I am not sure if this is the preferred method.

Thor77:
@daichimukai PV = physical volume in LVM. It doesn't need to be aware of them or do anything with that information, it would just be nice if that information could be exposed as a metric so it can be used for monitoring purposes.

daichimukai:
@Thor77 
I think such information should be exported by a dedicated exporter such as [prometheus-lvm-exporter](https://github.com/hansmi/prometheus-lvm-exporter), or node-exporter with [the textfile collector for LVM](
https://github.com/prometheus-community/node-exporter-textfile-collector-scripts/blob/6f710f88bb38625657f61aa42d5f61fbdc2b8066/lvm-prom-collector).

Thor77:
@daichimukai I'll look into those options as well, thanks. As I already implemented the patch to see how much effort it would be, I still opened a PR for it but can also understand if it's not accepted.

PR #654, opened 2023-02-09
Add physical volume count metric

Thor77:
Use case described in #650, can be used to verify the volume group was setup correctly.

I can also understand if this is out of scope as the information is purely used for monitoring and not really relevant for any scheduling decision etc.


daichimukai:
Thank you for your contribution.

I've discussed this PR with other maintainers. We agreed that the lvmd is more preferable than the topolvm-node if we export PV count metrics. Could you change your patch to do so?

PR #662, opened 2023-02-20
Export csi sidecar metrics

cupnes:
Signed-off-by: Yuma Ogami <yuma-ogami@cybozu.co.jp>

### DIFF
--- a/charts/topolvm/templates/controller/deployment.yaml
+++ b/charts/topolvm/templates/controller/deployment.yaml
@@ -106,10 +106,14 @@ spec:
             - --feature-gates=Topology=true
             - --leader-election
             - --leader-election-namespace={{ .Release.Namespace }}
+            - --http-endpoint=:9809
             {{- with .Values.controller.storageCapacityTracking.enabled }}
             - --enable-capacity
             - --capacity-ownerref-level=2
             {{- end }}
+          ports:
+            - containerPort: 9809
+              name: csi-provisioner
           {{- with .Values.controller.storageCapacityTracking.enabled }}
           env:
             - name: NAMESPACE
@@ -142,6 +146,10 @@ spec:
             - --csi-address=/run/topolvm/csi-topolvm.sock
             - --leader-election
             - --leader-election-namespace={{ .Release.Namespace }}
+            - --http-endpoint=:9810
+          ports:
+            - containerPort: 9810
+              name: csi-resizer
           volumeMounts:
             - name: socket-dir
               mountPath: /run/topolvm
@@ -164,6 +172,10 @@ spec:
             - --csi-address=/run/topolvm/csi-topolvm.sock
             - --leader-election
             - --leader-election-namespace={{ .Release.Namespace }}
+            - --http-endpoint=:9811
+          ports:
+            - containerPort: 9811
+              name: csi-snapshotter
           volumeMounts:
             - name: socket-dir
               mountPath: /run/topolvm
@@ -181,6 +193,10 @@ spec:
           command:
             - /livenessprobe
             - --csi-address=/run/topolvm/csi-topolvm.sock
+            - --http-endpoint=:9808
+          ports:
+            - containerPort: 9808
+              name: livenessprobe
           volumeMounts:
             - name: socket-dir
               mountPath: /run/topolvm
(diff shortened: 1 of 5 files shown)


charts/topolvm/templates/controller/deployment.yaml
```diff
@@ -164,6 +166,7 @@ spec:
             - --csi-address=/run/topolvm/csi-topolvm.sock
             - --leader-election
             - --leader-election-namespace={{ .Release.Namespace }}
+            - --http-endpoint=:9810
```
peng225, line 169: Could you exchange the port number with `csi-resizer`'s one (9811)?
Currently, the port number appears in the order of 9809, 9811, 9810, 9812.
I would like it to be 9809, **9810, 9811**, 9812.

charts/topolvm/templates/controller/deployment.yaml
```diff
@@ -181,6 +184,7 @@ spec:
           command:
             - /livenessprobe
             - --csi-address=/run/topolvm/csi-topolvm.sock
+            - --metrics-address=:9812
```
peng225, line 187: `metrics-address` is deprecated.
https://github.com/kubernetes-csi/livenessprobe/blob/c2c5608c8a8907efeff276b613a035df5343b858/cmd/livenessprobe/main.go#L46

Please use `http-endpoint` instead.
https://github.com/kubernetes-csi/livenessprobe#other-recognized-arguments

charts/topolvm/templates/node/daemonset.yaml
```diff
@@ -137,6 +137,7 @@ spec:
           command:
             - /livenessprobe
             - --csi-address={{ .Values.node.kubeletWorkDirectory }}/plugins/{{ include "topolvm.pluginName" . }}/node/csi-topolvm.sock
+            - --metrics-address=:9812
```
peng225, line 140: Please use `http-endpoint`.
(See also the comment above.)

peng225 (commented):
If possible, I want tests to check whether these metrics are exposed correctly.
What do you think?

toshipp (commented):
Would you define port names?
Some monitoring systems use port names to scrape metrics.
VictoriaMetrics is a exmple. https://docs.victoriametrics.com/operator/api.html#podmetricsendpoint

cupnes:
@peng225 
> If possible, I want tests to check whether these metrics are exposed correctly.
> What do you think?

Thank you. I think so too. I consider how it can be implemented.

e2e/controller_test.go
```diff
@@ -0,0 +1,72 @@
...
+	var cc CleanupContext
+	BeforeEach(func() {
+		cc = commonBeforeEach()
+	})
+	AfterEach(func() {
+		commonAfterEach(cc)
+	})
+
+	It("should be deployed", func() {
+		Eventually(func() error {
```
peng225, line 23: Can't you replace this `Eventually` section with `kubectl wait` command?

For example:
```
kubectl("get", "-n=topolvm-system", "deploy", "--selector=app.kubernetes.io/component=controller,app.kubernetes.io/name=topolvm", "--for=condition=Available", "--timeout=120s"
```
(Please change the timeout value to something good.)

peng225: I think that is correct.

cupnes: I fixed it. Please review.

peng225 (approved):

toshipp (changes requested):
We use unique port names in a pod.
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#containerport-v1-core

cupnes:
@toshipp Thank you. I forgot it. I fixed.

peng225 (dismissed):

e2e/controller_test.go
```diff
@@ -0,0 +1,36 @@
...
+		var pods corev1.PodList
+		err = json.Unmarshal(stdout, &pods)
+		Expect(err).ShouldNot(HaveOccurred())
+
+		pod := pods.Items[0]
+
+		// For CSI sidecar metrics access checking
+		for _, port := range []string{"9808", "9809", "9810", "9811"} {
+			stdout, stderr, err = kubectl("exec", "-n", "topolvm-system", pod.Name, "-c=topolvm-controller", "--", "curl", "http://localhost:"+port+"/metrics")
+			Expect(err).ShouldNot(HaveOccurred(), "stdout=%s, stderr=%s", stdout, stderr)
```
toshipp, line 33: This can be simplified as `kubectl exec -n topolvm-system deploy/topolvm-controller -c topolvm-controller -- curl http://...`.
Could you revise it?

toshipp (changes requested):

e2e/controller_test.go
cupnes: Thank you. I fixed it.

peng225 (approved):

Issue #657, opened 2023-02-17
Fix the `use_legacy` test to be done somewhere other than example


cupnes:
## What

The `use_legacy` is tested in example[*1]. However, in essence, it should not.
[*1] https://github.com/topolvm/topolvm/blob/2492a48255a113e8534b1140cf81ea55192b035d/.github/workflows/main.yaml#L59

## How

Fix the `use_legacy` test to be done somewhere other than example, such as e2e.

## Checklist

- [x] Finish implementation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

PR #665, opened 2023-02-28
move legacy mode test to e2e test

peng225:
issue: topolvm/topolvm#657


.github/workflows/e2e-k8s.yaml
```diff
@@ -30,3 +30,7 @@ jobs:
     with:
       thin_csi_sanity: "true"
 
+  use-legacy:
+    uses: ./.github/workflows/e2e-k8s-latest-only-workflow.yaml
```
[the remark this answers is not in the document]
daichimukai: https://github.com/topolvm/topolvm/blob/0239c1fb6a7631a0673f35f422ff542cc395a298/constants.go#L15-L17
OMG. This means that TopoLVM works in the legacy mode by passing environment variable `USE_LEGACY=false`. I've expected that TopoLVM works in the legacy mode _if and only if_ `USE_LEGACY=true`.

e2e/testdata/scheduler/capacity-pod-template.yaml
```diff
@@ -6,13 +6,13 @@ metadata:
...
-    capacity.topolvm.io/ssd: %q
+    %sssd: %q
 spec:
   containers:
   - name: pause
     image: registry.k8s.io/pause
     resources:
       requests:
-        topolvm.io/capacity: 1
+        %s/capacity: 1
```
daichimukai, line 16:
```suggestion
        %s: 1
```

Please use `GetCapacityResource()` in constants.go.

e2e/scheduler_test.go
```diff
@@ -73,7 +74,8 @@ func testScheduler() {
 
 		ns := testNamespacePrefix + randomString(10)
 		createNamespace(ns)
-		stdout, stderr, err := kubectlWithInput([]byte(fmt.Sprintf(capacityPodTemplateYAML, ns, "1073741824")), "apply", "-f", "-")
+		stdout, stderr, err := kubectlWithInput([]byte(fmt.Sprintf(capacityPodTemplateYAML, ns, topolvm.GetCapacityKeyPrefix(), "1073741824", topolvm.GetPluginName(), topolvm.GetPluginName())),
```
daichimukai, line 77:
```suggestion
		stdout, stderr, err := kubectlWithInput([]byte(fmt.Sprintf(capacityPodTemplateYAML, ns, topolvm.GetCapacityKeyPrefix(), "1073741824", topolvm.GetCapacityResource(), topolvm.GetCapacityResource())),
```

In addition, the corresponding format specifier should be replaced from `%s/capacity` to `%s`.

daichimukai: nits: this line seems too long to read. It may be nice to split.

e2e/scheduler_test.go
```diff
@@ -107,7 +109,8 @@ func testScheduler() {
 
 		ns := testNamespacePrefix + randomString(10)
 		createNamespace(ns)
-		stdout, stderr, err := kubectlWithInput([]byte(fmt.Sprintf(capacityPodTemplateYAML, ns, "21474836480")), "apply", "-f", "-")
+		stdout, stderr, err := kubectlWithInput([]byte(fmt.Sprintf(capacityPodTemplateYAML, ns, topolvm.GetCapacityKeyPrefix(), "21474836480", topolvm.GetPluginName(), topolvm.GetPluginName())),
```
daichimukai, line 112: ditto.

e2e/testdata/scheduler/capacity-pod-template.yaml
```diff
@@ -6,13 +6,13 @@ metadata:
...
       limits:
-        topolvm.io/capacity: 1
+        %s/capacity: 1
```
daichimukai, line 18: ditto.

daichimukai (changes requested):

e2e/Makefile
```diff
@@ -5,6 +5,7 @@ STORAGE_CAPACITY ?= false
 SKIP_NODE_FINALIZE ?= false
 SANITY_TEST_WITH_THIN_DEVICECLASS ?= false
 EXTERNAL_SNAPSHOTTER_VERSION ?= v6.0.1
+USE_LEGACY ?= ""
```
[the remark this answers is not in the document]
daichimukai: https://github.com/topolvm/topolvm/blob/0239c1fb6a7631a0673f35f422ff542cc395a298/constants.go#L15-L17
OMG. This means that TopoLVM works in the legacy mode by passing environment variable `USE_LEGACY=false`. I've expected that TopoLVM works in the legacy mode _if and only if_ `USE_LEGACY=true`.

e2e/Makefile
```diff
@@ -193,7 +201,11 @@ test: create-cluster
 	$(KUBECTL) apply -f https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/$(EXTERNAL_SNAPSHOTTER_VERSION)/client/config/crd/snapshot.storage.k8s.io_volumesnapshots.yaml
 	$(KUBECTL) apply -f https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/$(EXTERNAL_SNAPSHOTTER_VERSION)/deploy/kubernetes/snapshot-controller/rbac-snapshot-controller.yaml
 	$(KUBECTL) apply -f https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/$(EXTERNAL_SNAPSHOTTER_VERSION)/deploy/kubernetes/snapshot-controller/setup-snapshot-controller.yaml
-	$(KUBECTL) apply -f manifests/common/
+	if [ "$(USE_LEGACY)" = "true" ]; then \
+		$(KUBECTL) apply -f manifests/provisioner_legacy.yaml; \
+	else \
+		$(KUBECTL) apply -f manifests/provisioner.yaml; \
+	fi
```
daichimukai, line 208:
```suggestion
ifeq ($(USE_LEGACY),true)
	$(KUBECTL) apply -f manifests/provisioner_legacy.yaml
else
	$(KUBECTL) apply -f manifests/provisioner.yaml
endif
```

Or it may be nice to define a variable and switch its value like `SCHEDULER_CONFIG` and `HELM_VALUES_FILE`.

.github/workflows/e2e-k8s.yaml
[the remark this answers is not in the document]
peng225: Just beause it is time consuming.

e2e/Makefile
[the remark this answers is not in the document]
peng225: I agree that this is odd.
Do you want to fix it in this PR?

e2e/hook_test.go
```diff
@@ -112,7 +112,7 @@ func testHook() {
 		stdout, stderr, err := kubectlWithInput(hookGenericEphemeralVolumeYAML, "-n", nsHookTest, "apply", "-f", "-")
 		Expect(err).ShouldNot(HaveOccurred(), "stdout=%s, stderr=%s", stdout, stderr)
 
-		By("checking pod is annotated with topolvm.io/capacity")
+		By("checking pod is annotated with" + topolvm.GetPluginName() + "/capacity")
```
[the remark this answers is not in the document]
peng225: Because the return type of `GetCapacityResource()` is `corev1.ResourceName`, not string, we need a type conversion like:
```
By("checking pod is annotated with" + string(topolvm.GetCapacityResource()))
```

Is that you want? Or should I define another function to return string type value?

e2e/testdata/scheduler/capacity-pod-template.yaml
```diff
@@ -6,13 +6,13 @@ metadata:
   labels:
     app.kubernetes.io/name: pause
   annotations:
-    capacity.topolvm.io/ssd: %q
+    %sssd: %q
```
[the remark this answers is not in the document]
peng225: Other similar functions do not append `/`, so I will remove the trailing `/` in the definition of `GetCapacityKeyPrefix()`.

e2e/testdata/scheduler/capacity-pod-template.yaml
peng225: Same as https://github.com/topolvm/topolvm/pull/665#discussion_r1123908177

e2e/scheduler_test.go
peng225: Same as https://github.com/topolvm/topolvm/pull/665#discussion_r1123908177

> nits: this line seems too long to read. It may be nice to split.

Thanks. I will split it afther https://github.com/topolvm/topolvm/pull/665#discussion_r1123908177 is resolved.

e2e/scheduler_test.go
peng225: Same as https://github.com/topolvm/topolvm/pull/665#discussion_r1123908177

e2e/testdata/scheduler/capacity-pod-template.yaml
peng225: Same as https://github.com/topolvm/topolvm/pull/665#discussion_r1123908177

e2e/Makefile
peng225: > Or it may be nice to define a variable and switch its value like SCHEDULER_CONFIG and HELM_VALUES_FILE.

I fixed it in this way. Thanks.

e2e/testdata/scheduler/capacity-pod-template.yaml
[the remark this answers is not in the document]
daichimukai: Leave it as is for now.

e2e/hook_test.go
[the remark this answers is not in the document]
daichimukai: core/v1 implements a dedicated method that returns its name in string.
https://pkg.go.dev/k8s.io/api/core/v1#ResourceName.String

e2e/Makefile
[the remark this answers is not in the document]
daichimukai: No, leave it as is for now.

.github/workflows/e2e-k8s.yaml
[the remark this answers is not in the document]
toshipp: I'm considering moving the test for node finalizer from e2e to the middle test. This will reduce the number of jobs. After that, we can test the feature for all k8s versions.

e2e/testdata/scheduler/capacity-pod-template.yaml
peng225: I fixed it.

e2e/scheduler_test.go
peng225: Fixed.

e2e/scheduler_test.go
peng225: Fixed.

e2e/testdata/scheduler/capacity-pod-template.yaml
peng225: Fixed.

e2e/hook_test.go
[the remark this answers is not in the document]
peng225: Fixed.

.github/workflows/e2e-k8s.yaml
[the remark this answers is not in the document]
daichimukai: Then, I am not very happy about the large portion of `.github/workflows/e2e-k8s-latest-only-workflow.yaml` and `.github/workflows/e2e-k8s.yam`l that is duplicated.

Would it be possible to eliminate the duplication by, for example
1. change `e2e-k8s-latest-only-workflow.yaml` to a workflow that runs the e2e test for one k8s version received in input, instead of the latest k8s version,
2. in the e2e test for `USE_LEGACY=true`, call the workflow of 1. directly,
3. in `.github/workflows/e2e-k8s.yaml`, call the workflow of 1. with matrix strategy.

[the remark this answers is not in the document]
peng225: Because the test for node finalizer mentioned in https://github.com/topolvm/topolvm/pull/665#discussion_r1124031783 is now going to be removed from e2e test by https://github.com/topolvm/topolvm/pull/670, I changed my mind and let legacy mode be tested with three Kubernetes versions.
Sorry for bothering you.

e2e/sanity_test.go
```diff
@@ -74,6 +75,7 @@ func testSanity() {
 	}
 
 	if os.Getenv("SANITY_TEST_WITH_THIN_DEVICECLASS") == "true" {
+		// This test does not run in the legacy mode.
```
daichimukai, line 78: I'm not sure but is this test _need not_ to pass in the legacy mode?
If you mean that we do not run this test with the legacy mode in CI, it's just a limitation due to the bandwidth of CI runners. I think we should be able to run it in our local environment if we'd like to do so.

daichimukai (approved):
LGTM

pluser (approved):

PR #666, opened 2023-02-28
Fill Status.CurrentSize

toshipp:
The field is added in v0.4.0, and topolvm does not fill it always. However, we assume it exists, so fill it now.

### DIFF
--- a/controllers/logicalvolume_controller.go
+++ b/controllers/logicalvolume_controller.go
@@ -194,6 +194,7 @@ func (r *LogicalVolumeReconciler) createLV(ctx context.Context, log logr.Logger,
 		}
 		if found {
 			log.Info("set volumeID to existing LogicalVolume", "name", lv.Name, "uid", lv.UID, "status.volumeID", lv.Status.VolumeID)
+			// Don't set CurrentSize here because the Spec.Size field may be updated after the LVM LV is created.
 			lv.Status.VolumeID = string(lv.UID)
 			lv.Status.Code = codes.OK
 			lv.Status.Message = ""
@@ -274,18 +275,19 @@ func (r *LogicalVolumeReconciler) createLV(ctx context.Context, log logr.Logger,
 }
 
 func (r *LogicalVolumeReconciler) expandLV(ctx context.Context, log logr.Logger, lv *topolvmv1.LogicalVolume) error {
-	// lv.Status.CurrentSize is added in v0.4.0 and filled by topolvm-controller when resizing is triggered.
-	// The reconciliation loop of LogicalVolume may call expandLV before resizing is triggered.
-	// So, lv.Status.CurrentSize could be nil here.
-	if lv.Status.CurrentSize == nil {
+	// We denote unknown size as -1.
+	var origBytes int64 = -1
+	switch {
+	case lv.Status.CurrentSize == nil:
+		// topolvm-node may be crashed before setting Status.CurrentSize.
+		// Since the actual volume size is unknown,
+		// we need to do resizing to set Status.CurrentSize to the same value as Spec.Size.
+	case lv.Spec.Size.Cmp(*lv.Status.CurrentSize) <= 0:
 		return nil
+	default:
+		origBytes = (*lv.Status.CurrentSize).Value()
 	}
 
-	if lv.Spec.Size.Cmp(*lv.Status.CurrentSize) <= 0 {
-		return nil
-	}
-
-	origBytes := (*lv.Status.CurrentSize).Value()
 	reqBytes := lv.Spec.Size.Value()
 
 	err := func() error {
--- a/e2e/logical_volume_test.go
+++ b/e2e/logical_volume_test.go
@@ -1,0 +1,190 @@
+package e2e
+
+import (
+	"context"
+	_ "embed"
+	"encoding/json"
+	"fmt"
+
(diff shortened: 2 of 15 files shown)


controllers/logicalvolume_controller.go
```diff
@@ -273,10 +274,15 @@ func (r *LogicalVolumeReconciler) createLV(ctx context.Context, log logr.Logger,
...
 
 func (r *LogicalVolumeReconciler) expandLV(ctx context.Context, log logr.Logger, lv *topolvmv1.LogicalVolume) error {
-	// lv.Status.CurrentSize is added in v0.4.0 and filled by topolvm-controller when resizing is triggered.
-	// The reconciliation loop of LogicalVolume may call expandLV before resizing is triggered.
-	// So, lv.Status.CurrentSize could be nil here.
+	// WA: since Status.CurrentSize is added in v0.4.0, it may be missing.
+	// To ensure the field existed, update it here.
+	reqBytes := lv.Spec.Size.Value()
 	if lv.Status.CurrentSize == nil {
+		lv.Status.CurrentSize = resource.NewQuantity(reqBytes, resource.BinarySI)
```
peng225, line 281: I think you cannot use `reqBytes` here.

The value of `reqBytes` comes from `lv.Spec.Size.Value()`. This is set to the expanded capacity at [(s *LogicalVolumeService) ExpandVolume](https://github.com/topolvm/topolvm/blob/6b74e7f1d02ad141ae0aaa683bb804f7bc5785f5/driver/internal/k8s/logicalvolume_service.go#L310) before `expandLV` is called.

If we assign `reqBytes` to `lv.Status.CurrentSize` here, the status is updated in the subsequent line despite that the actual LV size has not been expanded.

Then, in the next reconciliation which is incurred by the status update, the following condition holds true, and the volume will never be expanded.
https://github.com/topolvm/topolvm/blob/6b74e7f1d02ad141ae0aaa683bb804f7bc5785f5/controllers/logicalvolume_controller.go#L289

IMO, when `lv.Status.CurrentSize` is nil, its value should be set **before the spec is updated**. Though you deleted the following lines in this PR, I guess they are necessary.
https://github.com/topolvm/topolvm/blob/af937939f2f984da31bc6977c9909b54a8579647/driver/controller.go#L593-L597

toshipp: Good catch! It is true that status fields should be updated after reconciliation succeeds for retrying. Your case that the status is updated before the spec is updated means reconciliation succeeds and the status update happens after the reconciliation. Therefore, we can move the status update from topolvm-controller to topolvm-node.

To move the update after resizing is completed, we ensure lvmd can accept expanding a volume to the original size. Happily, it is already done. So I'll just update the code to continue the resize logic same as actual resizing if CurrentSize is nil. Also, I'll add test cases for this.

e2e/logical_volume_test.go
```diff
@@ -0,0 +1,190 @@
...
+	It("should set Status.CurrentSize", func() {
+		pvcName := "check-current-size"
+		pvcYaml := fmt.Sprintf(pvcTemplateYAMLForLV, pvcName)
+		stdout, stderr, err := kubectlWithInput([]byte(pvcYaml), "apply", "-n", nsLogicalVolumeTest, "-f", "-")
+		Expect(err).ShouldNot(HaveOccurred(), "stdout=%s, stderr=%s", stdout, stderr)
+
+		By("checking that Status.CurrentSize exists")
+		var pvc corev1.PersistentVolumeClaim
+		Eventually(func(g Gomega) {
+			g.Expect(getObject(&pvc, "get", "pvc", "-n", nsLogicalVolumeTest, pvcName, "-ojson")).Should(Succeed())
```
peng225, line 55: When we want to use `getObject`, we always want to do `kubectl get`.
Specifying "get" every time we use it seems to be redundant.

I would be happy if I could use it as follows (without "get").
```
getObject(&pvc, "pvc", "-n", nsLogicalVolumeTest, pvcName, "-ojson")
```

controllers/logicalvolume_controller.go
```diff
@@ -274,18 +275,18 @@ func (r *LogicalVolumeReconciler) createLV(ctx context.Context, log logr.Logger,
...
-	// lv.Status.CurrentSize is added in v0.4.0 and filled by topolvm-controller when resizing is triggered.
-	// The reconciliation loop of LogicalVolume may call expandLV before resizing is triggered.
-	// So, lv.Status.CurrentSize could be nil here.
-	if lv.Status.CurrentSize == nil {
+	// We denote unknown size as -1.
+	var origBytes int64 = -1
+	switch {
+	case lv.Status.CurrentSize == nil:
+		// topolvm-node may be crashed before setting Status.CurrentSize.
+		// The actual size is unknown, we need to do resizing to set Spec.Size to Status.CurrentSize.
```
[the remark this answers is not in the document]
toshipp: Yes. I missed it. Thanks.

e2e/logical_volume_test.go
toshipp: Nice! I'll fix it.

toshipp: I also noticed that it needs `-ojson` always. I fixed it too.

peng225 (approved):

pluser (approved):

PR #669, opened 2023-03-06
Use csi go module instead of generating sources from proto

### LINKED ISSUE #634: Resolve the warnings for protobuf when building the program.
## What

Resolve the warnings like the ones below when building the program.
`WARNING: proto: message csi.v1.GetPluginInfoRequest is already registered`
Those messages are because of using both upstream csi.pb.go (by csi-sanity) and generating myself.

## How

- use upstream [csi.pb.go](https://github.com/container-storage-interface/spec/blob/master/lib/go/csi/csi.pb.go) instead of generating it myself

## Checklist

- [ ] Test all functions

toshipp:
Previously, we registered CSI gRPC twice because csi-sanity depends on CSI go module and we generate the module ourselves from proto. This causes warning messages like "WARNING: proto: message
csi.v1.GetPluginInfoRequest is already registered" To avoid such messages, use the Go module version instead.

Fixes #634.


docs/maintenance.md
```diff
@@ -137,10 +137,4 @@ Basically, CSI spec should NOT be upgraded aggressively.
...
 Upgrade the CSI version only if new features we should cover are introduced in newer versions, or the Kubernetes versions TopoLVM is going to support does not support the current CSI version.
 
-For updating CSI spec, we need to update `CSI_VERSION` in `Makefile`, and run the following commands.
-
-```bash
-$ rm csi.proto
-$ make csi.proto
-$ make generate
-```
+For updating CSI spec, update `go.mod`.
```
peng225, line 140: Do you mean `github.com/container-storage-interface/spec` in `go.mod`?
If so, I would be happy if you could specify it concretely.

```suggestion
For updating CSI spec, update the version of `github.com/container-storage-interface/spec` in `go.mod`.
```

peng225 (approved):

satoru-takeuchi (approved):

PR #671, opened 2023-03-07
Refactor: use controllerutil to operate finalizers

toshipp:
We operated finalizers with hand written logic, but controllerutil provides such that.

### DIFF
--- a/controllers/logicalvolume_controller.go
+++ b/controllers/logicalvolume_controller.go
@@ -17,6 +17,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	crlog "sigs.k8s.io/controller-runtime/pkg/log"
 )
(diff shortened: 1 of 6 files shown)


satoru-takeuchi (approved):

controllers/logicalvolume_controller.go
```diff
@@ -60,9 +61,9 @@ func (r *LogicalVolumeReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	}
 
 	if lv.ObjectMeta.DeletionTimestamp == nil {
-		if !containsString(lv.Finalizers, topolvm.GetLogicalVolumeFinalizer()) {
+		if !controllerutil.ContainsFinalizer(lv, topolvm.GetLogicalVolumeFinalizer()) {
```
llamerada-jp, line 64: `AddFinalizer` function checks the duplication of finalizers. And You can know from the return value if the finalizer already existed or not. So you can write it more simpler.
https://github.com/kubernetes-sigs/controller-runtime/blob/c304e7ec2ee7879f53ce69c12a7aa133853fccdb/pkg/controller/controllerutil/controllerutil.go#L353

controllers/node_controller.go
```diff
@@ -156,24 +142,9 @@ func (r *NodeReconciler) doFinalize(ctx context.Context, log logr.Logger, node *
...
-	finExists := false
-	for _, fin := range lv.Finalizers {
-		if fin == topolvm.GetLogicalVolumeFinalizer() {
-			finExists = true
-			break
-		}
-	}
-
-	if finExists {
+	if controllerutil.ContainsFinalizer(lv, topolvm.GetLogicalVolumeFinalizer()) {
```
llamerada-jp, line 145: `RemoveFinalizer` tells us whether the finalizer existed or not.
https://github.com/kubernetes-sigs/controller-runtime/blob/c304e7ec2ee7879f53ce69c12a7aa133853fccdb/pkg/controller/controllerutil/controllerutil.go#L366

controllers/persistentvolumeclaim_controller.go
```diff
@@ -148,17 +148,9 @@ func (r *PersistentVolumeClaimReconciler) SetupWithManager(mgr ctrl.Manager) err
...
 	// So we need to delete all of them.
-	for i := 0; i < len(pvc.Finalizers); {
-		if pvc.Finalizers[i] == topolvm.LegacyPVCFinalizer {
-			pvc.Finalizers = append(pvc.Finalizers[:i], pvc.Finalizers[i+1:]...)
-			removed = true
-			continue
-		}
-		i++
-	}
+	removed := controllerutil.RemoveFinalizer(pvc, topolvm.LegacyPVCFinalizer)
```
llamerada-jp, line 153: ditto

llamerada-jp (changes requested):

controllers/logicalvolume_controller.go
toshipp: I considered the option but did not choose it due to performance. 
`AddFinaizer` or `RemoveFinalizer` overwrites its argument, so we need to `DeepCopy` to create a patch before calling it. `DeepCopy` traverses whole object fields, but `AddFinaizer` or `RemoveFinalizer` does only finalizers, the cost is lower than `DeepCopy`.

llamerada-jp (approved):

PR #674, opened 2023-03-13
driver: use os.Stat() instead of filesystem.Stat()

daichimukai:
filesystem.Stat() is a wrapper of unix.Stat() that retries if a syscall returns -EINTR. os.Stat() in the standard library does this retry too.  This patch changes to use os.Stat() instead of filesystem.Stat() and removes the latter function.

---

By any chance there may be some reason to use unix.Stat() with retry instead of os.Stat(), but I cannot find out that. If you know something, please let me know.

### DIFF
--- a/driver/node.go
+++ b/driver/node.go
@@ -2,12 +2,15 @@ package driver
 
 import (
 	"context"
+	"errors"
 	"io"
+	"io/fs"
 	"os"
 	"path"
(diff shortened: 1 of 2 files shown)


daichimukai:
I completed the PR but don't think it as good patch. I no longer wanted to merge it.

PR #678, opened 2023-03-16
Refactor e2e

toshipp:
See commit messages.

### DIFF
--- a/e2e/helper_test.go
+++ b/e2e/helper_test.go
@@ -46,15 +46,6 @@ func kubectlWithInput(input []byte, args ...string) (stdout []byte, stderr []byt
 	return execAtLocal(kubectlPath, input, args...)
 }
 
-func containString(s []string, target string) bool {
-	for _, ss := range s {
-		if ss == target {
-			return true
-		}
-	}
-	return false
-}
-
 func countLVMs() (int, error) {
 	stdout, err := exec.Command("sudo", "lvs", "-o", "lv_name", "--noheadings").Output()
 	if err != nil {
--- a/e2e/helper_test.go
+++ b/e2e/helper_test.go
@@ -24,40 +24,53 @@ type CleanupContext struct {
 	CapacityAnnotations map[string]map[string]string
 }
 
-func execAtLocal(cmd string, input []byte, args ...string) ([]byte, []byte, error) {
-	var stdout, stderr bytes.Buffer
+// execAtLocal executes cmd.
+// If the cmd is failed, stdout and stderr is included in the error.
+// Therefore you don't need to print the returned stdout and stderr,
+// just print err itself.
+func execAtLocal(cmd string, input []byte, args ...string) (stdout []byte, stderr []byte, err error) {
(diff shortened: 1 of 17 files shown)


toshipp:
waiting for #666

e2e/helper_test.go
```diff
@@ -24,49 +24,52 @@ type CleanupContext struct {
 	CapacityAnnotations map[string]map[string]string
 }
 
-func execAtLocal(cmd string, input []byte, args ...string) ([]byte, []byte, error) {
-	var stdout, stderr bytes.Buffer
+// execAtLocal executes cmd.
+// If the cmd is failed, stdout and stderr is included in the error.
+// Therefore it is unnecessary to report them on caller.
```
peng225, line 29:
```suggestion
// Therefore it is unnecessary to report them to the caller.
```

e2e/helper_test.go
```diff
@@ -24,49 +24,52 @@ type CleanupContext struct {
...
+func execAtLocal(cmd string, input []byte, args ...string) (stdout []byte, stderr []byte, err error) {
```
peng225, line 30: `stderr` is not used in 100% of cases, if my understanding is correct. I doubt we still need it.
When we find a case where we really want to handle stderr, then we can implement the function for that purpose.

(On the other hand, I imagine that it is super hard and boring to fix it... So this is just a comment. I will leave it to you.)

toshipp: My comment seems not good. 
I mean if this function returns an error, the error contains stderr, so you don't need to print the 2nd return value (stderr) explicitly, just print the error to show stderr instead.

Since this function still returns stdout and stderr even if an error happens, I assume the existing code never breaks.

I'll update the comment to express my intent more.

peng225: Thank you for the clarification.

What I wanted to say is that we would be able to omit stderr from the returned values. Though it breaks the callers' code, no one is using the returned stderr because the callers now do not need to print it even when the error occurs.

If the signature looks like:
```Go
func execAtLocal(cmd string, input []byte, args ...string) (stdout []byte, err error)
```

then we can use it as follows
```Go
stdout, err := execAtLocal(foo, bar)
```

instead of
```Go
stdout, _, err := execAtLocal(foo, bar)
```

toshipp: Unfortunately, it can not. we use it.
https://github.com/topolvm/topolvm/pull/678/files#diff-e91d29c6d4c58503d9c2d72e2b04f1b70901a613e23008d1d1cfe8beff7038d4R214-R216

peng225 (approved):

e2e/thin_provisioning_test.go
```diff
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"os/exec"
```
daichimukai, line 8: I've struggled with the mystery of number of diff lines (+547 −548) and finally found this removal. Yay!

daichimukai (approved):

PR #683, opened 2023-03-22
Helm Chart - Support imagePullSecrets

dp-42:
Hi,

I noticed that the TopoLVM Helm Chart does not support configuration of `imagePullSecrets` although this seems to be generally considered as good practice - especially when you use on-prem K8s and your own image registry.  Therefore I implemented the support of this setting.

In addition, I found a bug 🐛:
Although you can set the `imagePullPolicy` globally in the `values.yaml` it is not applied to the `lvmd` DaemonSet. I fixed that as well.

Also feel free to cherry pick or let me know if I should adapt anything! 😁

### DIFF
--- a/charts/topolvm/values.yaml
+++ b/charts/topolvm/values.yaml
@@ -7,10 +7,13 @@ image:
 
   # image.tag -- TopoLVM image tag to use.
   # @default -- `{{ .Chart.AppVersion }}`
-  tag:  # 0.8.3
+  tag:  # 0.18.1
 
   # image.pullPolicy -- TopoLVM image pullPolicy.
   pullPolicy:  # Always
+
+  # image.pullSecrets -- List of imagePullSecrets.
+  pullSecrets: []
 
   csi:
     # image.csi.nodeDriverRegistrar -- Specify csi-node-driver-registrar: image.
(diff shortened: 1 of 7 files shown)


charts/topolvm/values.yaml
```diff
@@ -7,10 +7,15 @@ image:
 
   # image.tag -- TopoLVM image tag to use.
   # @default -- `{{ .Chart.AppVersion }}`
-  tag:  # 0.8.3
+  tag:
```
daichimukai, line 10: Please keep a comment.

charts/topolvm/values.yaml
```diff
@@ -7,10 +7,15 @@ image:
...
 
   # image.pullPolicy -- TopoLVM image pullPolicy.
-  pullPolicy:  # Always
+  pullPolicy:
```
daichimukai, line 13: Please keep a comment.

charts/topolvm/values.yaml
```diff
@@ -7,10 +7,15 @@ image:
...
+
+  # image.pullSecrets -- List of imagePullSecrets.
+  pullSecrets: []
+  # - pull-secret-1
+  # - pull-secret-2
```
daichimukai, line 18:
```suggestion
```

I don't think this comment is very useful

daichimukai (changes requested):

charts/topolvm/values.yaml
dp-42: I have removed those comments

charts/topolvm/values.yaml
dp-42: I have added this one again

daichimukai:
@dp-42 
Thank you. The changes looks good to me. Please squash your commits into one commit.

daichimukai:
@dp-42 
> I have squashed my commits to the 2 commits that make sense in my opinion

I agree you. Thanks.

satoru-takeuchi (approved):

PR #689, opened 2023-04-03
Combine capacity test and scheduler test

toshipp:
These tests if scheduling components can schedule pods with PVCs if
nodes have enough capacity. We support two scheduling methods, scheduler
extender and capacity tracking. Regardless of them, the scheduling feature
should work, so these tests can be into one.


e2e/scheduling_test.go
```diff
@@ -58,17 +53,8 @@ func testStorageCapacity() {
 		_, err = kubectlWithInput(lvYaml, "apply", "-f", "-")
 		Expect(err).ShouldNot(HaveOccurred())
 		Eventually(func() error {
-			var pvc corev1.PersistentVolumeClaim
```
toshipp, line 61: If a pod using PVCs is running, PVCs are bound, so this check is redundant.

peng225 (approved):

satoru-takeuchi (approved):

PR #693, opened 2023-04-04
Cleanup publish test

toshipp:
The test contains conformance tests, the validation test that can be done by unit test, and the mount option checking. I removed the first because we used csi-test, and converted the second as a unit test. The last only remains, so I changed the test file name to be a more appropriate one.


cupnes (approved):

driver/node_test.go
```diff
@@ -0,0 +1,16 @@
+package driver
+
+import (
+	"testing"
+
+	"github.com/container-storage-interface/spec/lib/go/csi"
+)
+
+func TestMakeMountOptions(t *testing.T) {
+	_, err := makeMountOptions(true, &csi.VolumeCapability_MountVolume{
```
llamerada-jp, line 10: Could you please add one more test for normal pattern.

toshipp: Do you mean a pattern for specifying read-only false and rw option? I think it is trivial case and if the pattern fails, all e2e will not pass, so there is not worth adding it.

toshipp: ah, the pattern I mentioned is not normal, I doubt the normal pattern is read-only false and no options. If it is your case, I will add it.

llamerada-jp: What I wanted to say is that "rw" is not specified. But, as you pointed out, all e2e will not pass. I withdraw my point.

llamerada-jp (approved):

llamerada-jp (approved):

PR #696, opened 2023-04-14
Update cleanup test to check if topolvm re-create pods of STS if the node is deleted.

toshipp:
topolvm deletes pods and PVCs if the node is deleted. The main perpuse is not cleaning them up because the node is deleted, actual containers or volumes are not working. But if it does not delete them, the remaining pods can not be re-scheduled because they use node-bounded PVCs. The test checks this case.


e2e/node_delete_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+	BeforeEach(func() {
+		// Skip because this test require multiple nodes but there is just one node in daemonset lvmd test environment.
+		skipIfDaemonsetLvmd()
+
+		createNamespace(nsNodeDeleteTest)
+	})
+
+	AfterEach(func() {
+		skipIfDaemonsetLvmd()
```
peng225, line 32: `skipIfDaemonsetLvmd` is called two times, in `BeforeEach` and `AfterEach`.
Is it needed? I guess we can delete one of them.

e2e/node_delete_test.go
```diff
@@ -0,0 +1,137 @@
...
+		err = getObjects(&logicalVolumeList, "logicalvolumes")
+		Expect(err).ShouldNot(HaveOccurred())
+
+		var targetLVs []topolvmv1.LogicalVolume
+		for _, lv := range logicalVolumeList.Items {
+			if lv.Spec.NodeName == targetNode {
+				targetLVs = append(targetLVs, lv)
+			}
+		}
+
```
peng225, line 97: Is it guaranteed that `targetLVs` is not empty?
Then, it would be good to check it after the loop.

e2e/node_delete_test.go
```diff
@@ -0,0 +1,137 @@
...
+		Expect(err).ShouldNot(HaveOccurred())
+
+		By("confirming a pod using a PVC is re-scheduled to another node")
+		Eventually(func() error {
+			var rescheduledPod corev1.Pod
+			err = getObjects(&rescheduledPod, "pod", "-n", nsNodeDeleteTest, targetPod.Name)
+			if err != nil {
+				return fmt.Errorf("can not get target pod: err=%w", err)
+			}
+			if rescheduledPod.Spec.NodeName != "" && rescheduledPod.Spec.NodeName != targetNode {
```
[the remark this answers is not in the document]
toshipp: You are right. I'll check it.

e2e/node_delete_test.go
toshipp: LogicalVolume is used internally, we don't have any user-visible specs. I think e2e should not test implementation detail but test user-visible specs. So, it is unnecessary to check it, IMO.

e2e/node_delete_test.go
toshipp: I think you are right. I'll check it. thanks.

peng225 (approved):

pluser (approved):

Issue #568, opened 2022-09-26
Specify kindest/node image digest


llamerada-jp:
## What

According to the kind's release note,  kindest/node images are built for each released kind version. Therefore, using the old kind and last kindest/node image may not work correctly. Upstream release note suggests specifying exact images including the digest like below.
> New Node images have been built for kind v0.16.0, please use these exact images (IE like kindest/node:v1.25.2@sha256: 9be91e9e9cdf116809841fc77ebdb8845443c4c72fe5218f3ae9eb57fdb4bace including the digest) or build your own as we may need to change the image format again in the future 😅
>
> Images built for this release:
>
> 1.25: kindest/node:v1.25.2@sha256:9be91e9e9cdf116809841fc77ebdb8845443c4c72fe5218f3ae9eb57fdb4bace
1.24: kindest/node:v1.24.6@sha256:97e8d00bc37a7598a0b32d1fabd155a96355c49fa0d4d4790aab0f161bf31be1
1.23: kindest/node:v1.23.12@sha256:9402cf1330bbd3a0d097d2033fa489b2abe40d479cc5ef47d0b6a6960613148a
1.22: kindest/node:v1.22.15@sha256:bfd5eaae36849bfb3c1e3b9442f3da17d730718248939d9d547e86bbac5da586
1.21: kindest/node:v1.21.14@sha256:ad5b7446dd8332439f22a1efdac73670f0da158c00f0a70b45716e7ef3fae20b
1.20: kindest/node:v1.20.15@sha256:45d0194a8069c46483a0e509088ab9249302af561ebee76a1281a1f08ecb4ed3
1.19: kindest/node:v1.19.16@sha256:a146f9819fece706b337d34125bbd5cb8ae4d25558427bf2fa3ee8ad231236f2

## How

Specify extract images including the digest for each k8s version on Makefile of e2e and example.

## Checklist

- [ ] Finish implementation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

PR #705, opened 2023-05-26
Specify kind node image digest

toshipp:
kind project recommends to specify kind node image digest.

Fix: #568

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -1,0 +1,13 @@
+# kind node image version is related to kind version.
+# if you change kind version, also change kind node image version.
+KIND_VERSION=v0.18.0
+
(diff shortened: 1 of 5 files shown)


common.mk
daichimukai, line 1: If I understand correctly, this digests are determined by version of kind. It might be better to manage `KIND_VERSION` in this file too.
https://github.com/topolvm/topolvm/blob/995c3e2dc8e299fbdb8731f447893906b966a4e4/Makefile#LL6C1-L7C1

toshipp: agree. I'll do that.

daichimukai (approved):

peng225 (approved):

PR #709, opened 2023-06-19
feat(chart): allow livenessprobe configuration

maxime1907:


charts/topolvm/Chart.yaml
```diff
@@ -4,7 +4,7 @@ type: application
 home: https://github.com/topolvm/topolvm
 name: topolvm
 description: Topolvm
-version: 11.2.1
+version: 11.3.0
```
toshipp, line 7: We separate code modification from releases. Please revert it.

charts/topolvm/templates/lvmd/daemonset.yaml
```diff
@@ -63,8 +63,10 @@ spec:
               command:
                 - /lvmd
                 - health
-            initialDelaySeconds: 10
-            timeoutSeconds: 3
+            failureThreshold: {{ .Values.livenessProbe.csi_registrar.failureThreshold }}
```
toshipp, line 66: This container is lvmd. 
```suggestion
            failureThreshold: {{ .Values.livenessProbe.lvmd.failureThreshold }}
```

toshipp (changes requested):

charts/topolvm/templates/lvmd/daemonset.yaml
maxime1907: my bad

charts/topolvm/values.yaml
toshipp, line 1: Personally, I think that the values defined by Kubernetes should not be written. But that is not a major objection.

maxime1907: every popular chart lets us configure these parameters because depending on cluster loads, they may differ from defaults.

you can take a look at bitnami charts for instance:
 -https://github.com/bitnami/charts/blob/main/bitnami/apache/values.yaml

toshipp: I mean we need not write values that are the same as kubernetes default but write a blank field.
For example, `failureThreshold` is 3 in default, so leave it to brank instead of specifying 3.
In templates, use `with` or `if` to render the value if it is specified.

maxime1907: oh okay i didnt understand what you meant, gonna change it

toshipp (approved):

llamerada-jp (approved):

PR #723, opened 2023-07-18
Don't re-add finalizer to LogicalVolumes that are about to be deleted

### LINKED ISSUE #715: LogicalVolumes "leak" and slow down provisioning
**Describe the bug**

`LogicalVolume` objects tied to deleted nodes are not cleaned up and cannot be cleaned up, because the `finalizer` field is cleared by the `topolvm-node` on the node itself (which is no longer running).

Over time this results in the number of LV objects building up in Kubernetes.  The controller is constantly trying to process these and that leads to a general slowdown in that and thus in the provisioning of new volumes (as it's spending a lot of time/threads trying to delete LVs which can't be deleted)

According to [the docs](https://github.com/topolvm/topolvm/blob/6ed49a173e07fdfe48fc98af3ea02a1e2ea2903c/docs/crd-logical-volume.md?plain=1#L54):

> `LogicalVolume` is created with a [finalizer](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#finalizers).
When a `LogicalVolume` is being deleted, `topolvm-node` on the target node deletes
the corresponding LVM logical volume and clears the finalizer.

I've seen this happening on two of my own clusters today: one had 400+ stale LVs and another 1200+, built up over a year or so (through multiple topolvm upgrades but there were stale LVs created today - I upgraded to 0.18.1 weeks ago). On the larger cluster it was taking many hours to do a full rolling redeploy of 170 pods over 22 deployments

Note that LVs actually seem to get into two states:

1. They are marked for deletion (have a deletionTimestamp), but the finalizer blocks it
2. They belong to an invalid node (`.spec.nodeName` is no longer a valid node) but _are not_ marked for deletion (and still have the finalizer, so once they're deleted 1 will become an issue)

**Environments**
- Version: 0.18.1
- OS: EKS v0.23

**To Reproduce**

I haven't figured out a surefire way to reproduce this - I can delete a node and the `topolvm-node` process _usually_ seems to pick that up. I suspect that given certain conditions that process is either slow or the pod isn't running or gets terminated too quickly, and it doesn't get to removing the finalizer

**Expected behavior**
If the node that a LogicalVolume belongs to no longer exists, the controller should remove the finalizer itself before asking for its deletion

**Additional context**

These two scripts tidy up the LVs for me, in the short term I'll keep running them on an ad-hoc basis:

1. Delete LVs tied to non-existant nodes:
```
kubectl get logicalvolumes -ojson | jq -c .items[] | grep -vf <(kubectl get no -ojson | jq .items[].metadata.name -r) | jq -r .metadata.name | xargs kubectl delete logicalvolume
```

2. Remove finalizer on deleted LVs to allow deletion to complete:
```
kubectl get LogicalVolume -ojson | jq .items[] -c | grep deletionTimestamp | jq -r .metadata.name | xargs -I{} kubectl patch LogicalVolume {} -p '{"metadata":{"finalizers":null}}' --type merge
```

spmason:
This fixes a race-condition between node_controller and logicalvolume_controller where node_controller removes the finalizer from the LogicalVolume, then the lv_controller re-adds it before the node_controller can delete the LV.

If this happens then the LogicalVolume gets 'stuck' in an undeleteable state, because the node is gone so topolvm-node isn't around to remove the finalizer again.

Over time this can result ina  cluster having many undeleteable LogicalVolumes, which slows down topolvm to the point where it can't provision new volumes at all

Closes #715

### DIFF
--- a/controllers/logicalvolume_controller_test.go
+++ b/controllers/logicalvolume_controller_test.go
@@ -1,0 +1,223 @@
+package controllers
+
+import (
+	"context"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/topolvm/topolvm"
+	topolvmv1 "github.com/topolvm/topolvm/api/v1"
+	proto "github.com/topolvm/topolvm/lvmd/proto"
+	"google.golang.org/grpc"
+	corev1 "k8s.io/api/core/v1"
+	storegev1 "k8s.io/api/storage/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+var volumes = &[]*proto.LogicalVolume{}
+
+type MockVGServiceClient struct {
+}
+
+// GetFreeBytes implements proto.VGServiceClient.
+func (MockVGServiceClient) GetFreeBytes(ctx context.Context, in *proto.GetFreeBytesRequest, opts ...grpc.CallOption) (*proto.GetFreeBytesResponse, error) {
+	panic("unimplemented")
+}
+
+// GetLVList implements proto.VGServiceClient.
+func (c MockVGServiceClient) GetLVList(ctx context.Context, in *proto.GetLVListRequest, opts ...grpc.CallOption) (*proto.GetLVListResponse, error) {
+	return &proto.GetLVListResponse{
+		Volumes: *volumes,
+	}, nil
+}
+
+// Watch implements proto.VGServiceClient.
+func (MockVGServiceClient) Watch(ctx context.Context, in *proto.Empty, opts ...grpc.CallOption) (proto.VGService_WatchClient, error) {
+	panic("unimplemented")
+}
+
+type MockLVServiceClient struct {
+}
+
+// CreateLV implements proto.LVServiceClient.
+func (c MockLVServiceClient) CreateLV(ctx context.Context, in *proto.CreateLVRequest, opts ...grpc.CallOption) (*proto.CreateLVResponse, error) {
+	lv := proto.LogicalVolume{
+		Name:   in.Name,
+		SizeGb: in.SizeGb,
+	}
+	*volumes = append(*volumes, &lv)
+	createResponse := proto.CreateLVResponse{
+		Volume: &lv,
+	}
+	return &createResponse, nil
+}
+
+// CreateLVSnapshot implements proto.LVServiceClient.
+func (MockLVServiceClient) CreateLVSnapshot(ctx context.Context, in *proto.CreateLVSnapshotRequest, opts ...grpc.CallOption) (*proto.CreateLVSnapshotResponse, error) {
+	panic("unimplemented")
+}
+
+// RemoveLV implements proto.LVServiceClient.
+func (MockLVServiceClient) RemoveLV(ctx context.Context, in *proto.RemoveLVRequest, opts ...grpc.CallOption) (*proto.Empty, error) {
+	panic("unimplemented")
+}
+
+// ResizeLV implements proto.LVServiceClient.
+func (MockLVServiceClient) ResizeLV(ctx context.Context, in *proto.ResizeLVRequest, opts ...grpc.CallOption) (*proto.Empty, error) {
+	panic("unimplemented")
+}
+
+var _ = Describe("LogicalVolume controller", func() {
+	ctx := context.Background()
+	var stopFunc func()
+	errCh := make(chan error)
+	var vgService MockVGServiceClient
+	var lvService MockLVServiceClient
+
+	startReconciler := func(suffix string) {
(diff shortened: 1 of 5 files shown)


peng225:
@spmason I run the tests and they all have passed.
Please change the status to "Ready for review" when it is convenient for you.

spmason:
Sorry @peng225 I've had a bit of a meeting-heavy day. I've taken this out of draft now

Are there any tests you'd like me to add for this cange itself? I'm not compeltely sure how I'd go about that but happy to try if you have any pointers

peng225:
@spmason 
> Are there any tests you'd like me to add for this change itself?

I want a test to check if the `LogicalVolumeReconciler` does nothing when the annotation is set on the LogicalVolume.
Currently, there is no test for `controllers/logicalvolume_controller.go`, so I will be happy if you could create it.
(Similar test files in the same directory might be your help.)

You do not need to add a whole test cases for `controllers/logicalvolume_controller.go`, but just one (or a few) case(s) for this PR.

As for `NodeReconciler`, it seems hard to write a test, so you can leave it as it is.

controllers/node_controller.go
```diff
@@ -144,6 +144,11 @@ func (r *NodeReconciler) doFinalize(ctx context.Context, log logr.Logger, node *
 func (r *NodeReconciler) cleanupLogicalVolume(ctx context.Context, log logr.Logger, lv *topolvmv1.LogicalVolume) error {
 	if controllerutil.ContainsFinalizer(lv, topolvm.GetLogicalVolumeFinalizer()) {
 		lv2 := lv.DeepCopy()
+		if lv2.Annotations == nil {
+			lv2.Annotations = make(map[string]string)
+		}
+		// Flag the LV as pending deletion, so the logicalvolume_controller doesn't re-add the finalizer before it sees the deletion
```
peng225, line 150:
```suggestion
		// Flag the LV as pending deletion, so the LogicalVolumeReconciler doesn't re-add the finalizer before it sees the deletion
```

spmason:
OK I've added `logicalvolume_controller_test.go` now

The controller has a dependency on the GRPC VG and LV services, so I've had to mock those out in order to be able to verify the basic functinality. I tried using `gomock` but fell back onto doing it manually, hopefully this is acceptable

controllers/logicalvolume_controller_test.go
```diff
@@ -0,0 +1,223 @@
...
+
+		// Verify
+		// ensure LV has finalizer
+		Eventually(func(g Gomega) bool {
+			err := k8sClient.Get(ctx, client.ObjectKeyFromObject(&lv), &lv)
+			if err != nil {
+				return false
+			}
+			return controllerutil.ContainsFinalizer(&lv, topolvm.GetLogicalVolumeFinalizer())
+		}).Should(Equal(true))
```
peng225, line 183:
```suggestion
		}).Should(BeTrue())
```

controllers/logicalvolume_controller_test.go
```diff
@@ -0,0 +1,223 @@
...
+		lv := setupResources(ctx, "-pendingdeletion")
+
+		// ensure LV gets finalizer
+		Eventually(func(g Gomega) bool {
+			err := k8sClient.Get(ctx, client.ObjectKeyFromObject(&lv), &lv)
+			if err != nil {
+				return false
+			}
+			return controllerutil.ContainsFinalizer(&lv, topolvm.GetLogicalVolumeFinalizer())
+		}).Should(Equal(true))
```
peng225, line 201:
```suggestion
		}).Should(BeTrue())
```

controllers/logicalvolume_controller_test.go
```diff
@@ -0,0 +1,223 @@
...
+		Expect(err).NotTo(HaveOccurred())
+
+		// ensure LV finalizer is removed
+		Eventually(func(g Gomega) bool {
+			err := k8sClient.Get(ctx, client.ObjectKeyFromObject(&lv), &lv)
+			if err != nil {
+				return false
+			}
+			return !controllerutil.ContainsFinalizer(&lv, topolvm.GetLogicalVolumeFinalizer())
+		}).Should(Equal(true))
```
peng225, line 221:
```suggestion
		}).Should(BeTrue())
```

controllers/logicalvolume_controller_test.go
```diff
@@ -0,0 +1,223 @@
...
+			topolvm.GetLVPendingDeletionKey(): "true",
+		}
+		controllerutil.RemoveFinalizer(lv2, topolvm.GetLogicalVolumeFinalizer())
+
+		patch := client.MergeFrom(&lv)
+		err := k8sClient.Patch(ctx, lv2, patch)
+		Expect(err).NotTo(HaveOccurred())
+
+		// ensure LV finalizer is removed
+		Eventually(func(g Gomega) bool {
```
peng225, line 215: Because the LogicalVolume reconciler may run after this check, I would like you to check that the finalizer is not added for 1 or 2 seconds.
I believe you can do it by using `Consistently` instead of `Eventually`.
cf. https://pkg.go.dev/github.com/onsi/gomega#Consistently

spmason: Ah nice. That's done now

peng225 (approved):

satoru-takeuchi (approved):

peng225:
@spmason 

> do you have any idea when the next release might be cut with this change in it?

We usually release a new version around the beginning of each month,
so please wait for our August release.

Issue #726, opened 2023-07-27
allow setting deployment environment variables/GOMAXPROCS in helm charts


juliantaylor:
**What should the feature do:**
It would be nice if the kubernetes helm charts would allow setting environment variables for the golang components via the values.

**What is use case behind this feature:**
Typically in golang programs you want to define the variable GOMAXPROCS to be in the range of the cpu limit of the deployment, otherwise golang uses the hosts logical cpu number as the limit which causes a lot of unnecessary overhead and throttling on larger machines.

(Alternatively you could use a go module like https://github.com/uber-go/automaxprocs to let it tune itself automatically)

For example in our clusters consisting of hardware with 40-128 physical cpus the topolvm-controllermanager uses more than half of a cpu purely for goproc overhead and frequently is throttled when some computation starts up 256 threads for a couple ms, setting GOMAXPROCS to 2 (cpu limit of 1) completely removes the overhead.

daichimukai:
Hi @juliantaylor, 

Thank you for your feature request. We've addressed the issue in #727. This patch will be available after next helm chart release planned in this month.

juliantaylor:
oh that was fast, thanks

PR #727, opened 2023-07-30
helm: allow environment variables configuration

daichimukai:
This commit allows us to set environment variables for all containers that this helm chart installs. This enables us, for instance, to set GOMAXPROCS for Go binaries, which might be appropriate to set along with resources.limits.cpu.

Issue: https://github.com/topolvm/topolvm/issues/726

### DIFF
--- a/charts/topolvm/README.md
+++ b/charts/topolvm/README.md
@@ -104,6 +104,14 @@ You need to configure kube-scheduler to use topolvm-scheduler extender by referr
 | controller.tolerations | list | `[]` | Specify tolerations. # ref: https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/ |
 | controller.updateStrategy | object | `{}` | Specify updateStrategy. |
 | controller.volumes | list | `[{"emptyDir":{},"name":"socket-dir"}]` | Specify volumes. |
+| env.csi_provisioner | list | `[]` | Specify environment variables for csi_provisioner container. |
+| env.csi_registrar | list | `[]` | Specify environment variables for csi_registrar container. |
+| env.csi_resizer | list | `[]` | Specify environment variables for csi_resizer container. |
+| env.csi_snapshotter | list | `[]` | Specify environment variables for csi_snapshotter container. |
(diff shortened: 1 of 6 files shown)


toshipp (approved):

charts/topolvm/README.md
```diff
@@ -104,6 +104,14 @@ You need to configure kube-scheduler to use topolvm-scheduler extender by referr
...
 | controller.updateStrategy | object | `{}` | Specify updateStrategy. |
 | controller.volumes | list | `[{"emptyDir":{},"name":"socket-dir"}]` | Specify volumes. |
+| env.csi_provisioner | list | `[]` | Specify environment variables for csi_provisioner container. |
+| env.csi_registrar | list | `[]` | Specify environment variables for csi_registrar container. |
+| env.csi_resizer | list | `[]` | Specify environment variables for csi_resizer container. |
+| env.csi_snapshotter | list | `[]` | Specify environment variables for csi_snapshotter container. |
+| env.liveness_probe | list | `[]` | Specify environment variables for liveness_probe container. |
+| env.topolvm_controller | list | `[]` | Specify environment variables for topolvm_controller container. |
+| env.topolvm_node | list | `[]` | Specify environment variables for topolvm_node container. |
+| env.topolvm_scheduler | list | `[]` | Specify environment variables for topolvm_scheduler container. |
```
llamerada-jp, line 114: The other parameters are camel cases. It is better to be the same.

llamerada-jp (changes requested):

charts/topolvm/README.md
daichimukai: Certainly most fields are camelCase, but it would be more consistent to make an exception if it represents a container identifier.
 
https://github.com/topolvm/topolvm/blob/d242dcc522ae3f4bc3fe692054205b1076319396/charts/topolvm/README.md?plain=1#L161-L169
https://github.com/topolvm/topolvm/blob/d242dcc522ae3f4bc3fe692054205b1076319396/charts/topolvm/README.md?plain=1#L116-L120

llamerada-jp (approved):

PR #728, opened 2023-07-31
feat: allow passing leaderelection config values to topolvm-controller

jakobmoellerdev:
uses default values from the [controller-runtime manager.go](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/manager/manager.go#L208-L219) and allows configuring leaderelection configuration.

This passes the `leaseDuration`, `renewDeadline` and `retryPeriod` from the `topolvm-controller` cmd down to the startup of the controller.

**Reasoning**: In certain clusters, we need to optimize lvm-controller lease configuration, so that it can efficiently startup, restart or shutdown when a node gets restarted that the controller sits on. Currently we can only adjust `csi-provisioner` and `csi-resizer`, but not the `topolvm-controller`.

Why is this needed?
This PR changes `topolvm-controller` to have these settings, because [`csi-provisioner`](https://github.com/kubernetes-csi/external-provisioner/blob/master/cmd/csi-provisioner/csi-provisioner.go#L97-L99) already has these settings, just like [`csi-resizer`](https://github.com/kubernetes-csi/external-resizer/blob/master/cmd/csi-resizer/main.go#L64-L67). With this change, one can adjust all 3 controller's lease settings and optimize them for their cluster runtime.

Additional Hints:
Most of these settings are not included into the topolvm HELM Chart, except `leaderelection-namespace` as this is also used within [`csi-provisioner`](https://github.com/topolvm/topolvm/blob/main/charts/topolvm/templates/controller/deployment.yaml#L120) and [`csi-resizer`](https://github.com/topolvm/topolvm/blob/main/charts/topolvm/templates/controller/deployment.yaml#L160). It should be backwards-compatible for most, since not setting a namespace uses the pod-namespace by default, which usually is the same as the release namespace in HELM.


charts/topolvm/templates/controller/deployment.yaml
```diff
@@ -42,6 +42,7 @@ spec:
           image: "{{ .Values.image.repository }}:{{ default .Chart.AppVersion .Values.image.tag }}"
           command:
             - /topolvm-controller
+            - --leader-election-namespace={{ .Release.Namespace }}
```
toshipp, line 45: You said that not specifying a namespace means that the controller will use the namespace where the pod exists, so could you let me know why this is needed?

jakobmoellerdev: I mainly did this to conform with the rest of the chart, because you also specified it for the sidecars. If this is not wanted I think we should think about removing it from the other sidecars as well. WDYT?

toshipp: That makes sense. I did not see this file entirely. LGTM!

toshipp (approved):

cupnes:
@jakobmoellerdev Thank you for your contribution. This PR was merged.

PR #738, opened 2023-08-17
feat: allow snapshotting to greater target than source volume

### LINKED ISSUE #695: Support restore to a PVC larger than the VolumeSnapshot
As part of https://github.com/topolvm/topolvm/pull/463 a TODO was added: https://github.com/topolvm/topolvm/blob/a2c90d2b87eaa216bfb00361854cd31ff41f7e7e/driver/controller.go#L180-L184 Was wondering if there was any more context around this, anything that might help tackle the issue.

Would `-L, --size LogicalVolumeSize` on the second snapshot help here? (the one that would be activated an act as the restored vol)

It is not uncommon to want to restore to a larger PVC, and is part of the CSI spec:
https://github.com/ceph/ceph-csi/issues/1566

jakobmoellerdev:
This is a first attempt at resizing snapshots after creation to make them larger than the original Source volume. This allows to request PVCs that consume from the snapshot LV which are larger than source.

Still verifying and testing, after that will put up for review, opening for transparency on plan.

Fix #695


jakobmoellerdev:
@cupnes Please let me know what you think of the provided solution. I think I found a sufficient middle-ground

cupnes:
Could you resize the filesystem after lv resize?

jakobmoellerdev:
LogicalVolumeService does not know about the mount path of the FS right now and provides `ExpandVolume` without Filesystem Resize. How should I implement Resizing in LogicalVolumeService if I do not know the mount path of the fs?

If the LVM snapshot is created, it is not known if the snapshot is of type Block or FS. (I think the VolumeSnapshot Resource contains an alpha field for this but its not usually known).

cupnes:
@jakobmoellerdev You are correct. That was my mistake. Forget my previous comment. I continue to review. Please wait for a while.

cupnes:
@jakobmoellerdev I will confirm so please wait a little longer.

cupnes:
@jakobmoellerdev
How about extending the file system during a NodePublishVolumeRequest?

On the other hand, maybe I should have asked earlier, but how about creating a snapshot volume with the same size as the source volume and then expanding it? Then no additional functionality is needed.

jakobmoellerdev:
> @jakobmoellerdev How about extending the file system during a NodePublishVolumeRequest?

I could introduce something like this:
```go
r := mountutil.NewResizeFs(s.mounter.Exec)
	if resize, err := r.NeedResize(device, req.GetTargetPath()); resize {
		if _, err := r.Resize(device, req.GetTargetPath()); err != nil {
			return status.Errorf(codes.Internal, "failed to resize filesystem %s (mounted at: %s): %v", req.VolumeId, req.GetTargetPath(), err)
		}
	} else if err != nil {
		return status.Errorf(codes.Internal, "could not determine if fs needed resize after mount: target=%s, error=%v", req.GetTargetPath(), err)
	}
```
on every `NodePublishVolume` for a Volume with an FS after the mount succeeded in https://github.com/topolvm/topolvm/blob/main/driver/node.go#L250. 

Is this something you want on this PR or in general? Did not want to touch the Volume Mounting behavior originally but maybe it makes sense.

> On the other hand, maybe I should have asked earlier, but how about creating a snapshot volume with the same size as the source volume and then expanding it? Then no additional functionality is needed.

So this is what I have been doing on lower level right now. I create the snapshot thinly and it will use the source size. After that I call resize on it and it will use the size of the request of fail during resizing: https://github.com/topolvm/topolvm/pull/738/files#diff-43cd6f25262031c6f6eba0651382223717adccec365e50b605765d766e68b21fR254-R260

I think this would still require a Filesystem resize though right?
If you mean something different, please let me know!

cupnes:
@jakobmoellerdev 
> @cupnes I could introduce something like this:

Thank you. LGTM

> Is this something you want on this PR or in general? Did not want to touch the Volume Mounting behavior originally but maybe it makes sense.

Yes, I don't think they should be left with different lv sizes and file system sizes.

> So this is what I have been doing on lower level right now. I create the snapshot thinly and it will use the source size. After that I call resize on it and it will use the size of the request of fail during resizing:

Perhaps I have been misunderstood. My previous comment is a proposal that does not change anything in the code. Even without this PR, the following steps can be used to create a snapshot of a size larger than the source volume.

1. Apply PVC to create a snapshot with the same size as the source volume
2. Edit the PVC applied in 1. to extend the size of `storage:`

(The `AllowVolumeExpansion` of the StorageClass must be set to `true` beforehand).

Another thing I noticed in this PR is that snapshots of thin lv can be created with a larger size than the source volume. For example, the following command. With this, `lvresize` is not necessary.

```
lvcreate -s -k n -n <lv name> -L <the size larger than source volume> <vg name>/<source volume name>
```

jakobmoellerdev:
`lvcreate -s -k n -n <lv name> -L <the size larger than source volume> <vg name>/<source volume name>`

AFAIK, this Snapshot will NOT be thin because you specify a size with `-L`. This is dangerous since after this, the snapshot volume will not be extendable in size and you will also not be able to create Snapshots of snapshots.

Other than that, I will include the part in `NodePublishVolume` so the resize always happens.

I think the case you describe should always be possible, but this PR is about creating a PVC with extended size from the start, not creating it with source size and then editing it.

Still you are right that this can be a workaround for people who do not have this feature yet.

cupnes:
@jakobmoellerdev 
> As far as I know, this Snapshot will NOT be thin because you specify a size with `-L`. This is dangerous since after this, the snapshot volume will not be extendable in size and you will also not be able to create Snapshots of snapshots.

Right. My idea was wrong.

> Other than that, I will include the part in `NodePublishVolume` so the resize always happens.

Thank you. LGTM. But could you add the e2e test to `e2e/e2e_test.go` for the behaviour of resizing the file system during `NodePublishVolume`? An idea for implementation is to create a snapshot larger than the source volume and test that the filesystem of the pod mounting it is extended.

cupnes (dismissed):
Thank you. LGTM

driver/controller.go
```diff
@@ -7,13 +7,14 @@ import (
 	"strings"
 	"time"
 
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
```
toshipp, line 11: We group together 3rd party libraries. If you have no specific reason, please revert it.

driver/node.go
```diff
@@ -9,15 +9,16 @@ import (
 	"strings"
 	"sync"
 
+	"golang.org/x/sys/unix"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
```
toshipp, line 16: dit

lvmd/lvservice_test.go
```diff
@@ -6,11 +6,12 @@ import (
 	"os/exec"
 	"testing"
 
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
```
toshipp, line 11: dit

lvmd/lvservice.go
```diff
@@ -5,11 +5,12 @@ import (
 	"fmt"
 	"math"
 
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
```
toshipp, line 10: dit

toshipp (commented):
LGTM

cupnes (approved):

jakobmoellerdev:
@pluser @ushitora-anqou could you kindly take a look and determine if this can be merged? Got previous reviews from @cupnes and @toshipp

pluser:
Sorry for the delay. I think it's good.

pluser (approved):

PR #741, opened 2023-08-25
remove redundant explanation of snapshot limitation.

peng225:
This PR is created as a result of [this discussion](https://github.com/topolvm/topolvm/discussions/737).

Current limitation says that snapshots can be created only for the bound
PVCs.  This is not a limitation specific to TopoLVM because there is no
way to create a snapshot of a PVC which has no actual data.  I removed
this explanation from the limitation.

### DIFF
--- a/docs/limitations.md
+++ b/docs/limitations.md
@@ -33,12 +33,15 @@ Therefore, when multiple pods requesting TopoLVM volumes are created at once, th
 Note that pod scheduling is also affected by the amount of CPU and memory.
 Because of this, this problem may not be observable.
 
-Snapshots should be created only for a `BOUND` PVC
+Snapshots can be created only for thin volumes
 -------------------------
-Snapshot are currently supported only for thin volumes and is an experimental feature because CSI Sanity is skipped.
-The LVM snapshots are required to be provisioned on the same node as the source logical volume. Therefore, the source PVC must be provisioned before the target so that scheduling decisions can be taken accordingly.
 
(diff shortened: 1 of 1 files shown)


docs/limitations.md
```diff
@@ -33,12 +33,10 @@ Therefore, when multiple pods requesting TopoLVM volumes are created at once, th
 Note that pod scheduling is also affected by the amount of CPU and memory.
 Because of this, this problem may not be observable.
 
-Snapshots should be created only for a `BOUND` PVC
+Snapshots can be created only for thin volumes
 -------------------------
-Snapshot are currently supported only for thin volumes and is an experimental feature because CSI Sanity is skipped.
-The LVM snapshots are required to be provisioned on the same node as the source logical volume. Therefore, the source PVC must be provisioned before the target so that scheduling decisions can be taken accordingly.
```
llamerada-jp, line 39: > The LVM snapshots are required to be provisioned on the same node as the source logical volume.

This limitation is a TopoLVM-specific. I would like to keep this description, especially in case someone wants to restore a snapshot to another node and may ask a question without realizing it. I think it is good to add a new entry to describe this issue because it is different from the current entry's title.

peng225: Do you mean that restored volumes are required to be provisioned on the same node as source volumes? If so, the current explanation does not seem to say so. I will add another section to explain the limitation of the restore.

peng225: I updated the document.

pluser (approved):

llamerada-jp (approved):

PR #749, opened 2023-09-18
feat: add option to enable/disable webhook server

suleymanakbas91:
This PR adds a new flag called `enable-webhooks` to TopoLVM Controller. The flag allows users to enable or disable the webhook server so that users, who do not deploy Webhooks, do not need to manage any certs either.

### DIFF
--- a/charts/topolvm/templates/certificates/certificates.yaml
+++ b/charts/topolvm/templates/certificates/certificates.yaml
@@ -1,3 +1,4 @@
+{{- if or .Values.webhook.podMutatingWebhook.enabled .Values.webhook.pvcMutatingWebhook.enabled }}
 {{- if not .Values.webhook.caBundle }}
 {{- if not .Values.webhook.existingCertManagerIssuer }}
 # Generate a CA Certificate used to sign certificates for the webhook
@@ -51,4 +52,5 @@ spec:
     - key encipherment
     - server auth
     - client auth
+{{- end }}
 {{- end }}
--- a/charts/topolvm/templates/certificates/issuers.yaml
+++ b/charts/topolvm/templates/certificates/issuers.yaml
@@ -1,3 +1,4 @@
+{{- if or .Values.webhook.podMutatingWebhook.enabled .Values.webhook.pvcMutatingWebhook.enabled }}
 {{- if not .Values.webhook.caBundle }}
 {{- if not .Values.webhook.existingCertManagerIssuer }}
 # Create a selfsigned Issuer, in order to create a root CA certificate for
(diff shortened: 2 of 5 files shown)


toshipp (commented):
We can disable both webooks by setting webhook.podMutatingWebhook.enabled and pvcMutatingWebhook to false, and also the certificate can be disabled by webhook.caBundle.
Why do we need another option?

suleymanakbas91:
> We can disable both webooks by setting webhook.podMutatingWebhook.enabled and pvcMutatingWebhook to false, and also the certificate can be disabled by webhook.caBundle. Why do we need another option?

Even if we disable the Webhooks as you suggest, TopoLVM Controller still starts the Webhook Server [in this line](https://github.com/topolvm/topolvm/pull/749/files#diff-5340e6ca13ed51b48e0238ce4cf4e896d88b488cfadb315f05c49d59958b5167R87). It looks for the certificates and fails to start if none exists. There should be an option to disable Webhook Server as well for the users who do not deploy any Webhooks. 

I can remove `webhook.enabled` and instead use `webhook.podMutatingWebhook.enabled` and `webhook.pvcMutatingWebhook.enabled` to set the new [enable-webhooks](https://github.com/topolvm/topolvm/pull/749/files#diff-566b26882305db360f29109740556b7222626c9bb52dbb7fcdb279700a062bf7R59) flag in TopoLVM Controller, if that's preferable.

toshipp:
I understand your motivation.

The webhook for PVC is to delete a pod when deleting the PVC used by it, this function is optional. On the other hand, the webhook for Pod is required for topolvm-scheduler. So, they should be configurable separately.

I have a just idea to solve the problem. Do we disable the certificate if both webhooks are disabled?

Anyway, the controller refers to the secret made from the certificate. I assume we should disable volume and volumeMounts.
https://github.com/topolvm/topolvm/blob/26c290784036df4223ba73bfa2309ef7293af445/charts/topolvm/templates/controller/deployment.yaml#L248

suleymanakbas91:
> I understand your motivation.
> 
> The webhook for PVC is to delete a pod when deleting the PVC used by it, this function is optional. On the other hand, the webhook for Pod is required for topolvm-scheduler. So, they should be configurable separately.
> 
> I have a just idea to solve the problem. Do we disable the certificate if both webhooks are disabled?
> 
> Anyway, the controller refers to the secret made from the certificate. I assume we should disable volume and volumeMounts.
> 
> https://github.com/topolvm/topolvm/blob/26c290784036df4223ba73bfa2309ef7293af445/charts/topolvm/templates/controller/deployment.yaml#L248

Yes, I agree that's a better approach. I've changed the implementation accordingly. Could you take another look?

toshipp (approved):
LGTM

peng225 (approved):

PR #754, opened 2023-09-27
docs: Enhancement Proposal: PVCs < 1 GB in TopoLVM

jakobmoellerdev:
Related to https://github.com/topolvm/topolvm/discussions/646

## Motivation

For most users of TopoLVM, local nodes are the focus. 
While TopoLVM was designed to be run in datacenters where the lowest provisional unit might not be an issue, edge use cases require us to create many volumes in a small amount of storage which may only be a few gigabytes.
To allow TopoLVM for proper use on edge devices and small form-factor use cases of Pods, we need to allow storage capacities that can supply storage lower than 1GB.

### DIFF
--- a/docs/proposals/pvcs-smaller-than-1gi.md
+++ b/docs/proposals/pvcs-smaller-than-1gi.md
@@ -1,0 +1,550 @@
+---
+title: pvcs-smaller-than-1gi
+authors:
+  - "@jakobmoellerdev"
+reviewers:
+  - "@pluser"
+  - "@toshipp"
+  - "@suleymanakbas91"
+  - "@jeff-roche"
+  - "@brandisher"
+approvers:
+  - "@pluser"
+  - "@toshipp"
+api-approvers:
+  - "@toshipp"
+creation-date: 2023-09-23
+last-updated: 2023-09-26
+---
+
+# Supporting PVCs smaller than 1 Gi in size
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+    - [Goals](#goals)
+- [Proposal](#proposal)
+    - [Decision Outcome](#decision-outcome)
+    - [Open Questions](#open-questions)
+- [Design Details](#design-details)
+    - [How to pass byte-level capacities to TopoLVM?](#how-to-pass-byte-level-capacities-to-topolvm)
+    - [How to get around the `<<30`/`>>30` bit shifts](#how-to-get-around-the-3030-bit-shifts)
+    - [Impact Analysis on Existing PVC creation](#impact-analysis-on-existing-pvc-creation)
+    - [Changing how `LogicalVolumeService` is creating `LogicalVolumeSpec`](#changing-how-logicalvolumeservice-is-creating-logicalvolumespec)
+    - [Modifying `controllers/logicalvolume_controller` to create correct LVMD calls via gRPC](#modifying-controllerslogicalvolumecontroller-to-create-correct-lvmd-calls-via-grpc)
+    - [Switching all occurrences of `uint64` to `int64` in lvmd and change `size_gb` to `size` in protobuf messages.](#switching-all-occurrences-of-uint64-to-int64-in-lvmd-and-change-sizegb-to-size-in-protobuf-messages)
+    - [Dealing with less than 1Gi storage in the TopoLVM scheduler prioritization](#dealing-with-less-than-1gi-storage-in-the-topolvm-scheduler-prioritization)
+- [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+- [Deprecation / Removal](#deprecation--removal)
+<!-- /toc -->
+
+## Summary
+
+Currently, TopoLVM lives with a design decision to have the smallest amount of storage be limited to 1 GB.
+This is mainly due to various bit shifting and assumptions made in the code and not because of limitations in CSI, Kubernetes or LVM.
+We need to remove this sizing limitation.
+
+## Motivation
+
+For most users of TopoLVM, local nodes are the focus. 
+While TopoLVM was designed to be run in datacenters where the lowest provisional unit might not be an issue, edge use cases require us to create many volumes in a small amount of storage which may only be a few gigabytes.
+To allow TopoLVM for proper use on edge devices and small form-factor use cases of Pods, we need to allow storage capacities that can supply storage lower than 1GB.
+
+### Goals
+
+- Ensure backwards-compatibility with schedulers and legacy bit shifts that would be replaced
+- Discover all places with hardcoded bit shifts and transition them to a more modular and sizable approach
+- Allow creation of PVCs that request less than 1GB of storage, with a minimum size that is reasonably set.
+
+### Current Workflow
+
+1. TopoLVM administrator installs TopoLVM as usual
+2. Cluster user creates PVCs with a Storage Class backed by TopoLVM following [the process of PVC creation during dynamic provisioning](../design.md#how-dynamic-provisioning-works):
+    ```bash
+    $ cat <<EOF | kubectl apply -f -
+    apiVersion: v1
+    kind: PersistentVolumeClaim
+    metadata:
+      name: test-claim
+    spec:
+      storageClassName: vg1
+      resources:
+        requests:
+          storage: 500Mi
+      accessModes:
+        - ReadWriteOnce
+      volumeMode: Filesystem
+    EOF
+    ```
+3. Once the PVC is used, the [`convertRequestCapacity` method in `topolvm-controller` and `topolvm-node`](../../driver/controller.go) kicks in and defaults to `1Gi` of storage as request capacity, resulting in an error
+   if we have only `500Mi` or less storage available to us. We want to change step 3 to allow capacities below 1Gi of storage.
+
+## Proposal
+
+The main idea of the proposal is to remove all in-code limitations that are currently in place that prohibit us from allowing sizing less than 1Gi to any PVCs.
+
+The main reason of hard coding 1Gi as minimum size is the `<<30`/`>>30` bitshift.
+Any occurrence of the bit shift needs to be identified and replaced with a more dynamic version.
+
+In total the entire codebase contains 73 occurrences of `<<30` and 8 occurrences of `>>30` as of writing this proposal.
+While it would not be completely trivial to replace, we believe this is still a reasonable amount to refactor with manual care.
+
+This means we need to **find all occurrences of `<<30` Bit shifts and `GB` values and replace them with _byte-level_ comparisons**.
+
+Breaking this down we will tackle the following questions:
+1. [How to pass byte-level capacities to TopoLVM?](#how-to-pass-byte-level-capacities-to-topolvm) 
+    
+    This will allow us to actually get the necessary capacity precision required to support less than 1Gi PVCs.
+    We want to achieve this by using the inbuilt [byte-level precision in the CSI specification](https://github.com/container-storage-interface/spec/blob/master/spec.md#controller-service-rpc).
+
+2. [How to get around the `<<30`/`>>30` bit shifts?](#how-to-get-around-the-3030-bit-shifts) 
+
+    This will allow us to use code paths in TopoLVM that do not require comparisons with Gi-level precision.
+    We are showcasing how it is possible to get around the bitshift and how to replace it with byte-level comparisons.
+
+### API Changes
+
+Additionally, as a result of an [Impact Analysis of the code path when creating a PVC](#impact-analysis-on-existing-pvc-creation) we will be facing two breaking API changes:
+
+1. **Any request/response made with lvmd using a `size_gb` field needs to be changed to a different `size` because otherwise we cannot communicate less than 1Gi capacities to `lvmd`.**
+2. **The [current prioritisation algorithm within the TopoLVM Scheduler Extension](../topolvm-scheduler.md#prioritize) needs to be adjusted and its current version needs to specify a limitation when working with storage less than 1Gi**
+
+### Code Changes related to the API 
+
+We are proposing how to change the current codebase [considering the breaking changes in the API](#api-changes):
+- How to deal with breaking changes in LVMD?
+   - [Switching all occurrences of `uint64` to `int64` in lvmd and change `size_gb` to `size` in protobuf messages.](#switching-all-occurrences-of-uint64-to-int64-in-lvmd-and-change-sizegb-to-size-in-protobuf-messages)
+- How to touch important components which are using bit shifts that were previously using `size_gb`
+    - [Changing how `LogicalVolumeService` is creating `LogicalVolumeSpec`](#changing-how-logicalvolumeservice-is-creating-logicalvolumespec)
+    - [Modifying `controllers/logicalvolume_controller` to create correct LVMD calls via gRPC](#modifying-controllerslogicalvolumecontroller-to-create-correct-lvmd-calls-via-grpc)
+- How to deal with breaking changes in the TopoLVM Scheduler?
+   - [Dealing with less than 1Gi storage in the TopoLVM scheduler prioritization](#dealing-with-less-than-1gi-storage-in-the-topolvm-scheduler-prioritization)
+
+### Decision Outcome
+
+
+### Open Questions
+
+Under the assumption that the change is deemed useful or accepted, we still need to decide on the following points:
+
+1. Should `lvmd` protobuf message set be considered `user-facing`? If so should we `reserve` or `deprecate` during the change? If not, we can remove the field. For more details see [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+2. Should we duplicate test cases for less than 1Gi volume sources, or adjust existing tests to cover the new scenarios?
+3. Should the old scheduling algorithm be kept with a new version to choose from and noted as being limited in the old version for storage less than 1Gi or should it be replaced/removed?
+
+## Design Details
+
+
+### How to pass byte-level capacities to TopoLVM?
+
+The answer to this question lies in `topolvm-controller` and `topolvm-node` when called by Kubernetes.
+
+With every call made to a volume request (e.g. `CreateVolume`), a `CapacityRange` is included:
+
+```go
+// The capacity of the storage space in bytes. To specify an exact size,
+// `required_bytes` and `limit_bytes` SHALL be set to the same value. At
+// least one of the these fields MUST be specified.
+type CapacityRange struct {
+	// Volume MUST be at least this big. This field is OPTIONAL.
+	// A value of 0 is equal to an unspecified field value.
+	// The value of this field MUST NOT be negative.
+	RequiredBytes int64 `protobuf:"varint,1,opt,name=required_bytes,json=requiredBytes,proto3" json:"required_bytes,omitempty"`
+	// Volume MUST not be bigger than this. This field is OPTIONAL.
+	// A value of 0 is equal to an unspecified field value.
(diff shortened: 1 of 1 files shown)


docs/proposals/pvcs-smaller-than-1gi.md
```diff
@@ -0,0 +1,543 @@
...
+
+This leads to every free Capacity between 1 byte and 1GB free to get the score of `1`, which means they will be scored the same.
+The only downside to this algorithm is that its precision is 1GB. So if there are 2 nodes with `500Mi` and `700Mi` free storage, both would receive the same score.
+
+The only way to get around this is by
+1. Changing the way existing scoring works to make room for byte-level comparison
+2. Creating a new scoring algorithm side-by-side and allow users to switch
+
+If we accept changing existing scoring under the assumption that all scored nodes will stay in the same order, just with higher precision we can just make use
+of the existing `item.Annotations[topolvm.GetCapacityKeyPrefix()+dc]` and make use of this byte amount (as `int64`) directly for scoring.
```
toshipp, line 502: The score value is limited in [0, 10]. This is why we use log2 to calculate the score value.
https://github.com/kubernetes/kube-scheduler/blob/master/extender/v1/types.go#L29

toshipp (commented):
I think the proposal can be acceptable. I left some comments we need to consider more.

docs/proposals/pvcs-smaller-than-1gi.md
```diff
@@ -0,0 +1,543 @@
...
+Additionally, any TopoLVM user making use of the [scheduler extension](../topolvm-scheduler.md) will also be impacted by the following changes:
+
+1. > `topolvm-node` exposes free storage capacity as `capacity.topolvm.io/<device-class>` annotation of each Node.
+
+   These capacities are already reported on byte-level, not in Gi, so they do not need to be touched
+2. > `topolvm-controller` works as a [mutating webhook](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/) for new Pods.
+   > - It adds `capacity.topolvm.io/<device-class>` annotation to a pod and `topolvm.io/capacity` resource to the first container of a pod.
+   > - The value of the annotation is the sum of the storage capacity requests of unbound TopoLVM PVCs for each volume group referenced by the pod.
+
+   This logic is not impacted by sizing less than 1Gi and can be left untouched
```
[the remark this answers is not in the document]
jakobmoellerdev: You are absolutely right, I missed this originally.

This logic currently rounds up to the next GB value and needs to be modified from `requested = ((req.Value()-1)>>30 + 1) << 30` to 
`requested = req.Value()` 
for the Mutator in all places.

Otherwise, the capacity annotation will be incorrect.

docs/proposals/pvcs-smaller-than-1gi.md
jakobmoellerdev: While implementing this btw I noticed that node scoring is inaccurate for multiple deviceClasses as it only uses the smallest available device Class for the host priority. Is that a documented limitation?

docs/proposals/pvcs-smaller-than-1gi.md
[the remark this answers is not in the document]
toshipp: >This logic currently rounds up to the next GB value and needs to be modified from
>requested = ((req.Value()-1)>>30 + 1) << 30 to
>requested = req.Value()
>for the Mutator in all places.

confirmed.

> Actually also then we also don't need the DefaultCapacity if we are on anything > 0.
Yeah. We can delete it entirely if we use byte units. IMO.

docs/proposals/pvcs-smaller-than-1gi.md
toshipp: >a linear ratio conversion instead of having to rely on log2

Could you explain why you prefer a linear ratio?
We use log2 for scoring because the score range is much smaller than PVC size. Using log2, we can handle sizes from 1GB(2^0) to 1TB(2^10).
I assume you use MB-sized PVCs, log2 produces 0, so it is meaningless. Right?
On the other hand, using a linear ratio, we could not handle GB or TB PVC.
I have no clear idea how to solve this now, but it may be helpful to add nobs to change algorithms or upper limits.

>it only uses the smallest available device Class for the host priority.

The intention of this scoring is to select nodes with high capacity in order to leave room for resizing. The capacity of the smallest dc is therefore used to score the nodes.

jakobmoellerdev: > Could you explain why you prefer a linear ratio?

So the advantage of a linear ratio is that we will always preserve the order of the nodes during scoring. The calculated scores may be different from log2, but we will always have the correct direction, no matter if bytes, gb, tb, etc. I tried to explain this in https://github.com/topolvm/topolvm/pull/754/files#diff-a5e76d9046c66bb6541fa6fa58911fdd82a13dfff9921cfa973e6826c78280cbR574-R580

> On the other hand, using a linear ratio, we could not handle GB or TB PVC.

I don't think this is the case. We can still handle GB or TB PVCs with linear ratio since it would work with byte precision. Linear algorithm guarantees to always preserve the correct order, which I think is what we want.

> The intention of this scoring is to select nodes with high capacity in order to leave room for resizing. The capacity of the smallest dc is therefore used to score the nodes.

This makes sense for one deviceClass. But if you have more than one deviceClass, then choosing the smallest deviceClass does not help the scheduling in the larger deviceClass right?

toshipp: I understand your algorithm now.
I could not find the algorithm before because github does not show diffs properly when doing force-push. So please don't use force-push.

>choosing the smallest deviceClass does not help the scheduling in the larger deviceClass right?
This is intentional in the multiple deviceClasses scenario too.
We don't know which deviceClass PVCs users will extend. At least other DCs are more expandable than the smallest DC, so choosing the smallest DC is a safe choice.

jakobmoellerdev: > I understand your algorithm now. I could not find the algorithm before because github does not show diffs properly when doing force-push. So please don't use force-push.

Understood, thanks for bringing that up. I will do regular merges until we are finished on reviews and then squash stuff together.

> > choosing the smallest deviceClass does not help the scheduling in the larger deviceClass right?
> 
> This is intentional in the multiple deviceClasses scenario too. We don't know which deviceClass PVCs users will extend. At least other DCs are more expandable than the smallest DC, so choosing the smallest DC is a safe choice.

Maybe we are drifting off from the proposal here and I will open up a separate discussion. 

I believe I still don't see how using the smallest deviceClass is a safe choice. A deviceClass availableBytes is backed by a volumegroup, so just because one volumegroup has capacity does not mean the other volumegroup can be expanded to the same amount right? If you take the smallest available capacity from a volumegroup on one node, that does not mean that the other volume groups can be extended by the same amount, because the volume groups contain different devices.

I will resolve this conversation and then I will open up something separate for my question because I do not think it is related to the proposal changes.

toshipp: To be clear, do you expect to replace the original scoring algorithm with yours? Since the original algorithm uses log2 of absolute capacity as a score, using a minimum score ensures a minimum usable capacity. However, the ratio comparison makes no such assumption.

toshipp (commented):
Please check this comment.
https://github.com/topolvm/topolvm/pull/754#discussion_r1345407304

docs/proposals/pvcs-smaller-than-1gi.md
jakobmoellerdev: Not sure if I understand your question correctly. The ratio comparison algorithm I showcased also has a case for minimum available capacity ([in this case it just checks >0](https://github.com/topolvm/topolvm/pull/754/files#diff-a5e76d9046c66bb6541fa6fa58911fdd82a13dfff9921cfa973e6826c78280cbR598-R602), but that can be adjusted). That way you will always score with 1 if there is a given minimum usable capacity.

I think we do not have to replace the old scoring algorithm but we should offer it alongside the other one. Otherwise we risk breaking legacy behavior.

I think we can simply switch algorithms via flag / env whatever we decide on.

jakobmoellerdev:
@toshipp @satoru-takeuchi @pluser would it be possible to get a general thumbs-up on this on a high level within this week? we would like to collect feedback and start implementing this and are eagerly waiting for you to take a look at it. I know that some details on the scheduler might still be TBD but the rest should be fine, right?

toshipp:
@jakobmoellerdev
I have no major objections to the high-level design.
@satoru-takeuchi does not work on this, could you remove him from reviewers?

docs/proposals/pvcs-smaller-than-1gi.md
toshipp: >I think we do not have to replace the old scoring algorithm but we should offer it alongside the other one. Otherwise we risk breaking legacy behavior.
>I think we can simply switch algorithms via flag / env whatever we decide on.

understood. thanks.

P.S. I left my concern, but it must not be solved now.
Assume that node A has a capacity of 1G for dc1 and 9G for dc2, and node B has a capacity of 4G for dc1 and 2G for dc2.
In this case, the old algorithm scores log(1) for node A and log(2) for node B, so node B wins.
On the other hand, your algorithm scores 1/4 for node A and 2/9 for node B, so node A wins.
The former ensures at least 2G usable capacity for both DCs, but the latter can only provide 1G for dc1.

pluser:
@jakobmoellerdev 
Sorry for the delay in responding.
I have read through the discussion and proposal so far. As for me, I have no particular questions.

docs/proposals/pvcs-smaller-than-1gi.md
```diff
@@ -0,0 +1,706 @@
...
+
+```go
+    func capacityToScore(capacity uint64, divisor float64) int {
+        // avoid logarithm of zero, which diverges to negative infinity.
+        if capacity == 0 {
+            return 0
+        }
+    
+    return score
+}
```
toshipp, line 485: Did you forget to change this?

docs/proposals/pvcs-smaller-than-1gi.md
```diff
@@ -0,0 +1,706 @@
...
+The resulting `CreateLV` call for example would look almost exactly like the original `CreateLV`:
+
+```go
+    func (s *lvService) CreateLV(_ context.Context, req *proto.CreateLVRequest) (*proto.CreateLVResponse, error) {
+        dc, err := s.dcmapper.DeviceClass(req.DeviceClass)
+        // ...
+        vg, err := command.FindVolumeGroup(dc.VolumeGroup)
+        // ...
+        oc := s.ocmapper.LvcreateOptionClass(req.LvcreateOptionClass)
++/-     requested := req.GetSize()
```
toshipp, line 420: When updating, the old lvmd and the new lvmd work simultaneously. Also, updating the controller and lvmd is not atomic.
Should we use `size_bg` when `size` is missing? Also, should we set `size_gb` along with `size` in response?

docs/proposals/pvcs-smaller-than-1gi.md
jakobmoellerdev: Interesting scenario, I just went and added these test cases into the test function below and i think the results are slightly different.

```go
func TestCapacityToScore(t *testing.T) {
	testCases := []struct {
		input   uint64
		divisor float64
		expect  int
	}{
		// NodeA - dc1=1Gi,dc2=9Gi
		// NodeB - dc1=4Gi,dc2=2Gi
		// dc1
		// NodeA
		{1 << 30, 1, 0},
		// NodeB
		{4 << 30, 1, 2}, // NodeB chosen for dc1
		// dc2
		// NodeA
		{9 << 30, 1, 3}, // NodeA chosen for dc2
		// dc2
		// NodeB
		{2 << 30, 1, 1},
		// NodeA: dc1=0,dc2=3, min-score=0
		// NodeB: dc1=2,dc2=1, min-score=2
		// Result => NodeB wins
	}

	for _, tt := range testCases {
		score := capacityToScore(tt.input, tt.divisor)
		if score != tt.expect {
			t.Errorf("score incorrect: input=%d expect=%d actual=%d",
				tt.input,
				tt.expect,
				score,
			)
		}
	}
}

func TestCapacityToScoreByRatio(t *testing.T) {
	testCases := []struct {
		input  string
		max    string
		expect int64
	}{
		// NodeA - dc1=1Gi,dc2=9Gi
		// NodeB - dc1=4Gi,dc2=2Gi
		// dc1-max=4Gi
		// dc2-max=9Gi
		//
		// dc1
		// NodeA
		{"1Gi", "4Gi", 2},
		// NodeB
		{"4Gi", "4Gi", 10}, // NodeB chosen for dc1
		// dc2
		// NodeA
		{"9Gi", "9Gi", 10}, // NodeA chosen for dc2
		// NodeB
		{"2Gi", "9Gi", 2},
		// NodeA: dc1=2,dc2=10, min-score=2
		// NodeB: dc1=10,dc2=2, min-score=2
		// Result => both nodes weighed equally, both nodes win equally
	}

	for i, tt := range testCases {
		t.Run(fmt.Sprintf("%v", i), func(t *testing.T) {
			in := resource.MustParse(tt.input)
			max := resource.MustParse(tt.max)
			score := capacityToScoreByRatio(in.Value(), max.Value())

			if score != tt.expect {
				t.Errorf("score incorrect: input=%v, max=%v expect=%v actual=%v",
					tt.input,
					tt.max,
					tt.expect,
					score,
				)
			}
		})

	}
}
```

As you can see in this attachment, the old algorithm indeed chooses node A as you said, but the proposed algorithm actually scores both nodes equally because both device classes have the same ratio of free space towards each other. meaning that pvcs will get distributed on both nodes. This is because the scoring algorithm actually chooses only the lowest scored dc for the entire node, even though technically this is inaccurate.

This is ultimately the issue i described above. Scoring via lowest scored device class per  node leaves out all other device classes for scoring. I believe this scoring is highly flawed because just because device-class1 has lots of capacity does not mean you can use it (e.g. for device-class2) without manual intervention

docs/proposals/pvcs-smaller-than-1gi.md
jakobmoellerdev: That is actually a great idea and concern! Like you suggested, In this case we should probably write a logic that is able to work with both like you suggested:

Assuming 
```protobuf
    // Represents the input for CreateLV.
    message CreateLVRequest {
      string name = 1;              // The logical volume name.
+/-   uint64 size_gb = 2 [deprecated = true];
      repeated string tags = 3;     // Tags to add to the volume during creation
      string device_class = 4;
      string lvcreate_option_class = 5;
+     int64 size = 6; // Volume size in bytes
    }
```

we can write 

```go
if req.GetSize() != 0 {
   requested = req.GetSize()
} else {
   requested := int64(req.GetSizeGb() << 30)
}
```

and just set both in the request like you suggested.

That saves us in the asynchronous updates of lvmd and topolvm because both will be able to deal with this. However I want to add that technically, after redeployment is completed, the errors would disappear so this might be acceptable.

docs/proposals/pvcs-smaller-than-1gi.md
jakobmoellerdev: Yup youre right, I adjusted this to reflect the comment above

docs/proposals/pvcs-smaller-than-1gi.md
toshipp: I think my suggestion is safe, but it is a rare case where there are in-flight requests when updating, and it does not affect the high-level design, so you are free to do this or not when implementing.

docs/proposals/pvcs-smaller-than-1gi.md
toshipp: > I also noticed that the pod capacity annotations are completely ignored right now, so if it expects 5 gb on dc1 and the node doesnt have that, its not reflected in the scoring.

Indeed. The annotations are only used for filtering. It may be better to take the request size into account. However, scoring is calculated one pod at a time, so overall optimisation may not be achieved. For this reason, we now only consider the capacity of the nodes.

jakobmoellerdev: > It is because we use int precision. If I choose more appropriate values, A will win, IMO.

Of course if we have other ratios, the scores will change. Im not sure if this is related to int precision though. Could you elaborate?

> I described above, the old algorithm is to choose a node that has room to expand volumes even if you choose any DCs.

That does not make sense to me, sorry that Im taking so long to understand. How do you know if you can expand the volume by choosing A over B? The capacity is reporting volume group capacities, not device capacities. That means you cannot derive if you can expand based on this, right? Since you do not take into account the pod capacity, you dont know what is a better candidate.

That aside: we should definetly not replace the old algorithm for now and simply introduce a score of 1 for all capacities 1Gi>capacity>0 like suggested in option A. I believe we need a separate ideation for changing the scheduler.
 
I think if we introduce a second capacity scheduling algorithm besides the original one, we should be safe. Then we make the new one opt-in and for upgrades, everything will run the same for capacities above 1Gi. What do you think?

toshipp: >Of course if we have other ratios, the scores will change. Im not sure if this is related to int precision though. Could you elaborate?

I wanted to say that the values I gave as examples were not examples of A winning on the new algorithm when calculated with Int.
Here is a correct example A wins on the new algorithm but B wins on the old.
```
A: dc1: 1G, dc2: 7G
B: dc1: 2G, dc2: 3G
```

>That does not make sense to me, sorry that Im taking so long to understand.

Not your fault. I tried to explain but I couldn't do well.

>How do you know if you can expand the volume by choosing A over B? The capacity is reporting volume group capacities, not device capacities. That means you cannot derive if you can expand based on this, right? Since you do not take into account the pod capacity, you dont know what is a better candidate.

As you know, prioritization will run after filtering nodes, so we can choose any node from all possible nodes. And we can't know which volumes users will expand **in the future**. So we assume that all volumes will be extended. With such an assumption, we **think** it is better to choose a node that has the highest minimum usable capacity of DCs.

Consider my example, both node A and node B are candidates, the volume using dc1 is at most 1G and another volume using dc2 is at most 3G because filtering is already done. In this case, the better choice is B because A does not provide more capacity for dc1 if a volume using dc1 is 1G. Of course, if the volume using dc2 is 3G and a user wants to expand the volume, A should be chosen.
As you can see, there is no best algorithm, but the old one works well in our experience and we do not want to break it.

>That aside: we should definetly not replace the old algorithm for now and simply introduce a score of 1 for all capacities 1Gi>capacity>0 like suggested in option A. I believe we need a separate ideation for changing the scheduler.
>
>I think if we introduce a second capacity scheduling algorithm besides the original one, we should be safe. Then we make the new one opt-in and for upgrades, everything will run the same for capacities above 1Gi. What do you think?

It is awesome. I approve of the idea.

jakobmoellerdev: Thanks so much for explaining, now it actually makes so much more sense to me!! And you are completely right, in this case the new algorithm proposal does not achieve our goal and I believe we should stick with the old one.

I now removed the new algorithm proposal with the note to create a separate enhancement proposal on how to enhance the scheduler precision.

I think with this we have the last big question of the proposal answered: We will simply stick to Gi level precision in the scheduler, and live with the fact that all capacities <1Gi are scored the same.

toshipp: Understood. Thanks, too.

toshipp (commented):
@jakobmoellerdev 
LGTM.
Could you squash these commits to merge this?

toshipp (approved):

jakobmoellerdev:
@pluser Would you mind checking a final time over this? With your approval this can go in I think. I would start working on that as soon as we have an agreement

pluser (approved):

PR #766, opened 2023-10-17
feat: allow PVCs <1 Gi

jakobmoellerdev:
Implements PVC support with increments in byte precision instead of gi precision.

For tradeoffs and limitations see the dedicated Design Proposal in https://github.com/topolvm/topolvm/blob/main/docs/proposals/pvcs-smaller-than-1gi.md


jakobmoellerdev:
i think the one failing test is a flake and needs a retrigger. Other than that the PR is dev complete. If accepted I will squash and add a resolution to the Proposal this PR implements.

jakobmoellerdev:
@toshipp It would be awesome if you could take a look here as well since you reviewed the design proposal.

lvmd/command/lvm.go
peng225, line 1: Don't you need to use byte instead of GiB in `func (l *LogicalVolume) Snapshot`, too?
https://github.com/topolvm/topolvm/blob/6f844ad86215113c6b04f3d9b69a83a608abd587/lvmd/command/lvm.go#L507-L522

lvmd/lvservice.go
```diff
@@ -45,7 +45,17 @@ func (s *lvService) CreateLV(_ context.Context, req *proto.CreateLVRequest) (*pr
 		return nil, err
 	}
 	oc := s.ocmapper.LvcreateOptionClass(req.LvcreateOptionClass)
-	requested := req.GetSizeGb() << 30
+
+	var requested uint64
```
peng225, line 49: In the proposal, you says that the type of the size should be changed from uint64 to int64 in lvmd, but it seems that you don't do that in this PR. Could you tell me why?
https://github.com/topolvm/topolvm/blob/main/docs/proposals/pvcs-smaller-than-1gi.md#switching-all-occurrences-of-uint64-to-int64-in-lvmd-and-change-size_gb-to-size-in-protobuf-messages

peng225 (commented):
Though I'm still reviewing your PR, I posted the comments so far.

driver/controller_test.go
```diff
@@ -4,51 +4,67 @@ import (
...
+	v, err := convertRequestCapacityBytes(0, 10)
 	if err != nil {
 		t.Error("should not be error")
 	}
-	if v != 1 {
-		t.Errorf("should be 1: %d", v)
+	if v != 10 {
+		t.Errorf("should be the limit capacity by default if 0 is supplied and limit is smaller than 1Gi: %d", v)
+	}
+	v, err = convertRequestCapacityBytes(0, 2>>30)
```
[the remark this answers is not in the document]
jakobmoellerdev: Good catch, I had a typo in here. The test case should test that if req is 0 and limit is 2 gi, then we default to 1 gi requested. Changed accordingly

lvmd/command/lvm.go
jakobmoellerdev: Good catch. We actually never call this code in TopoLVM because we officially dont support thick snapshots. I still adjusted it now to not use Gi precision, but maybe we should simply remove it?

lvmd/lvservice.go
jakobmoellerdev: You are right that was my initial idea. However, I did notice that if we stick to uint64 precision in lvmd domain, we do not have to touch the lvm_json parsing and casting from int64 to uint64 is safe as long as int64 is positive (which it is by CSI Spec). Thus it requires less code changes to do the conversion here instead of changing lvmd to run with int64. 

If you want, I can change to int64 in all places but it will require more code changes. What is your thought?

scheduler/prioritize.go
```diff
@@ -17,6 +17,13 @@ func capacityToScore(capacity uint64, divisor float64) int {
 
 	// avoid logarithm of zero, which diverges to negative infinity.
 	if gb == 0 {
+		// if there is a non-nil capacity but we dont have at least one gigabyte, we score it with one.
+		// this is because the capacityToScore precision is at the gigabyte level.
+		// TODO: introduce another scheduling algorithm for byte-level precision.
+		if capacity > 0 {
+			return 1
+		}
```
llamerada-jp, line 25: @jakobmoellerdev @toshipp 
When capacity is between 1 byte and 1GiB, I noticed there is a the following is a curious reversal.
```
capacity           score
0                  0
1Byte ~ 1GiB       1
1GiB ~ 2GiB        0
2GiB ~             1~
(default divisor : 1)
```

I have two ideas to solve this problem, and I think idea 1 is better. what do you think?

1.  Return 1 as a score when capacity != 0 && converted < 1.
2.  Return 0 as a score when capacity < 1GiB.

jakobmoellerdev: > 1. Return 1 as a score when capacity != 0 && converted < 1.

do you mean converted = size >> 30 ?

Because I think this first case is what happens right? 

When we return 0 as score when capacity <1GiB then we have an issue with scoring nodes because nodes without capacity are on the same score right?

jakobmoellerdev: I do think your issue is valid though, basically as long as capacity > 0, score must be at least 1.

jakobmoellerdev: The only way I could think of to solve this would be to add special cases to all statements that could return 0 if capacity > 0, also in the after the calculating log2

jakobmoellerdev: @llamerada-jp @toshipp I think this could solve the issue. I had to introduce one more edge case as sometimes the logarithm can also result in score 0, so i just bounded it there as well. This also makes it safe against any divisor value
```go
func capacityToScore(capacity uint64, divisor float64) int {
	gb := capacity >> 30

	// Avoid logarithm of zero, which diverges to negative infinity.
	if gb == 0 {
		// If there is a non-nil capacity but we dont have at least one gigabyte, we score it with one.
		// This is because the capacityToScore precision is at the gigabyte level.
		// TODO: introduce another scheduling algorithm for byte-level precision.
		if capacity > 0 {
			return 1
		}

		return 0
	}

	converted := int(math.Log2(float64(gb) / divisor))
	score := converted
	switch {
	case score < 0:
		score = 0
	case score > 10:
		score = 10
	}
	
	// Also return score 1 here if we retrieve a score of 0 but we have a non-0 capacity.
	if capacity > 0 && score == 0 {
		return 1
	}
	
	return score
}
```

llamerada-jp: When capacity == 0, gb becomes 0 and is returned before the switch statement.
So will the switch statement be like below?
```go
	converted := int(math.Log2(float64(gb) / divisor))
	switch {
	case converted < 1:
		return 1
	case converted > 10:
		return 10
	default:
		return converted
	}
```

lvmd/command/lvm.go
```diff
@@ -504,22 +506,23 @@ func (l *LogicalVolume) Snapshot(name string, cowSize uint64, tags []string, thi
...
-					gbSize = cowMax
+				// If COW size is not specified, use 20% of the volume size or COW max when it is higher than that.
+				size = l.size * 2 / 10
+				if size > cowMax {
+					size = cowMax
 				}
 			}
-			if gbSize < cowMin {
-				gbSize = cowMin
+			if cowSize < cowMin {
```
toshipp, line 519: I doubt that it should be `size` instead of `cowSize`. Could you check it?

jakobmoellerdev: You are right. However after further review, I am questioning why this code even exists. What is the reason to have a minimum Copy-on-Write Minimum/Maximum? Shouldn't this be irrelevant and we can simply take over the size of the request?

lvmd/command/lvm.go
peng225: > We actually never call this code in TopoLVM because we officially dont support thick snapshots. 

Oh... you are right.

> but maybe we should simply remove it?

IMO, we should not remove it in this PR.

lvmd/command/lvm.go
toshipp: The code existed before I joined to maintain it, so I don't know the exact reason. I suspect it is some sort of safeguard to prevent snapshots from becoming unwritable when the CoW space is exhausted.

toshipp: As you know, the code is not called if the volume is thick, we can change the function to only support thin volumes and return an error if the volume is thick.

scheduler/prioritize_test.go
```diff
@@ -16,22 +17,24 @@ func TestCapacityToScore(t *testing.T) {
 		expect  int
 	}{
 		{0, 1, 0},
-		{1, 1, 0},
+		{1, 1, 1}, // even one byte will lead to a score of at least 1
 		{128 << 30, 1, 7},
 		{128 << 30, 2, 6},
 		{128 << 30, 0.5, 8},
 		{^uint64(0), 1, 10},
```
peng225, line 24: Could you add a test for the case where `converted` is less than 1?
https://github.com/topolvm/topolvm/pull/766/files#diff-ffe695214ecf8679e2ecd8e76e947ced21ad9539fe64f32f116aeb181f7064cfR32-R33

driver/controller_test.go
peng225, line 1: I want a test case where both `requestBytes` and `limitBytes` is 0.
(In this case, the result should be 1Gi.)

lvmd/command/lvm.go
jakobmoellerdev: Agreed. Changed the method signature to 

```go
// ThinSnapshot takes a thin snapshot of a volume.
// The volume must be thinly-provisioned.
// snapshots can be created unconditionally.
func (l *LogicalVolume) ThinSnapshot(name string, tags []string) (*LogicalVolume, error)
```

We already error out if the snapshot is not thin at https://github.com/topolvm/topolvm/blob/0e37bda0e8f28c02ac782889047b7c8ede8d47c0/lvmd/lvservice.go#L231-L233 so there is no need for passing thin flag into this method.

scheduler/prioritize_test.go
jakobmoellerdev: I added the case
```
{1 << 30, 1, 1}, // converted < 1 but still gets a score of 1 because of the minimum
```
for this. Behaves as expected now.

driver/controller_test.go
jakobmoellerdev: added the test case

```go
        v, err = convertRequestCapacityBytes(0, 0)
	if err != nil {
		t.Error("should not be error")
	}
	if v != 1<<30 {
		t.Errorf("should be 1073741825 in byte precision: %d", v)
	}
```

scheduler/prioritize.go
jakobmoellerdev: Adjusted accordingly and all tests still pass.

lvmd/command/lvm.go
toshipp: The naming and the removal of the thin flag are reasonable. However, the LogicalVolume type represents both thin and thick, the ThinSnapshot function can be called even if it is thin. It is better to return an error in this case, IMO.

jakobmoellerdev: How should we determine if we are a thin snapshot if we do not pass the thin flag? I am not aware how we could deduce that from the given volume. Should we check for attributes first?

jakobmoellerdev: Removed everything that was executed in this method when `l.pool == nil` because for thin volumes we always have a pool. Also added an error in case we are running without a thin volume

lvmd/lvservice.go
peng225: Thank you for the explanation. That sounds good.

lvmd/command/lvm.go
```diff
@@ -16,8 +15,10 @@ const (
 	nsenter  = "/usr/bin/nsenter"
 	lvm      = "/sbin/lvm"
 	blockdev = "/sbin/blockdev"
-	cowMin   = 50
-	cowMax   = 300
+	cowMinGi = 50
+	cowMin   = cowMinGi << 30
+	cowMaxGi = 300
+	cowMax   = cowMaxGi << 30
```
peng225, line 21: These variables are no longer used.

lvmd/command/lvm.go
```diff
@@ -489,58 +490,12 @@ func (l *LogicalVolume) Tags() []string {
...
-		// without this, wrong data may read from the snapshot.
-		if err := wrapExecCommand(blockdev, "--flushbufs", snapLV.path).Run(); err != nil {
-			return nil, err
-		}
-		return snapLV, nil
+// ThinSnapshot takes a thin snapshot of a volume.
+// The volume must be thinly-provisioned.
+// snapshots can be created unconditionally.
+func (l *LogicalVolume) ThinSnapshot(name string, tags []string) (*LogicalVolume, error) {
+	if !l.IsThin() { // pool == nil
```
peng225, line 497: > // pool == nil

Please remove the comment if it is not necessary.

lvmd/command/lvm.go
toshipp: LGTM.

toshipp (commented):
LGTM.

peng225:
LGTM

jakobmoellerdev:
@toshipp @peng225 just squashed the commits, would be great if you could give an approval for the merge now

peng225 (approved):

peng225:
Since [@toshipp says LGTM](https://github.com/topolvm/topolvm/pull/766#pullrequestreview-1696773492), I will merge this PR.

PR #769, opened 2023-10-18
docs: Proposal for running LVMD from within topolvm-node instead of using gRPC Services

jakobmoellerdev:
To run TopoLVM on Edge currently requires quite a lot of memory for a minimal installation. One of the main
reasons for that is due to the lvmd container running as a separate DaemonSet from the topolvm-node DaemonSet.
The idea is to combine these two into a single DaemonSet to reduce the memory footprint as they are usually run
within the same pod anyway when lvmd is not running as a systemd service.

### DIFF
--- a/docs/proposals/lvmd-within-topolvm-node.md
+++ b/docs/proposals/lvmd-within-topolvm-node.md
@@ -1,0 +1,111 @@
+# Starting LVMD from within topolvm-node instead of using gRPC
+
+## Motivation
+
+To run TopoLVM on Edge currently requires quite a lot of memory for a minimal installation. One of the main
+reasons for that is due to the lvmd container running as a separate DaemonSet from the topolvm-node DaemonSet.
+The idea is to combine these two into a single DaemonSet to reduce the memory footprint as they are usually run
+within the same pod anyway when lvmd is not running as a systemd service.
+
+### Goals
+
+- Unify the topolvm-node and lvmd containers into a single container and not have to communicate via gRPC.
+- Allow lvmd to run as a systemd service or as a DaemonSet like before.
+- Do not break existing installations.
+- Reduce memory footprint.
+
+## Proposal
+
+TopoLVM is a storage plugin based on [CSI](https://github.com/container-storage-interface/spec/).
+Therefore, the architecture basically follows the one described in
+https://kubernetes-csi.github.io/docs/ .
+
+To manage LVM, `lvmd` should be run as a system service of the node OS.
+It provides gRPC services via UNIX domain socket to create/update/delete
+LVM logical volumes and watch a volume group status.
+
+`topolvm-node` implements CSI node services as well as miscellaneous control
+on each Node.  It communicates with `lvmd` to watch changes in free space
+of a volume group and exports the information by annotating Kubernetes
+`Node` resource of the running node.  In the meantime, it adds a finalizer
+to the `Node` to clean up PersistentVolumeClaims (PVC) bound on the node.
+
+Since both topolvm-node and lvmd need to run as DaemonSet, we can simply combine them and start them under a single binary.
+
+## Design Details
+
+### Allowing topolvm-node to start lvmd
+
+The main problem with running lvmd as a systemd service or DaemonSet is that it is not possible to start it from within the topolvm-node container.
+We can fix this by allowing topolvm-node to start lvmd by compiling a service into its startup command.
+This can be enabled via flag and then use a wrapper to call lvmd server commands directly from the client.
+
+### Flow of Operations for lvmd running within topolvm-node
+
+1. topolvm-node starts and checks if lvmd should be started as well via `--embed-lvmd` flag.
+2. If `--embed-lvmd` is set, it will not initialize gRPC clients via Connection but call a dedicated wrapper instead.
+3. The wrapper generates a `proto.LVServiceClient` and `proto.VGServiceClient` by calling `lvmd.NewEmbeddedServiceClients`.
+   The `lvmd.NewEmbeddedServiceClients` function will create the Services and bind the client directly to them:
+   ```go
+   // NewEmbeddedServiceClients creates clients locally calling instead of using gRPC.
+   func NewEmbeddedServiceClients(ctx context.Context, dcmapper *DeviceClassManager, ocmapper *LvcreateOptionClassManager) (
+       proto.LVServiceClient,
+       proto.VGServiceClient,
+   ) {
+       vgServiceServerInstance, notifier := NewVGService(dcmapper)
+       lvServiceServerInstance := NewLVService(dcmapper, ocmapper, notifier)
+	   // use the local wrapper to redirect client calls to the service instances.
+       caller := &localCaller{
+           lvServiceServer: lvServiceServerInstance,
+           vgServiceServer: vgServiceServerInstance.(*vgService),
+       }
+	   // automatically start a server watch so that watch clients get notified until context is cancelled.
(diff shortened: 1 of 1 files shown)


docs/proposals/lvmd-within-topolvm-node.md
```diff
@@ -0,0 +1,312 @@
...
+		}
+		dcm := lvmd.NewDeviceClassManager(config.lvmd.DeviceClasses)
+		ocm := lvmd.NewLvcreateOptionClassManager(config.lvmd.LvcreateOptionClasses)
+		_, lvclnt, _, vgclnt = lvmd.NewLocal(dcm, ocm)
+	} else {
+		dialer := &net.Dialer{}
+		dialFunc := func(ctx context.Context, a string) (net.Conn, error) {
+			return dialer.DialContext(ctx, "unix", a)
+		}
+		conn, err := grpc.Dial(config.lvmdSocket, grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithContextDialer(dialFunc))
```
peng225, line 206: `conn` is not used in the following lines.

In the latest code, it is passed to `NewLogicalVolumeReconciler`.
https://github.com/topolvm/topolvm/blob/52dbdcce42eba0509a4a0e94eaa04e9571f74cd9/pkg/topolvm-node/cmd/run.go#L76

You should change how to initialize `lvcontroller` depending on the value of `config.llvmd`.

docs/proposals/lvmd-within-topolvm-node.md
```diff
@@ -0,0 +1,312 @@
...
+}
+
+```
+
+This will allow starting and using lvmd without grpc communication and another container.
+
+## Caveats
+
+- LVMD can no longer be deployed and managed separately from topolvm-node if started together
+- We will need to decide if we want to offer HELM chart configurations for this style of configuration
```
peng225, line 306: IMO, it is preferable if you could update Helm Chart for the run-lvmd-local mode.
If updating the Helm Chart is not very important for you, it is also OK to work on it later in another PR.

@satoru-takeuchi 
I want to hear your opinion.

peng225 (commented):
The idea seems good to me.
I left some comments about the implementation.

docs/proposals/lvmd-within-topolvm-node.md
```diff
@@ -0,0 +1,312 @@
...
+	proto.VGServiceServer,
+	proto.VGServiceClient,
+) {
+	vgServiceServerInstance, notifier := NewVGService(dcmapper)
+	lvServiceServerInstance := NewLVService(dcmapper, ocmapper, notifier)
+	caller := &localCaller{
+		lvServiceServer: lvServiceServerInstance,
+		vgServiceServer: vgServiceServerInstance,
+	}
+	return caller.lvServiceServer, caller.lvServiceClient, caller.vgServiceServer, caller.vgServiceClient
```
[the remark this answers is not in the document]
jakobmoellerdev: localCaller is just a struct used to pass the client calls directly to the server calls as we do not have the gRPC layer in between. And that needs to implement the proto interfaces to be passable to lv/vg service

docs/proposals/lvmd-within-topolvm-node.md
```diff
@@ -0,0 +1,312 @@
...
+// NewLocal creates a locally calling LVMD
+func NewLocal(dcmapper *DeviceClassManager, ocmapper *LvcreateOptionClassManager) (
+	proto.LVServiceServer,
+	proto.LVServiceClient,
+	proto.VGServiceServer,
+	proto.VGServiceClient,
+) {
+	vgServiceServerInstance, notifier := NewVGService(dcmapper)
+	lvServiceServerInstance := NewLVService(dcmapper, ocmapper, notifier)
+	caller := &localCaller{
```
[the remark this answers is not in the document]
jakobmoellerdev: Nope, still wasn't implemented. unfortunately i also had to implement the vgservice watch stream with an unbounded channel so that the notification of watch clients works as expected.

jakobmoellerdev:
no idea why build is failing, just a .md in the pr. Any ideas?

docs/proposals/lvmd-within-topolvm-node.md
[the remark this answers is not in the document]
jakobmoellerdev: I changed it so that it doesnt return the servers anymore (those were not used anyhow). And just returns caller as the client.

docs/proposals/lvmd-within-topolvm-node.md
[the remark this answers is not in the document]
jakobmoellerdev: Btw I was now able to run the code as shown from one container and could successfully create PVCs from that. I think this is looking promising.

peng225:
> no idea why build is failing, just a .md in the pr. Any ideas? (also checked locally and cannot reproduce)

Could you rebase your branch onto the `main` branch?

docs/proposals/lvmd-within-topolvm-node.md
satoru-takeuchi: I want to update the Helm Chart too.

docs/proposals/lvmd-within-topolvm-node.md
```diff
@@ -0,0 +1,404 @@
...
+	return l.lvServiceServer.CreateLVSnapshot(ctx, in)
+}
+
+func (l *localCaller) GetLVList(ctx context.Context, in *proto.GetLVListRequest, _ ...grpc.CallOption) (*proto.GetLVListResponse, error) {
+	return l.vgServiceServer.GetLVList(ctx, in)
+}
+
+func (l *localCaller) GetFreeBytes(ctx context.Context, in *proto.GetFreeBytesRequest, _ ...grpc.CallOption) (*proto.GetFreeBytesResponse, error) {
+	return l.vgServiceServer.GetFreeBytes(ctx, in)
+}
```
satoru-takeuchi, line 171: How about replacing code with simple flows as in the following proposal?

https://github.com/topolvm/topolvm/blob/main/docs/proposals/thin-snapshots-restore.md#flow-of-operations-for-snapshot-creation

I believe it can reduce the review cost since we can focus on the design rather than the detail which will be done in code review.

docs/proposals/lvmd-within-topolvm-node.md
```diff
@@ -0,0 +1,111 @@
...
+
+### Allowing topolvm-node to start lvmd
+
+The main problem with running lvmd as a systemd service or DaemonSet is that it is not possible to start it from within the topolvm-node container.
+We can fix this by allowing topolvm-node to start lvmd by compiling a service into its startup command.
+This can be enabled via flag and then use a wrapper to call lvmd server commands directly from the client.
+
+### Flow of Operations for lvmd running within topolvm-node
+
+1. topolvm-node starts and checks if lvmd should be started as well via `--llvmd` flag.
```
satoru-takeuchi, line 45: How about changing the option name to embed-lvmd? It's for the following two reasons.

- llvmd would be misunderstood as something for LLVM.
- Embed is more proper word than local in this case. It's because lvmd runs on local node in both the existing implementation and in the new mode.

LocalServiceClients will also be changed to EmbeddedServiceClients.

After reflecting this change request, I'll approve this PR.

satoru-takeuchi (changes requested):

docs/proposals/lvmd-within-topolvm-node.md
jakobmoellerdev: Great proposal for naming, adjusted accordingly and squashed the PR to prepare it for merge

satoru-takeuchi (approved):

peng225 (approved):

PR #771, opened 2023-10-25
feat: Starting LVMD from within topolvm-node instead of using gRPC

jakobmoellerdev:
To run TopoLVM on Edge currently requires quite a lot of memory for a minimal installation. One of the main reasons for that is due to the lvmd container running as a separate DaemonSet from the topolvm-node DaemonSet. The idea is to combine these two into a single DaemonSet to reduce the memory footprint as they are usually run within the same pod anyway when lvmd is not running as a systemd service.

For more details see the [dedicated enhancement proposal](https://github.com/topolvm/topolvm/blob/main/docs/proposals/lvmd-within-topolvm-node.md#starting-lvmd-from-within-topolvm-node-instead-of-using-grpc)

TODO:
- [x] Create E2E Test Pipeline in the Matrix
- [x] Validate HELM Chart changes are correct

### DIFF
--- a/e2e/Makefile
+++ b/e2e/Makefile
@@ -36,7 +36,7 @@ endif
 KIND_CONFIG="topolvm-cluster.yaml"
 MINIKUBE_FEATURE_GATES="ReadWriteOncePod=true"
 
-HELM_VALUES_FILE_LVMD := manifests/values/daemonset-lvmd-storage-capacity.yaml
+HELM_VALUES_FILE_LVMD ?= manifests/values/daemonset-lvmd-storage-capacity.yaml
 
 SCHEDULER_CONFIG := scheduler-config-$(TEST_SCHEDULER_MANIFEST).yaml
 
@@ -238,8 +238,8 @@ $(KUBECTL): | $(BINDIR)
 	$(CURL) -o $@ https://storage.googleapis.com/kubernetes-release/release/v$(KUBERNETES_VERSION)/bin/linux/amd64/kubectl
 	chmod a+x $@
 
-.PHONY: daemonset-lvmd/create-vg
-daemonset-lvmd/create-vg:
+.PHONY: incluster-lvmd/create-vg
+incluster-lvmd/create-vg:
 	mkdir -p build $(BACKING_STORE)
 	if [ $$(ls -1 $(BACKING_STORE)/backing_store_lvmd* 2>/dev/null | wc -l) -ne 0 ]; then $(MAKE) $(@D)/remove-vg; fi
 
@@ -274,8 +274,8 @@ daemonset-lvmd/create-vg:
 		$$($(SUDO) losetup -j $(BACKING_STORE)/backing_store_lvmd_8 | cut -d: -f1)
 
 
-.PHONY: daemonset-lvmd/remove-vg
-daemonset-lvmd/remove-vg:
+.PHONY: incluster-lvmd/remove-vg
+incluster-lvmd/remove-vg:
 	for i in $$(seq 8); do \
 		if [ "$${j}" -lt 7 ]; then \
 			$(SUDO) vgremove -ffy node-myvg$${i}; \
@@ -285,16 +285,16 @@ daemonset-lvmd/remove-vg:
 		rm -f $(BACKING_STORE)/backing_store_lvmd_$${i}; \
 	done
 
-.PHONY: daemonset-lvmd/setup-minikube
-daemonset-lvmd/setup-minikube:
+.PHONY: incluster-lvmd/setup-minikube
+incluster-lvmd/setup-minikube:
 	mkdir -p $(BINDIR)
 	$(SUDO) apt-get update
 	DEBIAN_FRONTEND=noninteractive $(SUDO) apt-get install -y --no-install-recommends conntrack
 	$(CURL) -o $(BINDIR)/minikube https://github.com/kubernetes/minikube/releases/download/$(MINIKUBE_VERSION)/minikube-linux-amd64
 	chmod a+x $(BINDIR)/minikube
 
-.PHONY: daemonset-lvmd/launch-minikube
-daemonset-lvmd/launch-minikube:
+.PHONY: incluster-lvmd/launch-minikube
+incluster-lvmd/launch-minikube:
 	@if [ "$(STORAGE_CAPACITY)" = false ]; then echo "Only storage capacity mode is supported."; exit 1; fi
 	$(SUDO) -E $(BINDIR)/minikube start \
 		--vm-driver=none \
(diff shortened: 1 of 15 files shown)


lvmd/local.go
peng225, line 1: Because the option name is `embed-lvmd`, the file name should be something like "embed.go" instead of using the word "local."

peng225 (commented):
Comments so far.
I will continue my review next week.

.github/workflows/e2e-k8s-embedded-lvmd.yaml
```diff
@@ -0,0 +1,69 @@
...
+      - "**/*.md"
+      - "CODEOWNERS"
+  push:
+    paths-ignore:
+      - "**/*.md"
+      - "CODEOWNERS"
+    branches:
+      - "main"
+jobs:
+  e2e-k8s-daemonset-lvmd:
```
cupnes, line 14:
```suggestion
  e2e-k8s-embedded-lvmd:
```
Forgot to edit?

cupnes (changes requested):
I commented the following. But my review is not over yet, and I will continue to review it.

lvmd/embed.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	caller.vgWatch = &localWatch{ctx: ctx, watch: make(chan any)}
+	go caller.vgServiceServer.Watch(&proto.Empty{}, caller.vgWatch)
+	return caller, caller
+}
+
+// localCaller is a struct holding indirections to the local lvmd server.
+// It implements both the LVServiceClient and VGServiceClient interfaces.
+// It also includes a watch for redirecting the vg watch to the local caller.
+type localCaller struct {
```
peng225, line 33: I prefer to rename this struct's name to something like `embeddedServiceClient`.

lvmd/embed.go
```diff
@@ -0,0 +1,128 @@
...
+	lvServiceServer proto.LVServiceServer
+	vgServiceServer proto.VGServiceServer
+	vgWatch         *localWatch
+}
+
+// localWatch is a local implementation of the WatchClient and WatchServer that is used by vgService.
+// it uses an unbound blocking channel to send and receive messages.
+type localWatch struct {
```
peng225, line 41: I wonder if it is reasonable for `localWatch` to implement both `WatchClient`(= `proto.VGService_WatchClient`) and `WatchServer`(= `proto.VGService_WatchServer`).

`VGService_WatchClient` and `VGService_WatchServer` have common functions to be implemented (e.g. `Context`, `SendMsg`, `RecvMsg`). Though the current implementations of both interfaces happen to coincide, I believe they are semantically different.

How do you think about splitting`localWatch` into `embeddedWatchClient` and `embeddedWatchServer`?

peng225 (commented):
I posted additional comments, but I will need more time to finish my review.

cupnes:
I have confirmed that my point have been taken into account. The rest was left to @satoru-takeuchi and @peng225.

lvmd/embed.go
jakobmoellerdev: Usually I would agree, however this implementation is a pretty special case imo. Since the watchClient does nothing else then redirecting all calls into a channel that the watchServer has to react to, both share one unidirectional channel and have to implement Send and Receive methods. If I were to split the implementations, I would have to duplicate the code just to have different structs and share the channel pointer between both of them because they both need access to it. 

That would result in a lot more boilerplate code that doesn't really have any effect except for duplication since Send/Receive methods behave exactly the same.

I collected all of the methods I would have to duplicate below
```go
type SendReceiver interface {
	Send(m *proto.WatchResponse) error
	Recv() (*proto.WatchResponse, error)
	SendMsg(m any) error
	RecvMsg(m any) error
	Context() context.Context
}
```
if we split, then I would have to implement these twice or introduce a shared struct again that kind of defeats the purpose of what you are asking.

If you still think its better for style, I can adopt it, but I would like to keep it as is for simplicity

jakobmoellerdev: I added some additional method comments to explain what is stubbed and why so that it is easier to understand the codebase. Also cleaned up the interfaces a bit. I hope that is acceptable for you. If not feel free to nit again and I will adjust.

e2e/Makefile
```diff
@@ -337,3 +338,40 @@ daemonset-lvmd/test: topolvm.img $(GINKGO) $(HELM) $(KUBECTL)
...
+embedded-lvmd/setup-minikube: daemonset-lvmd/setup-minikube
+
+.PHONY: embedded-lvmd/launch-minikube
+embedded-lvmd/launch-minikube: daemonset-lvmd/launch-minikube
+
+.PHONY: embedded-lvmd/delete-minikube
+embedded-lvmd/delete-minikube: daemonset-lvmd/delete-minikube
+
+.PHONY: embedded-lvmd/test
+embedded-lvmd/test: topolvm.img $(GINKGO) $(HELM) $(KUBECTL)
```
peng225, line 355: The only difference between `daemonset-lvmd/XXX` and `embedded-lvmd/XXX` is the `values.yaml` file path, so I want to remove the duplication.

My idea is as follows:

- Rename `daemonset-lvmd/foo` target to something like `incluster-lvmd/foo` so that it can be used for both daemonset lvmd and embedded lvmd installation.
- Remove `embedded-lvmd/foo`.
- Remove `HELM_VALUES_FILE_LVMD_EMBEDDED` variable.
- In github workflow, change the value of `HELM_VALUES_FILE_LVMD` via command line option of `make` command.

Then, when we want to run e2e test for daemonset lvmd, we can do it as follows.
```
make -C e2e incluster-lvmd/create-vg
make -C e2e incluster-lvmd/setup-minikube
make -C e2e incluster-lvmd/launch-minikube
make -C e2e incluster-lvmd/test
```

When we want to run e2e test for embedded lvmd, we can do it as follows.
```
make -C e2e incluster-lvmd/create-vg  # Same as daemonset lvmd
make -C e2e incluster-lvmd/setup-minikube  # Same as daemonset lvmd
make -C e2e incluster-lvmd/launch-minikube  # Same as daemonset lvmd
make -C e2e incluster-lvmd/test HELM_VALUES_FILE_LVMD = manifests/values/embedded-lvmd-storage-capacity.yaml
```

@satoru-takeuchi 
I would also like to hear from you about my idea. What do you think?

e2e/manifests/values/embedded-lvmd-storage-capacity.yaml
```diff
@@ -0,0 +1,83 @@
...
+  - name: topolvm-provisioner-thin
+    storageClass:
+      fsType: xfs
+      isDefaultClass: false
+      volumeBindingMode: WaitForFirstConsumer
+      allowVolumeExpansion: true
+      additionalParameters:
+        topolvm.io/device-class: "thin"
+
+podSecurityPolicy:
```
peng225, line 75: This field no longer exists.

e2e/Makefile
jakobmoellerdev: Good idea. I will use a github matrix to extend the tests via env variable so we can also get rid of the additional github actions file. I think this is definetly a big improvement.

jakobmoellerdev: Changed it so that the file is changed based on the matrix, also took over your suggestion for `incluster-lvmd`

jakobmoellerdev:
I think lint job is flaky, only changed one related line

peng225:
LGTM

satoru-takeuchi:
@jakobmoellerdev I'm still reviewing this PR. Please wait for one more day.

satoru-takeuchi:
The failure of `Lint Charts` probably comes from `cosign-installer` or `cosign`. 

The dummy PR to reproduce this failure.
https://github.com/topolvm/topolvm/pull/775

In addition to this review, I'll try to fix this test problem.

satoru-takeuchi:
> Please rebase your PR after #775 has been merged.

Merged. Please rebase this PR to the latest main branch.

jakobmoellerdev:
@satoru-takeuchi lint is green again, thanks for fixing that. Squashed the commits and now should be ready to go

satoru-takeuchi:
@jakobmoellerdev My review is done and LGTM. I'll merge this PR after confirming the result of all tests.

peng225 (approved):

satoru-takeuchi (approved):

Issue #773, opened 2023-10-26
Upgrading the logging library of lvmd


jakobmoellerdev:
**What should the feature do:**

Allow configuring lvmd logging via klog or controller-runtime.

**What is use case behind this feature:**

Currently the logging library of lvmd is based on the archived https://github.com/cybozu-go/log library. Nowadays all other parts of topolvm are either using klog or controller-runtime logr implementations.

It is also apparent why. The majority of features found in klog / controller-runtime are not supported in lvmd and unifying and scraping logs is hard to accomplish.

peng225:
Though I don't know why `cybozu-go/log` is used, I guess that is for historical reasons.
As you know, that has been already archived, so I agree to replace it with other logging libraries.

jakobmoellerdev:
Awesome! Will this be tackled by you @cupnes or should I take care of that?

satoru-takeuchi:
I'll ask @cupnes tomorrow. Today, he is on PTO.

cupnes:
@jakobmoellerdev May I leave this to you? I will review it.

PR #779, opened 2023-11-08
chore: remove cybozu-log and replace with zap / controller-runtime log

jakobmoellerdev:
This fixes https://github.com/topolvm/topolvm/issues/773 by replacing every log statement in lvmd with a controller runtime logger coming from the command or controller context which aligns it with the other controllers. Effectively makes github.com/cybozu-go/log an indirect due to github.com/cybozu-go/well. Did not remove that for scope reasons

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,6 @@ go 1.20
 
 require (
 	github.com/container-storage-interface/spec v1.6.0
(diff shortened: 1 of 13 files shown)


cupnes (approved):
LGTM

PR #780, opened 2023-11-08
fix condition preventing multiple default device-classes

elproko:
When there are multiple volume groups and device-classes and more than 1 node then "!=1" condition is triggering false error on secondary and next nodes.

### DIFF
--- a/lvmd/device_class_manager.go
+++ b/lvmd/device_class_manager.go
@@ -131,7 +131,7 @@ func ValidateDeviceClasses(deviceClasses []*DeviceClass) error {
 			return fmt.Errorf("stripe-size format is \"Size[k|UNIT]\": %s", dc.Name)
 		}
 	}
-	if countDefault != 1 {
+	if countDefault > 1 {
 		return errors.New("should have only one default device-class")
 	}
 	return nil


toshipp:
I can't understand your case. Could you tell me more details? Example configurations could help me.

elproko:
I am using LVMCluster on 3-node Openshift. I have 3-volume groups: each vg for each node.

config below:

apiVersion: lvm.topolvm.io/v1alpha1
kind: LVMCluster
metadata:
  finalizers:
    - lvmcluster.topolvm.io
  name: lvmcluster
  namespace: openshift-storage
spec:
  storage:
    deviceClasses:
      - default: true
        deviceSelector:
          paths:
            - /dev/disk/by-id/scsi-3300xxxxabcde
        fstype: xfs
        name: vg1
        nodeSelector:
          nodeSelectorTerms:
            - matchExpressions:
                - key: topology.topolvm.io/node
                  operator: In
                  values:
                    - n1.openshift.fqdn
        thinPoolConfig:
          name: thin-pool-1
          overprovisionRatio: 10
          sizePercent: 90
      - default: false
        deviceSelector:
          paths:
            - /dev/disk/by-id/ata-TOSHIBA_MQ0abcdef
        fstype: xfs
        name: vg2
        nodeSelector:
          nodeSelectorTerms:
            - matchExpressions:
                - key: topology.topolvm.io/node
                  operator: In
                  values:
                    - n2.openshift.fqdn
        thinPoolConfig:
          name: thin-pool-1
          overprovisionRatio: 10
          sizePercent: 90
      - default: false
        deviceSelector:
          paths:
            - /dev/disk/by-id/ata-TOSHIBA_MQ0xxxyaT
        fstype: xfs
        name: vg3
        nodeSelector:
          nodeSelectorTerms:
            - matchExpressions:
                - key: topology.topolvm.io/node
                  operator: In
                  values:
                    - n3.openshift.fqdn
        thinPoolConfig:
          name: thin-pool-1
          overprovisionRatio: 10
          sizePercent: 90
---------

When cluster is built - on nodes n2 & n3 topolvm-node pods are failing with lvmd  "should have only one default device-class" error.
When I change lvmd.yaml configs on those nodes setting their volumegroups "default"  to "true" - then topolvm-node pods are starting correctly.
Since then I can provision PVs on nodes n2 & n3 when I specify appropriate storage-class/device-class pointing to volume groups vg2 & vg3.

regards
Przemek Dmochowski

toshipp:
Hi @elproko 
The default device-class is used when no specific device-classes are specified in the StorageClasses. As such, we cannot permit multiple default device-classes. Consequently, we are unable to accept your PR.

By the way, I presume you are using the [lvm-operator](https://github.com/openshift/lvm-operator), which is not a part of the TopoLVM project. Could you kindly report your issue directly to their project? Thank you.

jakobmoellerdev:
Hi @elproko dev from the lvm-operator here, please close this issue and move it to OCPBUGS under the LVMS component. 

That being said: For LVMCluster with 3 deviceClasses on 3 nodes you currently need to specify the paths as optionalPaths. Also one of the nodes has to act as default provisioner.

For more information see the [docs](https://docs.openshift.com/container-platform/4.14/storage/persistent_storage/persistent_storage_local/persistent-storage-using-lvms.html#lvms-creating-lvms-cluster_logical-volume-manager-storage):

> To control or restrict the volume group to your preferred devices, you can manually specify the local paths of the devices in the deviceSelector section of the LVMCluster YAML. The paths section refers to devices the LVMCluster adds, which means those paths must exist. The optionalPaths section refers to devices the LVMCluster might add. You must specify at least one of paths or optionalPaths when specifying the deviceSelector section. If you specify paths, it is not mandatory to specify optionalPaths. If you specify optionalPaths, it is not mandatory to specify paths but at least one optional path must be present on the node. If you do not specify any paths, it will add all unused devices on the node.

That being said, multi-node LVMS on OpenShift is not supported in 4.14 or below officially.

@toshipp we should really consider how we deal with us sharing the same API group as you guys. I don't know when that was decided but it regularly leads to you having to deal with support requests from LVMS, which is not nice. I will bring it up with the team. Please feel free to close the issue.

toshipp:
@jakobmoellerdev 
Thank you for your comment.

elproko:
@toshipp I think we have misunderstanding here.
My goal is: to use topolvm on more than single node with only **single** default device-class.
My previous explanation was only to show why lvmd is complaining with error when in reality there was **no default class** on the specific node. Setting "default" on nodes n2 & n3 was only a workaround against wrong constructed condition.

Again - currently error is triggering in 2 cases:
1. when there are more defauld device classes than one.
2. when there is no default device class on specific node.

By this PR I wanted to cancel second triggering case. 

Kindly please consider this case again.

elproko:
> Hi @elproko The default device-class is used when no specific device-classes are specified in the StorageClasses. As such, we cannot permit multiple default device-classes. Consequently, we are unable to accept your PR.
> 
> By the way, I presume you are using the [lvm-operator](https://github.com/openshift/lvm-operator), which is not a part of the TopoLVM project. Could you kindly report your issue directly to their project? Thank you.

I created bug as You suggested - thank You.

Again - to clarify: this PR was not to use multiple default device-classes but to fix wrong condition which triggers false-positive when there is no default device-class on specific node.

jakobmoellerdev:
Hi @elproko please find my answer on how you can solve your original issue (for lvm operator) in https://github.com/openshift/lvm-operator/pull/483#issuecomment-1806846303

That being said, I do understand your concern, and I think the misunderstanding is that lvmd.conf should be aligned on all nodes. If you do not do that (and configure lvmd differently on multiple nodes) you will have issues like above. From what I understand you attempted to configure lvmd.conf on multiple nodes and only one of the nodes contained a default deviceClass while the other node didn't. This is not intended. Instead lvmd should be deployed with a single config that is the same on every node.

@toshipp I think the use case of @elproko is valid. Should we think about allowing differing lvmd.conf files for individual nodes? Then the validation change that is suggested actually makes sense.

toshipp:
@elproko 
I misunderstood your case.
>My goal is: to use topolvm on more than single node with only single default device-class.

Did you mean that your issue is that you want to enable the default device-class on only a single node?
If so, this breaks our assumption that the default device-class is always used, which may cause another issue.
For example, this code returns a pointer to the default device class structure. When your change is applied, this will return nil, then the process will crash due to the nil dereference.
https://github.com/topolvm/topolvm/blob/main/lvmd/device_class_manager.go#L180

I will ask our team if we can accept relaxing the current limit.

elproko:
> @elproko I misunderstood your case.
> 
> > My goal is: to use topolvm on more than single node with only single default device-class.
> 
> [...]
> I will ask our team if we can accept relaxing the current limit.

Thank you very much for reconsidering and going deeper into this.

> When your change is applied, this will return nil, then the process will crash due to the nil dereference. 

After all it seems I had luck with my dirty hack :)
I am glad because I hope it can be changed with best results.

regards
Przemek

toshipp:
Hi @elproko 
I discussed this with our team and concluded that allowing no default device classes was reasonable.
To merge this PR, we will solve two major problems I found.
1. [The DeviceClass function](https://github.com/topolvm/topolvm/blob/main/lvmd/device_class_manager.go#L178-L186) should return `ErrNotFound` because the callers dereference the returned pointer even if it is nil.
2. [The checkFunc](https://github.com/topolvm/topolvm/blob/main/pkg/topolvm-node/cmd/run.go#L161) should be changed not to depend on the default device class. This RPC call is intended to check the connection is open, it is better to add a ping RPC and use it instead, IMO.

elproko:
> Hi @elproko I discussed this with our team and concluded that allowing no default device classes was reasonable. To merge this PR, we will solve two major problems I found.
> 
> 1. [The DeviceClass function](https://github.com/topolvm/topolvm/blob/main/lvmd/device_class_manager.go#L178-L186) should return `ErrNotFound` because the callers dereference the returned pointer even if it is nil.
> 2. [The checkFunc](https://github.com/topolvm/topolvm/blob/main/pkg/topolvm-node/cmd/run.go#L161) should be changed not to depend on the default device class. This RPC call is intended to check the connection is open, it is better to add a ping RPC and use it instead, IMO.
> 
> Please check and update the PR.

Sorry for delayed answer.

I agree with you about those two code places to fix, but I am not able to do that myself. I hope you could do that.

toshipp:
@elproko 
Ok. I'll take over this PR. It might take some time due to our resources. Please be patient.

PR #781, opened 2023-11-09
Merge versions together into versions.mk

toshipp:
This project uses many tools and their versions are scattered in many places. It is a burden for us to update them, so we merge them into a single versoins.mk file.


e2e/Makefile
```diff
@@ -22,8 +20,6 @@ GINKGO := $(BINDIR)/ginkgo-$(GINKGO_VERSION)
 MINIKUBE_HOME = $(BINDIR)
 export MINIKUBE_HOME
 
-export TEST_KUBERNETES_VERSION=$(shell echo $(KUBERNETES_VERSION) | grep -o '^[0-9]*\.[0-9]*')
```
toshipp, line 25: This environment variable is not used as of b6c1cea5821504d776489aa5038afa16cd862c6e

Makefile
```diff
@@ -33,7 +22,10 @@ IMAGE_TAG ?= latest
 ORIGINAL_IMAGE_TAG ?=
 STRUCTURE_TEST_TARGET ?= normal
 
-ENVTEST_KUBERNETES_VERSION=1.27
+CONTROLLER_RUNTIME_VERSION := $(shell awk '/sigs\.k8s\.io\/controller-runtime/ {print substr($$2, 2)}' go.mod)
+CONTROLLER_TOOLS_VERSION := $(shell awk '/sigs\.k8s\.io\/controller-tools/ {print substr($$2, 2)}' go.mod)
+
+ENVTEST_KUBERNETES_VERSION := $(shell echo $(KUBERNETES_VERSION) | cut -d "." -f 1-2)
```
daichimukai, line 28: These (and `PROTOC_GEN_GO_VERSION`, ` PROTOC_GEN_DOC_VERSION`,  
`PROTOC_GEN_GO_GRPC_VERSION`) are not defined in `version.mk`? I understand these versions are special in the sense that these are actually defined in `go.mod` rather than `Makefile`, but I think it would be easy to understand if `versions.mk` contains all version numbers of our dependencies that is referenced in `Makefile`s, instead of only part of these.

toshipp: agreed. I'll revise it.

versions.mk
```diff
@@ -0,0 +1,38 @@
...
+KIND_VERSION := v0.20.0
+# It is set by CI using the environment variable, use conditional assignment.
+KUBERNETES_VERSION ?= 1.27.3
+MINIKUBE_VERSION := v1.31.2
+PROTOC_VERSION :=  24.4
+
+# Tools versions which are defined in go.mod
+CONTROLLER_RUNTIME_VERSION := $(shell awk '/sigs\.k8s\.io\/controller-runtime/ {print substr($$2, 2)}' go.mod)
+CONTROLLER_TOOLS_VERSION := $(shell awk '/sigs\.k8s\.io\/controller-tools/ {print substr($$2, 2)}' go.mod)
+GINKGO_VERSION := $(shell awk '/github.com\/onsi\/ginkgo\/v2/ {print $$2}' ../go.mod)
```
ushitora-anqou, line 18:
```suggestion
GINKGO_VERSION := $(shell awk '/github.com\/onsi\/ginkgo\/v2/ {print $$2}' go.mod)
```

daichimukai: Since the path to `go.mod` is specified by a relative path, whether this variable can be used as intended in the Makefile that includes `version.mk` depends on where the Makefile is located.

ushitora-anqou: I have found that both `go.mod` and `../go.mod` have problems:
- If we use `go.mod`, then `make -C e2e COMMAND` fails (with the following error message: `awk: fatal: cannot open file 'go.mod' for reading: No such file or directory`)
- If we use `../go.mod`, then `make COMMAND` fails (with the following error message: `awk: fatal: cannot open file '../go.mod' for reading: No such file or directory`)

toshipp: I consider the problem can be solved by this trick.
https://github.com/topolvm/topolvm/pull/781/files#diff-995146baf6475898321c6ef6dbdeef8b1dc91c6f4a6cdff87dc13645617feb06R3
I'll dig it more.

toshipp:
@daichimukai @ushitora-anqou 
I didn't know github trims force-push diff history when doing multiple force-push, so please check the whole changes.

Makefile
```diff
@@ -33,11 +22,7 @@ IMAGE_TAG ?= latest
 ORIGINAL_IMAGE_TAG ?=
 STRUCTURE_TEST_TARGET ?= normal
 
-ENVTEST_KUBERNETES_VERSION=1.27
-
-PROTOC_GEN_GO_VERSION := $(shell awk '/google.golang.org\/protobuf/ {print substr($$2, 2)}' go.mod)
-PROTOC_GEN_DOC_VERSION := $(shell awk '/github.com\/pseudomuto\/protoc-gen-doc/ {print substr($$2, 2)}' go.mod)
-PROTOC_GEN_GO_GRPC_VERSION := $(shell awk '/google.golang.org\/grpc\/cmd\/protoc-gen-go-grpc/ {print substr($$2, 2)}' go.mod)
+ENVTEST_KUBERNETES_VERSION := $(shell echo $(KUBERNETES_VERSION) | cut -d "." -f 1-2)
```
daichimukai, line 25: Should this not to be moved to `version.mk`?

toshipp: it's supposed to be.

daichimukai (approved):
LGTM

ushitora-anqou (approved):

PR #796, opened 2023-12-20
Reduce duplicated contents

toshipp:
There are many duplicated contents, but they can be reduced by helm functionality.


charts/topolvm/templates/storageclass.yaml
```diff
@@ -19,9 +19,11 @@ metadata:
     {{- include "topolvm.labels" . | nindent 4 }}
 provisioner: {{ include "topolvm.pluginName" . }}
 parameters:
-  "csi.storage.k8s.io/fstype": {{ $val.storageClass.fsType | quote }}
+  {{- with $val.storageClass.fsType }}
+  "csi.storage.k8s.io/fstype": {{ . | quote }}
+  {{- end }}
   {{- with $val.storageClass.additionalParameters }}
-  {{- toYaml . | nindent 2 }}
+  {{- tpl (toYaml .) $global | nindent 2 }}
```
peng225, line 26: I'm sorry, but I couldn't understand this line.
I know that you tried to use some directives by the following commit.
https://github.com/topolvm/topolvm/pull/796/commits/c69e9a9da40100e5ab2d966282f1a39f66be22ed

However, I still cannot understand the effect of this change. Could you explain a little bit more?

toshipp: Since helm does not render the template defined in `values.yaml`, we should use `tpl` function explicitly to render it.
Without this, [this](https://github.com/topolvm/topolvm/pull/796/files#diff-4493f81269a51f822dd87f34aaf404827c57b6eeb2c32265591cac872d782413R60) can not be rendered.

peng225: I got it. Thank you for the clarification.

peng225 (approved):

satoru-takeuchi (approved):

PR #798, opened 2023-12-26
Avoid using cybozu-go/well

ushitora-anqou:
TopoLVM is using [cybozu-go/well](https://github.com/cybozu-go/well), but it's now deprecated. This patch makes TopoLVM avoid using cybozu-go/well.


pkg/lvmd/cmd/root.go
```diff
@@ -97,31 +99,51 @@ func subMain(ctx context.Context) error {
...
-		<-ctx.Done()
+
+	var wg sync.WaitGroup
+	errCh := make(chan error, 1)
+	ctx, cancel := context.WithCancel(context.Background())
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		if err := grpcServer.Serve(lis); err != nil {
```
llamerada-jp, line 110: `well.Go` terminate all go routine when any go routine is finished. This implementation, the program will not stop even if grpcServer.Serve get terminated before any other.

pkg/lvmd/cmd/root.go
```diff
@@ -97,31 +99,51 @@ func subMain(ctx context.Context) error {
...
+		} else {
+			errCh <- nil
+		}
+	}()
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+
+		sigCh := make(chan os.Signal, 1)
```
llamerada-jp, line 121: `well.Go` does not seem to be handling the signal. Why do you implement this logic? Would you show me where the logic is written about signal.

pkg/topolvm-scheduler/cmd/access_log.go
```diff
@@ -0,0 +1,60 @@
...
+	w.size += n
+	return n, err
+}
+
+func (w *accessLogResponseWriter) WriteHeader(statusCode int) {
+	w.statusCode = statusCode
+	w.ResponseWriter.WriteHeader(statusCode)
+}
+
+func accessLogHandler(ctx context.Context, next http.Handler) http.Handler {
```
llamerada-jp, line 29: Would you add some tests for accessLogHandler?

llamerada-jp (changes requested):

pkg/lvmd/cmd/root.go
ushitora-anqou: Actually, cybozu-go/well handles signals [here](https://github.com/cybozu-go/well/blob/8cbd22dd58421e97e3cd8a9ecc0ce65000dd682f/signal.go#L28), and `well.Wait` returns `well.errSignaled` when `SIGINT` or `SIGTERM` is raised. The original code of TopoLVM checks if a signal is raised by calling `well.IsSignaled` ([here](https://github.com/topolvm/topolvm/blob/c58ec847f7378509733c0d2b5570149ed625bd59/pkg/lvmd/cmd/root.go#L121)), so it will be necessary to handle signals also in this patch.

pkg/lvmd/cmd/root.go
ushitora-anqou: That's right. I'll fix my code.

pkg/topolvm-scheduler/cmd/access_log.go
ushitora-anqou: I wrote `TestAccessLogHandler`, the test for `accessLogHandler`. It's similar to [cybozu-go/well's tests](https://github.com/cybozu-go/well/blob/8cbd22dd58421e97e3cd8a9ecc0ce65000dd682f/http_test.go#L221), but there are some differences:

- I omitted tests for `logged_at` because encoding of time is done automatically by zap.
- I omitted tests for `severity` because logr doesn't support warning level logging and `accessLogHandler` only uses info level.

pkg/lvmd/cmd/root.go
ushitora-anqou: I fixed my code, and every goroutine now calls `cancel()` at the end to stop other goroutines.

ushitora-anqou:
Changes that I made after llamerada-jp's review:
https://github.com/topolvm/topolvm/compare/3398f277..711bbff5c42498048d55e86c3e5914b5228b546b

pkg/lvmd/cmd/root.go
```diff
@@ -97,31 +99,53 @@ func subMain(ctx context.Context) error {
...
+	go func() {
+		defer wg.Done()
+
+		childCtx, stop := signal.NotifyContext(ctx, os.Interrupt, syscall.SIGTERM)
+		defer stop()
+
+		<-childCtx.Done()
+
+		cancel()
 		grpcServer.GracefulStop()
```
llamerada-jp, line 128: Could you consider a smart way to this logic by merging this block with the `select` block of the ticker?

llamerada-jp (changes requested):

pkg/lvmd/cmd/root.go
ushitora-anqou: I fixed my code. Could you review it again? The changes I made is [here](https://github.com/topolvm/topolvm/compare/711bbff5c42498048d55e86c3e5914b5228b546b..e9a0ff19d8ea63d5509718f337859973c901f153).

pkg/topolvm-scheduler/cmd/root.go
```diff
@@ -80,23 +86,29 @@ func subMain() error {
...
+		ReadTimeout: 30 * time.Second,
 	}
 
-	err = serv.ListenAndServe()
-	if err != nil {
-		return err
-	}
-	err = well.Wait()
+	idleConnsClosed := make(chan struct{})
+	go func() {
```
llamerada-jp, line 96: This goroutine will continue to run although ListenAndServe() returns an error. It should be made to stop correctly.

llamerada-jp (changes requested):

pkg/topolvm-scheduler/cmd/root.go
ushitora-anqou: I fixed my code. Please review this again. The change is the following: https://github.com/topolvm/topolvm/compare/ae2be29c5ce1f1ac5ab4fd6777a3532f4866154f..cc0be51f79191ebf06ecdbfd8c805013f846b13b

pkg/topolvm-scheduler/cmd/root.go
```diff
@@ -80,23 +87,31 @@ func subMain() error {
...
+		ReadTimeout: 30 * time.Second,
 	}
 
-	err = serv.ListenAndServe()
-	if err != nil {
-		return err
-	}
-	err = well.Wait()
+	var wg sync.WaitGroup
+	defer wg.Wait()
```
llamerada-jp, line 97: Maybe it's a matter of preference, but you don't have to wait for goroutine to terminate using wg.

llamerada-jp (dismissed):

pkg/topolvm-scheduler/cmd/root.go
```diff
@@ -80,23 +86,26 @@ func subMain() error {
...
+	go func() {
+		<-ctx.Done()
+		if err := serv.Shutdown(parentCtx); err != nil {
+			logger.Error(err, "failed to shutdown gracefully")
+		}
+	}()
 
-	if err != nil && !well.IsSignaled(err) {
+	err = serv.ListenAndServe()
+	if !errors.Is(err, http.ErrServerClosed) {
```
daichimukai, line 106: Unlike `Server.Serve()` in grpc-go, `Server.ListenAndServe()` in net/http  does not wait until all connections are closed even if `Server.Shutdown()` is called. Instead, `ListenAndServe()` returns immediately, so we need to wait until `Shutdown()` returns to shutdown gracefully. With this code the main go routine exits when `Shutdown()` called, which results in the process itself finishes immediately without waiting existing connections. Is this ok?

> When Shutdown is called, Serve, ListenAndServe, and ListenAndServeTLS immediately return ErrServerClosed. Make sure the program doesn't exit and waits instead for Shutdown to return. 

https://pkg.go.dev/net/http#Server.Shutdown

ushitora-anqou: It seems not ok to me. Thank you for pointing this out. I'll fix my code.

ushitora-anqou: Now it's fixed. The change I made: https://github.com/topolvm/topolvm/compare/0684eb256a5fb0afddcc47124b79fc3a0757413a..6c6bf8ca1a3f1bb8c6d93d365505386543fa7103

daichimukai: Perfect! Thank you.

pkg/topolvm-scheduler/cmd/root.go
```diff
@@ -80,23 +87,31 @@ func subMain() error {
...
 
-	serv := &well.HTTPServer{
-		Server: &http.Server{
-			Addr:    config.ListenAddr,
-			Handler: h,
-		},
+	serv := &http.Server{
+		Addr:        config.ListenAddr,
+		Handler:     accessLogHandler(parentCtx, h),
+		ReadTimeout: 30 * time.Second,
```
daichimukai, line 93: Why did you set `ReadTimeout`?

daichimukai: I solved it myself. This is the default value set by cybozu-go/well.
https://github.com/cybozu-go/well/blob/8cbd22dd58421e97e3cd8a9ecc0ce65000dd682f/http.go#L22

daichimukai (approved):

llamerada-jp (approved):

PR #808, opened 2024-01-09
chore: optimize node watches and controller to only watch metadata and no spec

jakobmoellerdev:
This PR has 2 changes:

- includes a small Memory Optimization for the TopoLVM node controller. It makes use of the OnlyMetadata watch and cachebuilder from https://github.com/kubernetes-sigs/controller-runtime/blob/v0.16.3/pkg/builder/options.go#L102-L133 as the controllers in topolvm do not need to interact with Pod Spec / Status, only with annotations or labels
- It reduces the verbosity of GetPluginInfo and GetPluginCapabilities as they are regularly called by k8s and do not need to log a message every time

### DIFF
--- a/controllers/node_controller.go
+++ b/controllers/node_controller.go
@@ -10,10 +10,12 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


jakobmoellerdev:
PR is now ready for review, failures are flakes afaik, would be great if someone can retrigger

pluser:
Could you make a separate commit for the part that changes the logger level?
It looks good for that other part of the PR.

jakobmoellerdev:
Good callout, thanks! Rebased on a 2-commit history now

pluser (approved):

PR #811, opened 2024-01-11
Change VG names for readability

toshipp:
We used different VG names on kind and minikube and the names are too agnostic, so it is hard to maintain.


e2e/thin_provisioning_test.go
```diff
@@ -105,7 +102,7 @@ func testThinProvisioning() {
 		By("deploying multiple PVCS with total size < thinpoolsize * overprovisioning")
 		// The actual thinpool size is 4 GB . With an overprovisioning limit of 5, it should allow
 		// PVCs totalling upto 20 GB for each node
-		nodeName := "topolvm-e2e-worker2"
+		nodeName := "topolvm-e2e-worker"
```
toshipp, line 105: We create a namespace for each `It` and delete the namespace after the spec of the `It` ends. Therefore, it is ok to use the same node.

e2e/thin_provisioning_test.go
```diff
@@ -191,7 +185,7 @@ func testThinProvisioning() {
 	It("should check overprovision limits", func() {
 		By("Deploying a PVC to use up the available thinpoolsize * overprovisioning")
 
-		nodeName := "topolvm-e2e-worker3"
+		nodeName := "topolvm-e2e-worker"
```
toshipp, line 188: ditto

satoru-takeuchi (approved):

daichimukai (approved):

PR #816, opened 2024-01-15
feat!: helm chart and base image refactor DO NOT MERGE

ryanfaircloth:
Individual commits for the specific changes at a high level this pr removes the need for hostPID (container is still privileged) 
* remove lvmd as a independent ds it can still be external or embedded
* in embedded mode runs lvmd in its own container within the pod created by the DS
* moves the socket for inter-process communication to a emptydir()
* removes the dependency on the host having /sbin/lvm in the distro
* adds logging of the lvm command line (base64 encoded)
* Removes uselegacy
* moves CRDs to the crd folder for safer uninstall/reinstall

I really don't expect to see this PR merged but the feedback is useful and I will pull the parts out that make sense and submit as bit size PRs.


toshipp:
@ryanfaircloth 
hostPID is needed because lvm command should run in the host namespaces due to its posix semaphore usage.

lvmd/command/lvm.go
```diff
@@ -44,16 +44,20 @@ func callLVM(ctx context.Context, cmd string, args ...string) error {
...
 	c.Env = os.Environ()
 	c.Env = append(c.Env, "LC_ALL=C")
 	c.Stdout = &stdout
-	c.Stderr = os.Stderr
+	c.Stderr = &stderr
 
 	log.FromContext(ctx).Info("invoking LVM command", "args", args)
 	err := c.Run()
+
+	log.FromContext(ctx).Info("result of LVM command", "stdout", stdout.Bytes(), "stderr", stderr.Bytes())
```
jakobmoellerdev, line 59: Can we introduce this at a lower log level? I do not think this is necessary for common debugging

charts/topolvm/templates/node/daemonset.yaml
```diff
@@ -37,29 +37,73 @@ spec:
...
       {{- if .Values.node.lvmdEmbedded }}
-      hostPID: true
+      # hostPID: true
       {{- end }}
       {{- with .Values.node.initContainers }}
       initContainers: {{ toYaml . | nindent 6 }}
       {{- end }}
       containers:
-        - name: topolvm-node
+      {{- if .Values.node.lvmdEmbedded }}
```
jakobmoellerdev, line 46: lvmdEmbedded is fundamentally different from what you are trying to do here. Embedded means embedded in topolvm-node, not a separate binary. If you run lvmd separately from topolvm-node it is no longer embedded.

jakobmoellerdev (changes requested):

Dockerfile
```diff
@@ -1,56 +1,56 @@
...
+RUN apk add --no-cache \
+    make patch curl
 
 RUN make csi-sidecars GOARCH=${TARGETARCH}
 
 # TopoLVM container with sidecar
-FROM --platform=$TARGETPLATFORM topolvm as topolvm-with-sidecar
+FROM --platform=$TARGETPLATFORM alpine:3.19 as topolvm-with-sidecar
+
+RUN apk add --no-cache \
```
jakobmoellerdev, line 31: This will add licensing issues to the released TopoLVM images as they now package various OS packages that need to be mentioned in the licensing

.values.yaml
```diff
@@ -0,0 +1,223 @@
+image:
+  repository: ryanfaircloth/topolvm
```
jakobmoellerdev, line 2: leftover repo

charts/topolvm/templates/node/daemonset.yaml
```diff
@@ -24,8 +24,8 @@ spec:
         {{- with .Values.node.podLabels }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
-      {{- if and .Values.node.metrics.enabled .Values.node.metrics.annotations }}
-      annotations: {{ toYaml .Values.node.metrics.annotations | nindent 8 }}
+      {{- if and .Values.node.csi.metrics.enabled .Values.node.csi.metrics.annotations }}
```
jakobmoellerdev, line 27: breaking change with previous chart versions, needs a major version

charts/topolvm/templates/node/daemonset.yaml
```diff
@@ -220,22 +251,23 @@ spec:
...
           configMap:
-            name: {{ template "topolvm.fullname" . }}-lvmd-{{ $lvmdidx }}
-            {{- end }}
-          {{- end }}
+            name: {{ template "topolvm.fullname" . }}-lvmd
+        {{- end }}
+        {{- if .Values.node.lvmdEmbedded }}
+        - name: lvmd-socket-dir
+          emptyDir: 
+            medium: Memory
```
jakobmoellerdev, line 261: nice change, but are you sure this is the only thing requiring hostPID? Just because it starts doesnt mean all block devices are recognized properly

charts/topolvm/templates/node/daemonset.yaml
ryanfaircloth: Yes I do agree I changed the implementation and re-used the same flag the admin experience should be the same though would you like a new word here or a more verbose description of the change

charts/topolvm/templates/node/daemonset.yaml
ryanfaircloth: Agree added ! and Breaking Change to the commits to ensure this was communicated

lvmd/command/lvm.go
jakobmoellerdev: I would argument for V(2) at the very least for both commands then but maybe we put that into a separate issue as it affects pretty much all commands? WDYT @pluser @toshipp ?

charts/topolvm/templates/node/daemonset.yaml
jakobmoellerdev: Just trying to add some context as the original implementor of the embedded functionality: IMO this is a fundamentally different deployment from using `--embed-lvmd` in topolvm-node so we should find a different name for it. 

From what I can see, this functionality should not be removed just because we change the way that we move from 2 DS to 1 DS without using this flag. When using the flag, we do not even need to run LVMD so there will always be 1 DS.

I have 2 follow-up questions from your work:
1. How in the future would we allow users to run with `--embed-lvmd` on topolvm-node? IMO this is answered by running 1 DS with 2 containers in the default setting (lvmdEmbedded = false) or introducing a new flag that is mutually exclusive with lvmdEmbedded =  true
2. What tangible benefit does it bring to have a single daemon set with 2 containers over 2 ds with 1 container. I currently do not see major gains, would be cool to have this in the PR explanation

charts/topolvm/templates/node/daemonset.yaml
jakobmoellerdev: Thats nice! But maybe we should think about wether we need to move this CSI section at all? What benefit does it bring or is it necessary? If not, maybe we can avoid the breaking change alltogether?

Dockerfile
jakobmoellerdev: Honestly this is a great point. @toshipp I suggest we open up an issue on LICENSE tracking for all of our open source dependencies in TopoLVM so that we remain compliant in the project and questions like these dont arise in PRs. I suggest we do this before we merge here.

We (speaking on behalf of Red Hat) are interested in remaining License Compliant for TopoLVM and we personally do not use the Dockerfile in the open source version because we rebuild with Red Hat Maintained versions of packages such as lvm2 (which is also maintained and owned by Red Hat and published as open source).

Until this is not cleared up however, I suggest not to move forward with this change to make sure no licensing aspects change while we move for more transparency.

lvmd/command/lvm.go
ryanfaircloth: Overall what is logged and when could use some improvement its not "bad" but its not very intentional, for example at the info level I would like to know what vgs were found on startup so in the list function we should have a dump of the VGs, as LVs actions we should see that at info level. this really should be debug but as a non go programmer I wasn't really wanting to do more than I could safely copy and paste.

charts/topolvm/templates/node/daemonset.yaml
ryanfaircloth: 1) I see we have multiple implementation choices right now LVMD can be systemd, in its own DS or as a child process of the node DS container. In this PR I removed "embeded" as it is currently defined because that didn't seem to solve a unique problem or offer any value on its own and to me made troubleshooting more complicated. having LVMD as its own daemon set also doesn't seem to be particularly useful as an option because we could have race conditions between the pods or node may not start for some reason such as lack of resources. So really I moved the LVMD "managed" into this pod and redefined embeded. I think adding node.lvmd.external would be more descriptive as this really just reduces the deployment choices to two. DS (via node pod) and external

charts/topolvm/templates/node/daemonset.yaml
ryanfaircloth: Might have been over agressive but these options don't apply to the pod only the csi container so I was attempting to not introduce ambiguity with the changes to move lvmd into the pod

Dockerfile
ryanfaircloth: I also see the way this is being built today the sidecars from the kube project are embeded into the container, I don't see a THIRDPARTY or other sbom so I would like to continue with the change as its more likely slightly better due to alpine being a smaller base image with fewer inclusions, or a net same

jakobmoellerdev: Its a different story if you use third party container images or if you build your own container image imo. But lets wait for other maintainers opinions

charts/topolvm/templates/node/daemonset.yaml
jakobmoellerdev: > I see we have multiple implementation choices right now LVMD can be systemd, in its own DS or as a child process of the node DS container. 

It is not a child process if you run it within topolvm-node. It is one process and we do not open up a gRPC server in this case.

> In this PR I removed "embeded" as it is currently defined because that didn't seem to solve a unique problem or offer any value on its own and to me made troubleshooting more complicated. 

I believe you are quite quick to come to this resolution. Please reference the [dedicated design proposal](https://github.com/topolvm/topolvm/blob/main/docs/proposals/lvmd-within-topolvm-node.md) and see if you still do not understand the use case. Just because this option of deployment is not applicable for you does not mean it is not useful for other users.

> having LVMD as its own daemon set also doesn't seem to be particularly useful as an option because we could have race conditions between the pods or node may not start for some reason such as lack of resources. 

This does not make sense because 1. race conditions between pods are not a thing (we are an eventually consistent system, and starting one before the other does not result in problems) and 2. If the node does not have enough resources for both lvmd and topolvm-node and you run with 1DS the daemonset will still not start and run successfully.

> So really I moved the LVMD "managed" into this pod and redefined embeded. I think adding node.lvmd.external would be more descriptive as this really just reduces the deployment choices to two. DS (via node pod) and external

I disagree that you redefined embedded. You removed the option to configure it entirely. There is no way in your chart to enable this kind of configuration.
However I agree that having 2 deployment choices should be the goal. 
My conclusion is different however: I think your way of deploying (1 DS with 2 containers ) should succeed the 2 DS approach and NOT embedded LVMD.

charts/topolvm/templates/node/daemonset.yaml
jakobmoellerdev: Ah understood, makes sense to me :) . Then lets wait for other maintainers to give feedback on if this will be accepted.

Dockerfile
jakobmoellerdev: https://github.com/topolvm/topolvm/issues/818 for LICENSE discussions. Thanks for creating!

charts/topolvm/templates/node/daemonset.yaml
ryanfaircloth: re race conditions I have seen instances of the node pod crashing when lvmd is not ready, I have also seen a few instances where workload pods start before node/lvmd is ready

Thanks for the doc pointer, after reading I think my approach achieves the same goals in this design doc the goal of not needing a second DS is achieved by treating LVMd as a library (I thought it was process) with the consequence of having two very different interface methods. My approach does result in retention of an independent process and grpc but means the interface between node and lvmd is the same rather it is deployed in the pod or deployed as a systemd service. The goals is I understand them are the same.  My approach has a slightly smaller code surface less to maintain and would make the embeded moot assuming I understood the goals correctly. While I am removing choices have I caused an unforeseen consequence by impacting a feature or function.

charts/topolvm/templates/node/daemonset.yaml
ryanfaircloth: I did test in my environment which at the moment is EKS with k8s 1.28

jakobmoellerdev: See https://github.com/topolvm/topolvm/pull/816#issuecomment-1892928961

charts/topolvm/templates/node/daemonset.yaml
ryanfaircloth: Thanks for the back and forth I see the issue and will revise the approach to use embedded rather than side car

Dockerfile
satoru-takeuchi: @jakobmoellerdev @ryanfaircloth I'll handle #818.

toshipp:
@ryanfaircloth 
This PR is too big for me (as you said). Please make it smaller to solve a specific problem when you send the next PR, even if you need feedback.

lvmd/command/lvm.go
toshipp: I agree with @jakobmoellerdev . Please separate it.

toshipp:
I leave my comments for each commit.

7b33e32874e1244b19ef65d423204366ac9fc247
Since it is hard for existing users to migrate from the old API group to the new one, we don't want to delete it. So I decline this.

6cdae0549be75eb5ea9cb5912371871f633b7c68
I suppose there are both advantages and disadvantages to this change. Also, this is a break, we should separate it from the PR if you want to continue.

5c9d8599730fc53f1e9dfd04fd2e00f5c88f8904
This was introduced by the contributor for IaaS environments. We have less knowledge of them, it may work well, but this is a break. Please separate it. 

f62a1ac3d802fe8d3408c1b1b2b77c501ab99133
We want users to choose to use priorityClasses. I object to this.

3cce93dcaab5c5afd4097aff6c81b5786ce4a9e6
We are familiar with Ubuntu and chose Ubuntu because of its ease of maintenance rather than image size. I decline it.

ryanfaircloth:
> I leave my comments for each commit.
> 
> [7b33e32](https://github.com/topolvm/topolvm/commit/7b33e32874e1244b19ef65d423204366ac9fc247) Since it is hard for existing users to migrate from the old API group to the new one, we don't want to delete it. So I decline this.
> 
> [6cdae05](https://github.com/topolvm/topolvm/commit/6cdae0549be75eb5ea9cb5912371871f633b7c68) I suppose there are both advantages and disadvantages to this change. Also, this is a break, we should separate it from the PR if you want to continue.
> 
> [5c9d859](https://github.com/topolvm/topolvm/commit/5c9d8599730fc53f1e9dfd04fd2e00f5c88f8904) This was introduced by the contributor for IaaS environments. We have less knowledge of them, it may work well, but this is a break. Please separate it.
> 
> [f62a1ac](https://github.com/topolvm/topolvm/commit/f62a1ac3d802fe8d3408c1b1b2b77c501ab99133) We want users to choose to use priorityClasses. I object to this.
> 
> [3cce93d](https://github.com/topolvm/topolvm/commit/3cce93dcaab5c5afd4097aff6c81b5786ce4a9e6) We are familiar with Ubuntu and chose Ubuntu because of its ease of maintenance rather than image size. I decline it.

Thanks I will keep this open for easy access to the feedback and close it off when I've proposed the smaller PRs for proper discussion.

PR #817, opened 2024-01-16
feat: allow disabling leader election in topolvm-controller

jakobmoellerdev:
This allows disabling leader election for TopoLVM controller. This is important because all sidecars support it and we should offer running TopoLVM in a non-HA controller configuration for CPU/Memory Savings and to avoid surges. This is especially used in the edge context and will not affect existing Helm Charts or Datacenter deployments.

New helm chart version has a new flag for leader election in the controller section that triggers it for topolvm-controller and all sidecars. Flag is set to true by default if omitted.

Also gives an example of a Recreate Deployment strategy that will not impact the controller cache synchronization but is non-HA

### DIFF
--- a/charts/topolvm/values.yaml
+++ b/charts/topolvm/values.yaml
@@ -357,6 +357,10 @@ controller:
     # controller.nodeFinalize.skipped -- Skip automatic cleanup of PhysicalVolumeClaims when a Node is deleted.
     skipped: false
 
+  leaderElection:
+    # controller.leaderElection.enabled -- Enable leader election for controller and all sidecars.
+    enabled: true
+
   prometheus:
     podMonitor:
       # controller.prometheus.podMonitor.enabled -- Set this to `true` to create PodMonitor for Prometheus operator.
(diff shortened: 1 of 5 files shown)


charts/topolvm/values.yaml
```diff
@@ -402,6 +406,13 @@ controller:
   #  rollingUpdate:
   #    maxSurge: 50%
   #    maxUnavailable: 50%
+  #
+  # For non-leader-election mode, you can use this non-HA non-conflicting strategy:
+  #  type: Recreate
+  #  rollingUpdate:
+  #    maxSurge: 0
+  #    maxUnavailable: 1
+
```
pluser, line 415: Is this part necessary?
It seems that rollingUpdate.maxSurge and rollingUpdate.maxUnavailable cannot be specified for type: Recreate

pluser:
LGTM

pluser (approved):

toshipp (approved):

PR #819, opened 2024-01-16
feat: use upstream builds of side-cars

ryanfaircloth:
The current process was building the upstream side-cars with no specific options or modifications. All required arch versions are provided upstream this change simply uses as provided a future change could adopt the upstream containers


satoru-takeuchi:
We have a policy to build sidecar containers from source code. Since it's not documented, we'll state this policy if necessary.

PR #820, opened 2024-01-16
doc: Update maint process for k8s changes

ryanfaircloth:
Add metadata to chart.yaml and add maint docs instructions

### DIFF
--- a/charts/topolvm/Chart.yaml
+++ b/charts/topolvm/Chart.yaml
@@ -6,6 +6,7 @@ name: topolvm
 description: Topolvm
 version: 13.0.1
 appVersion: 0.25.0
+kubeVersion: "1.25-1.27"
 sources:
   - https://github.com/topolvm/topolvm
 
--- a/docs/maintenance.md
+++ b/docs/maintenance.md
@@ -17,53 +17,56 @@ The format is up to you, but this is very important to keep track of what change
 Basically, we should pay attention to breaking changes and security fixes first.
 If we find some interesting features added in new versions, please consider if we are going to use them or not and make a GitHub issue to incorporate them after the upgrading task is done.
 
-Note: Ubuntu 18.04 is used to keep the backward compatibility of XFS formatting. (https://github.com/topolvm/topolvm/pull/306)
+Note: Ubuntu 18.04 is used to keep the backward compatibility of XFS formatting. (<https://github.com/topolvm/topolvm/pull/306>)
 
 #### Kubernetes
 
 Choose the next version and check the [release note](https://kubernetes.io/docs/setup/release/notes/). e.g. 1.17, 1.18, 1.19 -> 1.18, 1.19, 1.20
 
 Edit the following files.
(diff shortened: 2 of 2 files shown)


satoru-takeuchi:
We don't use `kubeVersion` to allow users to use topolvm with unsupported k8s versions at their own risk. It's intentional.

charts/topolvm/Chart.yaml
```diff
@@ -6,6 +6,7 @@ name: topolvm
 description: Topolvm
 version: 13.0.1
 appVersion: 0.25.0
+kubeVersion: "1.25-1.27"
```
satoru-takeuchi, line 9: We don't use kubeVersion to allow users to use topolvm with unsupported k8s versions at their own risk. It's intentional.

docs/maintenance.md
```diff
@@ -17,53 +17,56 @@ The format is up to you, but this is very important to keep track of what change
 Basically, we should pay attention to breaking changes and security fixes first.
 If we find some interesting features added in new versions, please consider if we are going to use them or not and make a GitHub issue to incorporate them after the upgrading task is done.
 
-Note: Ubuntu 18.04 is used to keep the backward compatibility of XFS formatting. (https://github.com/topolvm/topolvm/pull/306)
+Note: Ubuntu 18.04 is used to keep the backward compatibility of XFS formatting. (<https://github.com/topolvm/topolvm/pull/306>)
```
satoru-takeuchi, line 20: Is there any differences between with or without `<` and `>` for readers? At a first glance, these characters are not visible in the formatted markdown.

docs/maintenance.md
```diff
@@ -17,53 +17,56 @@ The format is up to you, but this is very important to keep track of what change
...
+- `charts/topolvm/Chart.yaml` "kubeVersion"
 
 Next, we should update `go.mod` by the following commands.
 Please note that Kubernetes v1 corresponds with v0 for the release tags. For example, v1.17.2 corresponds with the `v0.17.2` tag.
+
 ```console
-$ VERSION=<upgrading Kubernetes release version>
-$ go get k8s.io/api@v${VERSION} k8s.io/apimachinery@v${VERSION} k8s.io/client-go@v${VERSION} k8s.io/mount-utils@v${VERSION}
+VERSION=<upgrading Kubernetes release version>
+go get k8s.io/api@v${VERSION} k8s.io/apimachinery@v${VERSION} k8s.io/client-go@v${VERSION} k8s.io/mount-utils@v${VERSION}
```
satoru-takeuchi, line 40: Whether `$ ` is necessary or not depends on the project. In topolvm project (not only topolvm but also other tools), we specify `$ ` to clarify whether commands should be run by root or not. So I'd like to keep this code snippet as is.

ryanfaircloth:
Feedback is the desired behavior is to allow customers to deploy outside of test matrix without the need to bypass a protection so this one can be closed

PR #823, opened 2024-01-17
feat!: Use lvmdEmebbed by default

ryanfaircloth:
BREAKING CHANGE:  Using lvmdEmbedded by default reduces pod count per host by 1 and causes no loss of functionality. Users are only impacted if they have added initicontainers or modified volumes/mounts.

### DIFF
--- a/charts/topolvm/values.yaml
+++ b/charts/topolvm/values.yaml
@@ -132,7 +132,7 @@ scheduler:
 # lvmd service
 lvmd:
   # lvmd.managed -- If true, set up lvmd service with DaemonSet.
-  managed: true
+  managed: false
 
   # lvmd.socketName -- Specify socketName.
   socketName: /run/topolvm/lvmd.sock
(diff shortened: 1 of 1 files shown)


pluser:
@jakobmoellerdev
The lvmd embedded option has only [been in place](https://github.com/topolvm/topolvm/pull/771) for a short time.
How long do you think we should wait for stabilization?

jakobmoellerdev:
lvmd embedded mode will be used in all Openshift TopoLVM installations by default starting with Openshift 4.15 Release which is coming end of next month. After that I suggest waiting 2-3 months so we can report issues upstream. So I suggest putting this on hold for the next quarter

pluser:
@jakobmoellerdev Thank you for the information.

Personally, I think it is a little early to get solid assurance of safety, so I would like to wait a few more months or so.

@ryanfaircloth Do you have any issues with waiting a few months?

ryanfaircloth:
> @jakobmoellerdev Thank you for the information.
> 
> Personally, I think it is a little early to get solid assurance of safety, so I would like to wait a few more months or so.
> 
> @ryanfaircloth Do you have any issues with waiting a few months?

I do not I just adopted this via helm chart config

jakobmoellerdev:
Hey guys revisiting this PR as part of regular maintenance. Do we want to get this merged at all? (its now been long enough that embedded lvmd is out and we have received positive resonance thus far in the OpenShift version of it which is productized in Red Hat).

pluser:
Hello.

After observing the feature for a sufficient period, I have no objections to enabling it by default.

However, upon investigating the impact of enabling it by default, I found that some tests fail. Therefore, we cannot merge this PR as it is. @ryanfaircloth, Would you like to revise the PR to include the necessary test fixes?

The current e2e tests for TopoLVM assume that lvmd is outside of Kubernetes, so it does not function correctly when moved inside Kubernetes. We need to fix so that the e2e tests pass with the default settings.

llamerada-jp:
create new issue and close this PR.
https://github.com/topolvm/topolvm/issues/1016

Issue #809, opened 2024-01-09
Startup taint support


ryanfaircloth:
There is a race condition where pods may be scheduled before node initialization implement a feature similar to the same in the ebs and efs csi components

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1232

peng225:
Hi @ryanfaircloth

My first impression is that using the taint to control the startup order will be complicated if many services adopt the same strategy.

Anyway, I will discuss this topic with other maintainers next week, so please wait for a while.

peng225:
@ryanfaircloth 

I talked with other maintainers about this issue.
We think that the [Pod Scheduling Readiness](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-scheduling-readiness/) is suitable for this purpose.

1. Add a gate by the mutating webhook to pods that use TopoLVM's PVC for which PV is already bound.
2. If topolvm-node becomes ready on a node, delete the gate.

If you create a PR, we can review it.

ryanfaircloth:
I think this is not the same use case as pod readiness. What I am looking to do is more like the use case for startupTaints with cillium, ebs and efs cni/csi when a DS is must be ready before a pod is suitable for scheduling the platform is configured to add the taint at startup and it is removed by the pod when it becomes ready.  See the pattern and solution for ebs

ryanfaircloth:
PR from EBS I could port here https://github.com/kubernetes-sigs/aws-ebs-csi-driver/commit/c731fb0f9a05a6a656e86b9412756e918a4a6624

peng225:
I read EBS's PR, and I still have a concern about what I said in the first comment.
> My first impression is that using the taint to control the startup order will be complicated if many services adopt the same strategy.

If topolvm and cilium reside in the same cluster, I guess cilium also needs to tolerate the topolvm's taint. If many services use their own taint for this use case, we need to handle who to tolerate what kind of taint. That can be bothering.

Also, I don't understand why you think the pod scheduling readiness is not suitable for this use case. The pod scheduling readiness is used to control when a pod should be ready for scheduling, so I believe this is exactly the kind of use case for which the pod scheduling is intended.

ryanfaircloth:
[...]

PR #824, opened 2024-01-17
feat: Support Startup taints

ryanfaircloth:
I common practice with csi and cni ds is to set a startup taints so that workload pods will not schedule until the ds is ready indicated by removal of the taint. This is optional the defined taint removal will be completed only if it exists. #809

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -44,6 +44,12 @@ Finally, check if TopoLVM is running. All TopoLVM pods should be running.
 kubectl get pod -n topolvm-system
 ```
 
+## Configure node startup taint
+
+There are potential race conditions on node startup (especially when a node is first joining the cluster) where pods/processes that rely on the a CSI Driver can act on a node before the CSI Driver is able to startup up and become fully ready. To combat this, the TopoLVM contains a feature to automatically remove a taint from the node on startup. Users can taint their nodes when they join the cluster and/or on startup, to prevent other pods from running and/or being scheduled on the node prior to the CSI Driver becoming ready.
+
+This feature is activated by default, and cluster administrators should use the taint `topolvm.io/agent-not-ready:NoExecute` (any effect will work, but `NoExecute` is recommended).
+
 ## Usage
 
 You can create PersistentVolumes (PV) by TopoLVM after installation succeeded.
--- a/pkg/topolvm-node/cmd/run.go
+++ b/pkg/topolvm-node/cmd/run.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"net"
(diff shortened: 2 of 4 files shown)


pkg/topolvm-node/cmd/run.go
```diff
@@ -200,3 +224,98 @@ func loadConfFile(ctx context.Context, cfgFilePath string) error {
 	)
 	return nil
 }
+
+// Struct for JSON patch operations
```
peng225, line 228: I guess you copied these code from [aws-ebs-csi-driver](https://github.com/kubernetes-sigs/aws-ebs-csi-driver) project. Am I right?

In that case, plese add following information as a comment at the top of every function you copied.

- Where the function is copied from.
- The fact that you have modified the copied function.
- (If exists) The contents of the NOTICE file in the source project.
  - In case of EBS, add "Copyright 2022 The Kubernetes Authors."
  - ref: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/5a56f6672dc474604bb90d65a3fc067560cf6c1b/NOTICE#L2

Example: 
https://github.com/topolvm/topolvm/blob/1b19381babdddd94a3f96cd0b2bed861ffead95f/filesystem/util.go#L189-L192
(In this case, the code is copied and not modified.)

docs/getting-started.md
```diff
@@ -44,6 +44,12 @@ Finally, check if TopoLVM is running. All TopoLVM pods should be running.
 kubectl get pod -n topolvm-system
 ```
 
+## Configure node startup taint
+
+There are potential race conditions on node startup (especially when a node is first joining the cluster) where pods/processes that rely on the a CSI Driver can act on a node before the CSI Driver is able to startup up and become fully ready. To combat this, the TopoLVM contains a feature to automatically remove a taint from the node on startup. Users can taint their nodes when they join the cluster and/or on startup, to prevent other pods from running and/or being scheduled on the node prior to the CSI Driver becoming ready.
```
peng225, line 49: I guess you copied these code from [aws-ebs-csi-driver](https://github.com/kubernetes-sigs/aws-ebs-csi-driver) project. Am I right?

In that case, please replace the contents of this section with the link to the original document.

(I guess [this one](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/5a56f6672dc474604bb90d65a3fc067560cf6c1b/docs/install.md#configure-node-startup-taint) is the original.)

pkg/topolvm-node/cmd/run.go
peng225: Sorry, but I've changed my mind.
Could you create a new file and put all copied functions in the new file?
Then, add a comment at the top of the file.

docs/getting-started.md
ryanfaircloth: sure I will probably have time over the weekend. I may go ahead with a new PR on this so I can compare the two functionally

Issue #822, opened 2024-01-17
Tacking issue undocumented "policy" causes increased build time and additional time tracking and updating dependencies


ryanfaircloth:
We have a policy to build sidecar containers from source code. Since it's not documented, we'll state this policy if necessary.

_Originally posted by @satoru-takeuchi in https://github.com/topolvm/topolvm/issues/819#issuecomment-1894783498_

satoru-takeuchi:
Could you fix the subject and the description of this issue? I do not handle such the issue that the title is just a copy-and-pasted sentences from another issue.

Please stop making maintainers pay high communication costs.

ryanfaircloth:
@satoru-takeuchi in the context your replied that the project has a policy which is not documented but fairly different than similar other projects. What could have saved some steps would have been opening the issue to track yourself rather than reply.

satoru-takeuchi:
I'll state the policy about building sidecar containers by our own in the official document.

PR #825, opened 2024-01-18
doc: state the reason to build csi-sidecars

satoru-takeuchi:
We build csi-sidecars in our own. It's convenient for developers to clarify why we do so.

Fixes: https://github.com/topolvm/topolvm/issues/822

### DIFF
--- a/docs/maintenance.md
+++ b/docs/maintenance.md
@@ -78,6 +78,9 @@ Edit the following files.
 
 #### CSI Sidecars
 
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


toshipp (approved):

ryanfaircloth:
While the reason makes sense on the surface any emergency would likely impact all similar projects equally has this ever occurred is the overhead actually worth the reword. As a counter point I think a more likely issue is conflict with a distro provided side car, aks, eks and open shift all provide their own builds. Redhat does not use the project container ,building their own. While there reason you raise is valid it appears to me to simply be a trade off. As a middle ground the project could build and provide the side cars as independent builds and default to use those images or make that a option selected as part of the deployment

peng225 (approved):

PR #826, opened 2024-01-18
fix: CVE-2023-47108 & CVE-2023-45142

suleymanakbas91:
This PR fixes CVE-2023-47108 & CVE-2023-45142 by excluding otelhttp and otelgrpc dependencies that are not used by TopoLVM. 

`otelhttp` exclude can be removed once TopoLVM updates to the latest version of `apiextensions-apiserver`. 

However, to remove `otelgrpc` exclude, `apiextensions-apiserver` should first update to the `v0.46.0` of `otelgrpc`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -108,3 +108,8 @@ require (
 	sigs.k8s.io/json v0.0.0-20221116044647-bc3834ca7abd // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 )
+
+exclude (
+	go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.35.0 // to fix CVE-2023-47108
+	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.35.1 // to fix CVE-2023-45142
+)


daichimukai:
Hi @suleymanakbas91,

Thank you for PR. If I understand correctly, otelhttp and otelgrpc appear in the dependency graph, but actually neither is used to build TopoLVM. So I think CVE-2023-47108 and CVE-2023-45142 don't affect TopoLVM. Am I missing something?

```console
$ go mod why go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp
# go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp
(main module does not need package go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp)
$ go mod why go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc
# go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc
(main module does not need package go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc)
```

jakobmoellerdev:
Hey @daichimukai, while it is true go mod why does not reflect this dependency (because it is a transitive dependency), it is still used as part of the [MVS](https://go.dev/ref/mod#minimal-version-selection) algorithm for the build, which determines which packages in which version are used for source code build.

You can verify this by calling `go list -m go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` which will show `v0.35.1` being selected for build.

Note that the module doesn't appear in the graph, because the graph is based on go.sum and - quoted from MVS algorithm -

> Unlike other dependency management systems, the build list is not saved in a “lock” file. MVS is deterministic, and the build list doesn’t change when new versions of dependencies are released, so MVS is used to compute it at the beginning of every module-aware command.

daichimukai:
I rebuilt all packages of `topolvm/topolvm` along with its dependencies, but I couldn't obtained an evidence of otelgrcp or otelhttp being used.

```console
$ go build -a -v github.com/topolvm/topolvm/... 2>&1 | grep -E 'otel|opentelemetry'
```

jakobmoellerdev:
I think you are right in that it is not part of the compiled topolvm binary. I think it will still be a potential transitive in the module graph so sec scanners will still report it oftentimes.

suleymanakbas91:
Thanks for the comments, @daichimukai! You're right, these are not part of the binary. I'm closing the PR.

Issue #760, opened 2023-10-06
Use kubebuilder project v4.


toshipp:
## What

Since kubebuilder deprecated v3, migrate it to v4.
https://github.com/kubernetes-sigs/kubebuilder/releases/tag/v3.11.0

## How

Describe how to address the issue.

## Checklist

- [x] Finish implementation of the issue
- [x] Test all functions
- [x] Have enough logs to trace activities
- [x] Notify developers of necessary actions

llamerada-jp:
The code that violates lint will be handled in a follow-up issue.
https://github.com/topolvm/topolvm/issues/853

Issue #736, opened 2023-08-16
Update supporting Kubernetes (1.28)


cupnes:
## Update Procedure

- Read [this document](https://github.com/topolvm/topolvm/blob/main/docs/maintenance.md).

## Before Check List

There is a check list to confirm depending libraries or tools are released. The release notes for Kubernetes should also be checked.

### Must Update Dependencies

Must update Kubernetes with each new version of Kubernetes.

- [x] k8s.io/api
  - https://github.com/kubernetes/api/tags
    - The supported Kubernetes version is written in the description of each tag.
- [x] k8s.io/apimachinery
  - https://github.com/kubernetes/apimachinery/tags
    - The supported Kubernetes version is written in the description of each tag.
- [x] k8s.io/client-go
  - https://github.com/kubernetes/client-go/tags
    - The supported Kubernetes version is written in the description of each tag.
- [x] k8s.io/mount-utils
  - https://github.com/kubernetes/mount-utils/tags
    - The supported Kubernetes version is written in the description of each tag.
- [x] sigs.k8s.io/controller-runtime
  - https://github.com/kubernetes-sigs/controller-runtime/releases
- [x] sigs.k8s.io/controller-tools
  - https://github.com/kubernetes-sigs/controller-tools/releases
- [x] external provisioner
  - https://github.com/kubernetes-csi/external-provisioner/tree/master/CHANGELOG
- [x] external resizer
  - https://github.com/kubernetes-csi/external-resizer/tree/master/CHANGELOG
- [x] external snapshotter
  - https://github.com/kubernetes-csi/external-snapshotter/tree/master/CHANGELOG
- [x] kind
  - https://github.com/kubernetes-sigs/kind/releases
- [x] minikube
  - https://github.com/kubernetes/minikube/releases
- [x] crictl
  - https://github.com/kubernetes-sigs/cri-tools/releases

### Semi-required Dependencies

These are not released on the occasion of a Kubernetes version upgrade, so if there is a release that corresponds to a Kubernetes version, use it.

- [x] node driver registrar
  - https://github.com/kubernetes-csi/node-driver-registrar/tree/master/CHANGELOG
- [x] liveness probe
  - https://github.com/kubernetes-csi/livenessprobe/tree/master/CHANGELOG
- [x] cri-docker
  - https://github.com/Mirantis/cri-dockerd/releases

### Release notes check

- [x] Read the necessary release notes for Kubernetes.

## Checklist

- [x] Finish implementation of the issue
[...]

PR #830, opened 2024-01-24
use kubebuilder go/v4 template

llamerada-jp:
Update kubebuilder template to go/v4 from go/v3. However, no code is moved to the cmd, internal, and test directories. This would be done in a different PR to avoid conflicts with other PRs. And the resolutions to lint's point will be addressed after the directory move is complete.

a part of: https://github.com/topolvm/topolvm/issues/760


peng225 (commented):
In [this commit message](https://github.com/topolvm/topolvm/pull/830/commits/3d0c79741596a9659a379754fe40739f3b1da1cd), you say as follows:

> And since this project does not use custom
resources, generate-api is not generated either. Therefore, the
related codes are removed.

However, TopoLVM uses LogicalVolume custom resource. Is that OK to remove `generate-api` target?

cupnes (commented):
Other than what @peng225 pointed out, there seems to be no problem.

llamerada-jp:
@peng225 @cupnes 
The change was wrong and generate-api has been reverted.

go.mod
```diff
@@ -8,29 +8,31 @@ require (
...
 	github.com/spf13/viper v1.10.1
 	golang.org/x/sys v0.15.0
 	google.golang.org/grpc v1.58.3
 	google.golang.org/grpc/cmd/protoc-gen-go-grpc v1.2.0
 	google.golang.org/protobuf v1.31.0
-	k8s.io/api v0.27.3
-	k8s.io/apimachinery v0.27.3
-	k8s.io/client-go v0.27.3
-	k8s.io/klog/v2 v2.90.1
+	k8s.io/api v0.28.0
```
peng225, line 23: api version is upgraded to v0.28.0. IIUC, This means that the supported Kubernetes version will be changed. 

I thought this should be done in [the Kubernetes upgrade task](https://github.com/topolvm/topolvm/issues/736).

llamerada-jp: I work on the k8s 1.28 support before doing the merge of this PR. After that, I will rebase.
https://github.com/topolvm/topolvm/issues/736

llamerada-jp: I have rebased this PR. Would you review it again?

cupnes (approved):

peng225 (approved):

PR #831, opened 2024-01-24
chore: reduce GetCapacity log level

jakobmoellerdev:
This reduces the log level of the GetCapacity called message and the ignored nil capability argument as they are regularly called by K8s/CSI and spam the logs without adding significant value for debugging or troubleshooting.

This will result in these messages not being logged by default but still being accessible with higher log level


llamerada-jp (commented):
According to the official documentation, if V(2) is set, the logs are not output even if DebugLevel is enabled. I think V(1) is better because I want it to output when DebugLevel is enabled. What do you think?

https://github.com/go-logr/zapr?tab=readme-ov-file#increasing-verbosity

satoru-takeuchi (approved):

llamerada-jp (approved):

PR #832, opened 2024-01-24
chore: reduce consistent reads per mount to 1 in `IsMounted` based on kernel version

jakobmoellerdev:
The consistent read issue reported in https://github.com/topolvm/topolvm/issues/371 has been fixed by doing a consistent read 3 times. This is not necessary as 2 consistent reads will be enough to cover page size timings. 

This will severely reduce memory allocations and gc stability during LV creation:

Reasoning: 
See a profiling made with 32 PVCs created and allocs collected.

image

### DIFF
--- a/filesystem/util.go
+++ b/filesystem/util.go
@@ -1,11 +1,22 @@
+/*
+ * kernelHasMountinfoBug() and the constants only used in this function are copied
+ * from the following code:
+ * https://github.com/kubernetes/mount-utils/blob/6f4aae5a6ab58574cac605cdd48bf5c0862c047f/mount_helper_unix.go#L211-L242
+ *    LICENSE: http://www.apache.org/licenses/LICENSE-2.0
+ *    Copyright The Kubernetes Authors.
+ */
+
 package filesystem
 
 import (
+	"bytes"
 	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"strings"
+	"sync"
 
 	"golang.org/x/sys/unix"
 	"k8s.io/utils/io"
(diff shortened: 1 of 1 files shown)


ushitora-anqou:
I'm hesitating to change the number from 3 to 2 because most Kubernetes components seem to adopt 3 or more ([link 1][1], [link 2][2], [link 3][3], and [link 4][4]).

One of my team members found that this bug in `/proc/mounts` was fixed in [Linux 5.8][5]. How about updating your code to check the Linux version the code is running on, and just use `os.ReadFile` if it is new enough?

[1]: https://github.com/kubernetes/utils/blob/e7106e64919e62d73c0926352134656a22dc2e38/mount/mount_helper_unix.go#L36
[2]: https://github.com/kubernetes/utils/blob/e7106e64919e62d73c0926352134656a22dc2e38/io/read_test.go#L80
[3]: https://github.com/kubernetes/utils/blob/e7106e64919e62d73c0926352134656a22dc2e38/io/read_test.go#L93
[4]: https://github.com/kubernetes/mount-utils/blob/6f4aae5a6ab58574cac605cdd48bf5c0862c047f/mount_helper_unix.go#L42
[5]: https://github.com/kubernetes/mount-utils/blob/6f4aae5a6ab58574cac605cdd48bf5c0862c047f/mount_helper_unix.go#L211-L242

jakobmoellerdev:
@ushitora-anqou thanks for the insight! sure that works for me. Anything that is able to reduce these file reads will significantly reduce Allocs / Processor usage for FS based volumes. 

Regarding the linux check, any preference on how to do that?

peng225 (dismissed):

ushitora-anqou (approved):

peng225 (approved):

PR #834, opened 2024-01-29
support Kubernetes v1.28

llamerada-jp:
issue: #736 

Containing the change to be noted:
`node-role.kubernetes.io/master` toleration is removed from topolvm-scheduler, since it is not used from kubernetes v1.25 or later.
see: https://github.com/kubernetes/enhancements/blob/master/keps/sig-cluster-lifecycle/kubeadm/2067-rename-master-label-taint/README.md


pluser (approved):

go.mod
```diff
@@ -8,51 +8,59 @@ require (
...
 
+require (
+	github.com/google/gnostic-models v0.6.8 // indirect
+	github.com/pelletier/go-toml/v2 v2.1.0 // indirect
+	github.com/sagikazarmark/locafero v0.3.0 // indirect
+	github.com/sagikazarmark/slog-shim v0.1.0 // indirect
+	github.com/sourcegraph/conc v0.3.0 // indirect
+	golang.org/x/exp v0.0.0-20230905200255-921286631fa9 // indirect
+)
+
```
daichimukai, line 42: Why do these indirect dependencies has its own independent block?

daichimukai (commented):
Minor comments only.

> support Kubernetes v1.28
> 
> The version of dependent libraries and cis sidecars are upgraded for
k8s 1.28.

Typo in the commit message.

go.mod
llamerada-jp: Thanks. Both points have been updated.

go.mod
```diff
@@ -5,29 +5,32 @@ go 1.20
...
 	github.com/container-storage-interface/spec v1.6.0
 	github.com/go-logr/logr v1.2.4
+	github.com/go-logr/zapr v1.2.4
 	github.com/golang/protobuf v1.5.3
 	github.com/google/go-cmp v0.5.9
 	github.com/kubernetes-csi/csi-test/v5 v5.0.0
-	github.com/onsi/ginkgo/v2 v2.9.5
-	github.com/onsi/gomega v1.27.7
-	github.com/prometheus/client_golang v1.15.1
+	github.com/kubernetes-csi/external-snapshotter/client/v6 v6.3.0
```
daichimukai, line 12: external-snapshotter is upgraded to v6.3.3 in `versions.mk`, so it might be better to use v6.3.3.

llamerada-jp: I thought so too. But errors like the below happen when `go mod tidy` with specified v6.3.3.
```
go mod tidy
go: downloading github.com/kubernetes-csi/external-snapshotter/client/v6 v6.3.3
github.com/topolvm/topolvm/e2e tested by
        github.com/topolvm/topolvm/e2e.test imports
        github.com/kubernetes-csi/external-snapshotter/client/v6/apis/volumesnapshot/v1: reading github.com/kubernetes-csi/external-snapshotter/client/go.mod at revision client/v6.3.3: unknown revision client/v6.3.3
github.com/topolvm/topolvm/api/legacy/v1 imports
        k8s.io/apimachinery/pkg/runtime tested by
        k8s.io/apimachinery/pkg/runtime.test imports
        github.com/stretchr/testify/assert: github.com/kubernetes-csi/external-snapshotter/client/v6@v6.3.3: reading github.com/kubernetes-csi/external-snapshotter/client/go.mod at revision client/v6.3.3: unknown revision client/v6.3.3
...
```

I don't know the correct reason but I guessed that `github.com/kubernetes-csi/external-snapshotter/client/v6/` is independent of `github.com/kubernetes-csi/external-snapshotter`. and there are no changes in `client/v6` from v6.3.0. Therefore, there may be `external-snapshotter` v6.3.3 but not `client/v6` v6.3.3.
https://github.com/kubernetes-csi/external-snapshotter/tree/v6.3.3/client

daichimukai: 🤯

> there are no changes in client/v6 from v6.3.0. 

Anyway I understand it doesn't matter that we select which version. Thank you.

daichimukai (approved):

pluser (approved):

PR #843, opened 2024-02-26
lvmd: fix Watch RPC to notify VG info periodically when embedded

daichimukai:
The Watch RPC of lvmd's VGService is expected to notify VG information periodically to its watchers. However, when lvmd is embedded into topolvm-node, Watch RPC sends VG info only once on startup time and does nothing after that unless LVService invokes LVM commands for RPCs. This commit addresses the issue.

Fixes: c5f17a4eba68 ("feat: implement embed-lvmd")

### DIFF
--- a/lvmd/embed.go
+++ b/lvmd/embed.go
@@ -2,6 +2,7 @@ package lvmd
 
 import (
 	"context"
+	"time"
 
 	"github.com/topolvm/topolvm/lvmd/proto"
 	"google.golang.org/grpc"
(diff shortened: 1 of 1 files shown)


lvmd/embed.go
```diff
@@ -29,6 +31,20 @@ func NewEmbeddedServiceClients(ctx context.Context, dcmapper *DeviceClassManager
 			log.FromContext(ctx).Error(err, "embedded channel watch error")
 		}
 	}()
+
+	go func() {
+		ticker := time.NewTicker(10 * time.Minute)
```
jakobmoellerdev, line 36: Thanks for the fix idea! QQ: How does the regular trigger for notify work on non-embedded mode? Im wondering how its able to do this refresh in the non-embedded mode

jakobmoellerdev: NVM, just found https://github.com/topolvm/topolvm/blob/baaea34e4eedbda153232ad01f62504ddcf88395/pkg/lvmd/cmd/root.go#L105-L117

Should we extract this into a common method and also make the time interval modifiable now?

jakobmoellerdev (approved):

lvmd/embed.go
daichimukai: Yes, that's a great idea. But I would like to leave that as a separate task, as my priority is to merge this fix and include it in next month's release.

peng225 (approved):

cupnes (approved):

PR #844, opened 2024-02-28
Introduce 4096 byte Sector Size Minimum for Rounding LV Sizes on Creation and Resizing

jakobmoellerdev:
This PR adds a 4096 Minimum Size limitation for all Logical Volumes that are created. The reason is that if one decides to use byte level precision on a PVC, that byte amount is now passed through to lvcreate. However if this byte-amount is not a multiple of the sector size (for most machines either 512, 1024 or 4096) then LVM will throw an error like

```
Size is not a multiple of 512. Try using 221920768 or 221921280.
  Invalid argument for --virtualsize: 221920847b 
```

In this case, we should round down in order to allow for a PVC creation with slightly different Size than the one requested as this is much better than getting rejected the PVC scheduling. 

The code is now changed to automatically round down or up (based on wether the limit or the request bytes are used and which one is larger). For edge cases during rounding we introduce new error cases.

Note that even though the 4KiB limit here is important for validation, the size might in the end still be rounded up to the nearest physical extent size (4MiB by default)

### DIFF
--- a/internal/driver/controller.go
+++ b/internal/driver/controller.go
@@ -430,17 +430,50 @@ func convertRequestCapacityBytes(requestBytes, limitBytes int64) (int64, error) 
 		)
 	}
 
-	// if requestBytes is 0, default to max(1Gi, limitBytes>0)
-	// 1. if limitBytes is 0, default to 1Gi
-	// 2. if limitBytes is greater than 1Gi, default to 1Gi
-	// 3. if limitBytes is less than 1Gi, default to limitBytes
+	// if requestBytes is 0 and
+	// 1. if limitBytes == 0, default to 1Gi
+	// 2. if limitBytes >= 1Gi, default to 1Gi
+	// 3. if limitBytes < 1Gi, default to rounded-down 4096 multiple of limitBytes.
+	//    if rounded-down 4096 multiple of limitBytes == 0, return error
 	if requestBytes == 0 {
-		if limitBytes > 0 && topolvm.DefaultSize > limitBytes {
-			return limitBytes, nil
+		// if there is no limit or the limit is bigger or equal to the default, use the default
+		if limitBytes == 0 || limitBytes >= topolvm.DefaultSize {
+			return topolvm.DefaultSize, nil
 		}
-		return topolvm.DefaultSize, nil
+
+		roundedLimit := roundDown(limitBytes, topolvm.MinimumSectorSize)
+		if roundedLimit == 0 {
+			return 0, fmt.Errorf("requested capacity is 0, because it defaulted to the limit (%d) and was rounded down to the nearest sector size (%d). "+
+				"specify the limit to be at least %d bytes", limitBytes, topolvm.MinimumSectorSize, topolvm.MinimumSectorSize)
+		}
+
+		return roundedLimit, nil
 	}
+
+	if requestBytes%topolvm.MinimumSectorSize != 0 {
+		// round up to the nearest multiple of the sector size
+		requestBytes = roundUp(requestBytes, topolvm.MinimumSectorSize)
+		// after rounding up, we might overshoot the limit
+		if limitBytes > 0 && requestBytes > limitBytes {
+			return 0, fmt.Errorf(
+				"requested capacity rounded to nearest sector size (%d) exceeds limit capacity, "+
+					"either specify a lower request or a higher limit: request=%d limit=%d",
+				topolvm.MinimumSectorSize, requestBytes, limitBytes,
+			)
+		}
+	}
+
 	return requestBytes, nil
+}
+
+// roundUp rounds up the size to the nearest given multiple.
+func roundUp(size int64, multiple int64) int64 {
+	return (size + multiple - 1) / multiple * multiple
+}
+
+// roundDown rounds down the size to the nearest given multiple.
+func roundDown(size int64, multiple int64) int64 {
+	return size - size%multiple
 }
 
 func (s controllerServerNoLocked) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeRequest) (*csi.DeleteVolumeResponse, error) {
--- a/constants.go
+++ b/constants.go
@@ -94,6 +94,13 @@ const DefaultSizeGb = 1
 // DefaultSize is DefaultSizeGb in bytes
 const DefaultSize = int64(DefaultSizeGb << 30)
 
+// MinimumSectorSize is the minimum size in bytes for volumes (PVC or generic ephemeral volumes).
+// It is derived from the usual sector size of 512,1024 or 4096 bytes for logical volumes.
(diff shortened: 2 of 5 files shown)


daichimukai:
Hi @jakobmoellerdev,

Thank you for your PR! Looks like a reasonable change to me. Before we start a review of the implementation, let me discuss the implementation strategy.

I agree that we introduce LV size limitation to fit into the 4096 byte alignment. But I'm not sure it would be reasonable to implement this logic in lvmd.

The lvmd creates LVs with the size decided by the topolvm-controller based on the `capacity_range` field of the CSI's CreateVolume RPC request. According to the spec, `capacity_range` must be honored if specified. I'm worried that we will be violating this spec if lvmd rounds down the requested size. (For example, consider the case where the request in `capacity_range` is 4MiB+1B.)

>  // This field is OPTIONAL. This allows the CO to specify the capacity
>  // requirement of the volume to be provisioned. If not specified, the
>  // Plugin MAY choose an implementation-defined capacity range. If
>  // specified it MUST always be honored, even when creating volumes
>  // from a source; which MAY force some backends to internally extend
>  // the volume after creating it.
>  CapacityRange capacity_range = 2;

https://github.com/container-storage-interface/spec/blob/886cd4877954568f9722cbe14d46b9fc16379d62/spec.md#createvolume

So, I'll suggest the following strategy:

1. The topolvm-controller creates LogicalVolume CR with the size which is the request of `capacity_range` rounded **up** to fit into the 4096 byte alignment.
2. (optional) The lvmd checks if the requested LV size is aligned with the 4096 byte alignment.

Additionally, we might need to round up the size when the topolvm-controller adds requests to Pods.

jakobmoellerdev:
Thanks for the input. Agreed that we can change the approach to make it better fit into the CSI Spec.

I just checked CapacityRange and it seems like there are two components to it: RequiredBytes and LimitBytes.

I suggest we use your approach then and do the following logic:

1. [CreateVolume](https://github.com/topolvm/topolvm/blob/94b7f88fa74c391368e8906ceb6649aa52d29302/driver/controller.go#L114) verifies `req.GetCapacityRange().GetRequiredBytes() % 4096`, if it does not solve to 0, round up to nearest 4096 sector. If that new rounded value is larger than req.GetCapacityRange().GetLimitBytes(), throw `status.Error(codes.InvalidArgument, "capacity range is invalid, request was rounded up to nearest 4096 sector but cannot fit into capacity limit")` 
2. LVMD verifies that `size % 4096 != 0` and if it is, we should throw an error


Towards your point of

> Additionally, we might need to round up the size when the topolvm-controller adds requests to Pods.

Could you tell me what needs to change and why here? Then I will implement accordingly. Not sure what this means right now (maybe you mean the resource extensions on the Pod)

daichimukai:
> >  Additionally, we might need to round up the size when the topolvm-controller adds requests to Pods.
> 
> Could you tell me what needs to change and why here? Then I will implement accordingly. Not sure what this means right now (maybe you mean the resource extensions on the Pod)

I changed my mind. Let's leave it as is. When the scheduler decides a node to which allocate Pods, it would be better to know the accurate size of the LVs to create as much as possible. However, since the discrepancy is at most 4096 bytes, we can leave it as it is until it becomes a problem.

driver/controller.go
```diff
@@ -440,9 +440,28 @@ func convertRequestCapacityBytes(requestBytes, limitBytes int64) (int64, error)
...
+			)
+		}
+	}
+
 	return requestBytes, nil
 }
 
+// roundUp rounds up the size to the nearest given multiple.
+func roundUp(size int64, multiple int64) int64 {
+	return (multiple - size%multiple) + size
```
daichimukai, line 462: If `multiple` divides `size`, `roundUp(size, multiple)` returns `size + multiple` that would be not what intended.

```suggestion
	return (size + multiple - 1) / multiple * multiple
```

daichimukai (changes requested):
https://github.com/topolvm/topolvm/blob/976abaf1ffc6c5ab3a6336e0f6a9d8b1852e5d54/driver/controller.go#L433-L442

Take care of the case where `requestByte == 0` and `limitByte < 4096`. It is not allowed now. It might be better not to use early returns and force return values always pass the "round up logic".

https://github.com/topolvm/topolvm/blob/976abaf1ffc6c5ab3a6336e0f6a9d8b1852e5d54/driver/controller_test.go#L32-L38

Then this test case needs to be fixed.

jakobmoellerdev:
> Take care of the case where requestByte == 0 and limitByte < 4096. It is not allowed now. It might be better not to use early returns and force return values always pass the "round up logic".

I believe this is covered already:

```
if requestBytes == 0 {
   if limitBytes > 0 && topolvm.DefaultSize > limitBytes { 
      return limitBytes, nil 
   } 
}
```

translates to 

```
if requestBytes == 0 & limitBytes > 0 && limitBytes < 4096
   return limitBytes
```

driver/controller.go
jakobmoellerdev: Good catch, thanks :)

internal/driver/controller.go
```diff
@@ -430,19 +430,40 @@ func convertRequestCapacityBytes(requestBytes, limitBytes int64) (int64, error)
...
+	// 3. if limitBytes < 1Gi, default to rounded-up 4096 multiple of limitBytes
 	if requestBytes == 0 {
-		if limitBytes > 0 && topolvm.DefaultSize > limitBytes {
-			return limitBytes, nil
+		// if there is no limit or the limit is bigger or equal to the default, use the default
+		if limitBytes == 0 || limitBytes >= topolvm.DefaultSize {
+			return topolvm.DefaultSize, nil
+		}
+
+		return roundUp(limitBytes, topolvm.MinimumSectorSize), nil
```
daichimukai, line 443: As we can see from the test for `convertRequestCapacityBytes(0, topolvm.MinimumSectorSize-1)` ), in this case,  the return value might exceed the limit, which violates the CSI spec. I think we should round **down** the limit and

1. if zero return error,
2. otherwise return it.

daichimukai (changes requested):

internal/driver/controller.go
jakobmoellerdev: Instead of changing to rounding down, we can also decide to simply reject this case and return an error if the limit<MinimumSectorSize. I believe this is actually a much cleaner behavior. Rounding down will provision less capacity than a PVC requests and that might become a problem, while overprovisioning never becomes problematic for the Application. WDYT?

daichimukai:
@jakobmoellerdev
Thank you, LGTM. Please squash commits into one commit.

daichimukai (approved):
LGTM

jakobmoellerdev:
@pluser mind taking a look over this one now that first review has passed?

pluser (approved):

PR #849, opened 2024-03-05
Avoid disk nospace when release workflow

toshipp:
The release workflow failed because of disk no space. see https://github.com/topolvm/topolvm/actions/runs/8137581069/job/22237189098
I doubt actions runner has consumed more disk space than before. To avoid disk exhaustion, use dedicated jobs to build images.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -4,9 +4,13 @@ on:
     tags:
       - "v*"
 jobs:
-  release:
-    name: "release"
-    runs-on: "ubuntu-20.04"
+  prepare:
+    name: "prepare"
+    outputs:
+      version: ${{ steps.check_version.outputs.version }}
+      prerelease: ${{ steps.check_version.outputs.prerelease }}
+      image_prefix: ${{ steps.export_image_prefix.outputs.image_prefix }}
+    runs-on: "ubuntu-latest"
     steps:
       - name: "Validate Release Version"
         id: check_version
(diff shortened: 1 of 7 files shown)


toshipp:
The CI passed
https://github.com/topolvm/topolvm/actions/runs/8155224884

.github/workflows/release.yaml
```diff
@@ -19,35 +23,45 @@ jobs:
...
       - name: cache go dependencies
-        uses: actions/cache@v4
-        with:
-          path: |
-            ~/go/pkg/mod
-            ~/.cache/go-build
-          key: go-${{ hashFiles('go.sum', 'Makefile', 'versions.mk') }}
-          restore-keys: |
-            go-
-      - run: make setup
```
toshipp, line 42: I deleted this because tools seem not to be used by other steps.

.github/actions/cache-go-deps/action.yaml
daichimukai, line 1: Is this cache useful still now? The `actions/setup-go` also caches dependencies.

toshipp: I don't know that. I will remove our cache solution.

toshipp: The main workflow runs `make setup` and it runs `go install` for tools whose versions are described in `versions.mk`. I don't describe it as a dependency, but no performance regression has appeared. So, I simply deleted our cache code.

daichimukai (approved):
LGTM

pluser (approved):

PR #860, opened 2024-03-19
fix: cleanup metrics state management in runner

jakobmoellerdev:
This properly deregisters and registers metrics based on the runner of the metrics exporter. This allows to stop the manager runnable and start it at a later time again (e.g. for in-process restarting). This will not affect existing behavior and just allows proper restarting.

All Unregister limitations apply:

> Note that even after unregistering, it will not be possible to register a new Collector that is inconsistent with the unregistered Collector, e.g. a Collector collecting metrics with the same name but a different help string. The rationale here is that the same registry instance must only collect consistent metrics throughout its lifetime

### DIFF
--- a/internal/runners/metrics_exporter.go
+++ b/internal/runners/metrics_exporter.go
@@ -63,7 +63,6 @@ var _ manager.LeaderElectionRunnable = &metricsExporter{}
 // NewMetricsExporter creates controller-runtime's manager.Runnable to run
 // a metrics exporter for a node.
 func NewMetricsExporter(vgServiceClient proto.VGServiceClient, client client.Client, nodeName string) manager.Runnable {
-
 	// metrics available under volumegroup subsystem
 	availableBytes := prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace:   metricsNamespace,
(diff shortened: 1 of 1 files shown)


internal/runners/metrics_exporter.go
```diff
@@ -135,13 +128,44 @@ func NewMetricsExporter(vgServiceClient proto.VGServiceClient, client client.Cli
...
+		}
+	}
+	return nil
+}
+
+func (m *metricsExporter) UnregisterAll() {
+	for _, c := range m.GetCollectors() {
+		metrics.Registry.Unregister(c)
+	}
+}
```
toshipp, line 155: I prefer to make these methods unexported because they are used only by Start.

llamerada-jp:
Is this change to address memory (or some resources) leaks? If so, I think I have to change other products (like pvc-autoresizer).  I would be grateful if you could tell me.

jakobmoellerdev:
@llamerada-jp no leak potential the only thing that won't work is restarting the affected runner in process (any case where Register is called twice before calling Unregister). No other side effects AFAIK.

llamerada-jp (approved):

Issue #861, opened 2024-03-22
In rare cases, deletion of LogicalVolume CR fails


llamerada-jp:
**Describe the bug**

If topolvm-controller or topolvm-node is restarted while deleting a volume, the LogicalVolume CR may remain even though the volume has been deleted.

**Environments**
- Version: v0.27.0
- OS:

**To Reproduce**

Restart topolvm-controller and topolvm-node at the time of PVC removal. 
If it occurs, the following LogicalVolume remains.

```yaml
$ kubectl get logicalvolumes.topolvm.cybozu.com pvc-79883466-cad8-4883-a2a8-12be06cd9589 -o yaml
apiVersion: topolvm.cybozu.com/v1
kind: LogicalVolume
metadata:
  creationTimestamp: "2024-03-21T01:29:39Z"
  deletionGracePeriodSeconds: 0
  deletionTimestamp: "2024-03-21T01:29:42Z"
  finalizers:
  - topolvm.cybozu.com/logicalvolume
  generation: 2
  labels:
    app.kubernetes.io/created-by: topolvm-controller
  name: pvc-79883466-cad8-4883-a2a8-12be06cd9589
  resourceVersion: "7113467672"
  uid: 555eeb7e-94d0-4898-ad53-b66199cf6c2e
spec:
  name: pvc-79883466-cad8-4883-a2a8-12be06cd9589
  nodeName: 10.69.5.82
  size: 100Mi
status:
  code: 13
  message: failed to check volume existence
```

However, volume does not exist.

```console
$ sudo lvs | grep 555eeb7e-94d0-4898-ad53-b66199cf6c2e
<empty>
```

**Expected behavior**

LogicalVolume and PV should be deleted without error.

**Additional context**

The delete operation should succeed if the volume already does not exist. However, the following `if` block seems wrong. There are two different ErrNotFound definitions, and the evaluation may be incorrect.

https://github.com/topolvm/topolvm/blob/2597dc75821009230af3be4067d377eef4bbe0a4/internal/lvmd/lvservice.go#L169

PR #863, opened 2024-03-22
fix: use correct ErrNotFound from command package

jakobmoellerdev:
addresses the issue reported by #861 where the wrong ErrNotFound Code was used.

### DIFF
--- a/internal/lvmd/device_class_manager.go
+++ b/internal/lvmd/device_class_manager.go
@@ -9,8 +9,8 @@ import (
 	lvmdTypes "github.com/topolvm/topolvm/pkg/lvmd/types"
 )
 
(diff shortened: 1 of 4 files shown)


ushitora-anqou (changes requested):
Thank you for fixing this issue!

- Could you update [this line](https://github.com/topolvm/topolvm/blob/2597dc75821009230af3be4067d377eef4bbe0a4/internal/lvmd/lvservice.go#L162) too?
- To avoid similar bugs in the future, I'd like to rename one of the definitions. Could you rewrite `lvmd.ErrNotFound` to `lvmd.ErrDeviceClassNotFound`? I think it is much clearer.

ushitora-anqou (approved):
LGTM

pluser (approved):

PR #865, opened 2024-03-22
fix: pin setup-envtest to go 1.20 before TopoLVM migrates to go 1.21

### LINKED ISSUE #864: Test Regression due to go.mod `go x.y.z`  directive causing errors in go 1.20
**Describe the bug**

When running tests with the go setup github action, it uses the go version from the go.mod file. This version is currently 1.20. 1.20 does not use the new syntax for versioning yet so now we have an issue in a package due to this:

`go: sigs.k8s.io/controller-runtime/tools/setup-envtest@latest (in sigs.k8s.io/controller-runtime/tools/setup-envtest@v0.0.0-20240322105421-affb96708000): go.mod:3: invalid go version '1.22.0': must match format 1.23`

**Environments**
- Version: Go 1.20, Latest Setup-Envtest which introduces go z-stream tag
- OS: All

**To Reproduce**
1. Use Go 1.20 locally
2. Run make tools

**Expected behavior**

We should be able to build this without problems.

**Additional context**

I think we have a few options here. One would be to use a newer Go runtime for testing but that will change the tested go version. Another is to pin setup-envtest until we upgrade ourselves.

jakobmoellerdev:
Attempts to temporarily fix #864 by pinning envtest to a version that didnt have the x.y.z syntax

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -269,8 +269,7 @@ install-helm-docs: | $(BINDIR)
 tools: install-kind install-container-structure-test install-helm install-helm-docs | $(BINDIR) ## Install development tools.
 	GOBIN=$(BINDIR) go install honnef.co/go/tools/cmd/staticcheck@latest
 	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(shell dirname $(GOLANGCI_LINT)) $(GOLANGCI_LINT_VERSION)
-	# Follow the official documentation to install the `latest` version, because explicitly specifying the version will get an error.
-	GOBIN=$(BINDIR) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
+	GOBIN=$(BINDIR) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@$(ENVTEST_VERSION)
 	GOBIN=$(BINDIR) go install sigs.k8s.io/controller-tools/cmd/controller-gen@v$(CONTROLLER_TOOLS_VERSION)
 
 	$(CURL) -o protoc.zip https://github.com/protocolbuffers/protobuf/releases/download/v$(PROTOC_VERSION)/protoc-$(PROTOC_VERSION)-linux-x86_64.zip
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -270,7 +270,7 @@ tools: install-kind install-container-structure-test install-helm install-helm-d
 	GOBIN=$(BINDIR) go install honnef.co/go/tools/cmd/staticcheck@latest
 	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(shell dirname $(GOLANGCI_LINT)) $(GOLANGCI_LINT_VERSION)
 	# Follow the official documentation to install the `latest` version, because explicitly specifying the version will get an error.
```
daichimukai, line 272: @llamerada-jp 
According to `git blame` you had wrote this comment. Do you remember what's meant by "explicitly specifying the version will get an error"? Is it safe to change this to specify concrete version?

daichimukai: Understood. Thank you.

Makefile
```diff
@@ -270,7 +270,7 @@ tools: install-kind install-container-structure-test install-helm install-helm-d
...
-	GOBIN=$(BINDIR) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
+	GOBIN=$(BINDIR) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@bf15e44028f908c790721fc8fe67c7bf2d06a611
```
daichimukai, line 273: To improve dependency upgrade task efficiency, we keep dependency versions in the `versions.mk` file. Please add this version to there.

Makefile
daichimukai: Let's remove this comment. We cannot use the "latest" for a while, and I don't see any reason to avoid specifying hash as long as it works.

daichimukai (changes requested):

llamerada-jp (approved):

daichimukai (approved):
LGTM. Thanks @jakobmoellerdev.

PR #878, opened 2024-04-08
chore: use exposed controller functions in topolvm controllers

jakobmoellerdev:
In retrospect of https://github.com/topolvm/topolvm/pull/875 I noticed that everytime we add a field, we also have to export all embedded fields with it to work correctly outside of the topolvm package. This makes going back and forth for every field extremely tedious.

I thus changed the current command runners in TopoLVM to use the pkg variants of the runners. This will not change any logic (we were already using the same logic just with internal imports) but will let us know via compiler error whenever we have a field change that needs to be exported as well.

### DIFF
--- a/cmd/topolvm-node/app/run.go
+++ b/cmd/topolvm-node/app/run.go
@@ -16,11 +16,10 @@ import (
 	topolvmlegacyv1 "github.com/topolvm/topolvm/api/legacy/v1"
 	topolvmv1 "github.com/topolvm/topolvm/api/v1"
 	clientwrapper "github.com/topolvm/topolvm/internal/client"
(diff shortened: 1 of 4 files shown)


cmd/topolvm-node/app/run.go
```diff
@@ -89,13 +89,17 @@ func subMain(ctx context.Context) error {
 	var health grpc_health_v1.HealthClient
 
 	if config.embedLvmd {
+		lvmd.Containerized(true)
 		command.Containerized = true
```
toshipp, line 93: This can be deleted.

toshipp (changes requested):

daichimukai (approved):

toshipp (approved):

Issue #876, opened 2024-04-05
RFE: moving btrfs from experimental to official Support


jakobmoellerdev:
**What should the feature do:**

The feature aims to transition the support status of btrfs within TopoLVM from 'experimental' to 'officially supported'. This involves ensuring full compatibility with btrfs, rigorous testing under various scenarios specific to btrfs filesystems, and documentation updates to reflect this support. Additionally, it should address any known issues or limitations that were previously tagged as 'experimental'.

**What is use case behind this feature:**

Btrfs, being a filesystem that offers advanced features like snapshotting, compression, and subvolume management, is becoming increasingly popular among Linux users, especially in environments where data integrity and flexibility are paramount. For users and organizations that have adopted btrfs as their filesystem of choice, having official support from TopoLVM means they can fully leverage the features and performance optimizations of btrfs in their storage volumes managed by TopoLVM. This transition also signifies a commitment to the broader storage and filesystem community, showcasing TopoLVM's dedication to supporting a wide array of storage technologies and configurations. By moving btrfs to official support status, TopoLVM can attract a wider user base, foster community contributions, and enhance its utility in diverse environments.

jakobmoellerdev:
I agree. Ultimately I think we should think what we have for ext4/xfs. I believe in our current E2E tests we only test mainly with ext4 and do not have a big matrix for xfs. We can simply extend the matrix to test all E2E scenarios with both xfs and ext4.

jakobmoellerdev:
I am mainly inquiring if this is something we want in the end and if it will be accepted to introduce more vectors like this in the test matrix (as this will increase the test amount significantly). If it is something we desire, I believe I can accomodate that.

ushitora-anqou:
Thanks for your description. We'll be happy to merge this enhancement if you submit a PR.

> We can simply extend the matrix to test all E2E scenarios with both xfs and ext4.

My (and my colleagues') concern is that it's not acceptable for the changes to make the E2E tests and CI much slower than they are now, since they are already relatively slow. Extending the E2E tests to the same extent as the current tests for xfs would be the best.

ushitora-anqou:
I agree with you. It seems that the current E2E tests don't enable ginkgo's `-p` option, so it might be worth a try:
https://github.com/topolvm/topolvm/blob/567532964002c8c9f8f54e1824010269f71e0413/test/e2e/Makefile#L16
https://onsi.github.io/ginkgo/#spec-parallelization

[...]

PR #879, opened 2024-04-09
fix: btrfs unmounting + e2e tests

### LINKED ISSUE #877: topolvm-node does not unmount btrfs volumes after running `parted -l`
**Describe the bug**

Looks like `parted -l` triggers udev rules that rename topolvm devices (`/dev/topolvm/c168dad9-978b-40e4-a45b-b77e141bff7a` -> `/dev/mapper/topolvm-c168dad9--978b--40e4--a45b--b77e141bff7a`). This is not an issue on the host system, but inside topolvm-node container there are no symlinks under `/dev/mapper/`, so the topolvm-node thinks the device is not mounted.

**Environments**

- Version: 0.27.0 (installed via Helm chart).
- OS: Ubuntu 22.04 (with 5.15 kernel).
- Kubernetes version: 1.28.8.

**To Reproduce**

1. Create a StorageClass with `fsType: btrfs`.
2. Create a Pod with generic ephemeral volume. Something like this:
   ```yaml
    apiVersion: v1
    kind: Pod
    metadata:
      name: test-pod
    spec:
      containers:
      - command:
        - sleep
        - infinity
        image: ubuntu:22.04
        name: test-container
        volumeMounts:
        - mountPath: /mnt/ephemeral-pv
          name: ephemeral-pv
      volumes:
      - ephemeral:
          volumeClaimTemplate:
            spec:
              accessModes:
              - ReadWriteOnce
              resources:
                requests:
                  storage: 1Gi
              storageClassName: topolvm-btrfs
              volumeMode: Filesystem
        name: ephemeral-pv
    ```
3. Wait until the Pod will be scheduled and ready.
4. Check `/proc/mounts` on the node:
    ```
    # cat /proc/mounts | grep pvc-
    /dev/topolvm/c168dad9-978b-40e4-a45b-b77e141bff7a /var/lib/kubelet/pods/a50597ec-22df-4c28-9f1b-6e5f7660499d/volumes/kubernetes.io~csi/pvc-40dfde50-d302-47d3-ba97-bebc4d003857/mount btrfs rw,relatime,space_cache=v2,subvolid=5,subvol=/ 0 0
    ```
5. Call `parted -l` on the node.
    ```
    # parted -l >/dev/null 2>&1
    ```
5. Check `/proc/mounts` on the node again (note that the device path changed from `/dev/topolvm/` to `/dev/mapper/`):
    ```
    # cat /proc/mounts | grep pvc-
    /dev/mapper/topolvm-c168dad9--978b--40e4--a45b--b77e141bff7a /var/lib/kubelet/pods/a50597ec-22df-4c28-9f1b-6e5f7660499d/volumes/kubernetes.io~csi/pvc-40dfde50-d302-47d3-ba97-bebc4d003857/mount btrfs rw,relatime,space_cache=v2,subvolid=5,subvol=/ 0 0
    ```
6. Delete the Pod.
7. Verify the Pod stuck in Terminating state.
8. Check topolvm-node logs on the node. You will see something like this:
    ```
    {"level":"info","ts":"2024-04-05T14:33:23Z","logger":"driver.node","msg":"NodeUnpublishVolume called","volume_id":"c168dad9-978b-40e4-a45b-b77e141bff7a","target_path":"/var/lib/kubelet/pods/a50597ec-22df-4c28-9f1b-6e5f7660499d/volumes/kubernetes.io~csi/pvc-40dfde50-d302-47d3-ba97-bebc4d003857/mount"}
    {"level":"error","ts":"2024-04-05T14:33:23Z","msg":"error on grpc call","method":"/csi.v1.Node/NodeUnpublishVolume","error":"rpc error: code = Internal desc = remove dir failed for /var/lib/kubelet/pods/a50597ec-22df-4c28-9f1b-6e5f7660499d/volumes/kubernetes.io~csi/pvc-40dfde50-d302-47d3-ba97-bebc4d003857/mount: error=unlinkat /var/lib/kubelet/pods/a50597ec-22df-4c28-9f1b-6e5f7660499d/volumes/kubernetes.io~csi/pvc-40dfde50-d302-47d3-ba97-bebc4d003857/mount: device or resource busy","stacktrace":"github.com/topolvm/topolvm/cmd/topolvm-node/app.ErrorLoggingInterceptor\n\t/workdir/cmd/topolvm-node/app/run.go:186\ngithub.com/container-storage-interface/spec/lib/go/csi._Node_NodeUnpublishVolume_Handler\n\t/go/pkg/mod/github.com/container-storage-interface/spec@v1.6.0/lib/go/csi/csi.pb.go:6168\ngoogle.golang.org/grpc.(*Server).processUnaryRPC\n\t/go/pkg/mod/google.golang.org/grpc@v1.58.3/server.go:1374\ngoogle.golang.org/grpc.(*Server).handleStream\n\t/go/pkg/mod/google.golang.org/grpc@v1.58.3/server.go:1751\ngoogle.golang.org/grpc.(*Server).serveStreams.func1.1\n\t/go/pkg/mod/google.golang.org/grpc@v1.58.3/server.go:986"}
    ```

**Expected behavior**

Pod should not stuck in Terminating state.

**Additional context**

To recover you can do something from the following list:

1. Create symlinks in `/dev/mapper/` and `/dev/topolvm/` inside topolvm-node container (same as on the host).
2. Manually unmount PV on the node (path can be found in topolvm-node's logs).
3. Reboot the node.

I can't reproduce it with ext4. Only with btrfs.

jakobmoellerdev:
fix #877 
partially address #876 

Adds table driven tests for the xfs scenarios and runs them for a btrfs storageClass as well. also adds a minimum size for btrfs

### DIFF
--- a/cmd/topolvm-controller/app/root.go
+++ b/cmd/topolvm-controller/app/root.go
@@ -25,6 +25,9 @@ const (
 	// Derived from the usual 4096K blocks, 1024 inode default and journaling overhead,
 	// Allows for more than 80% free space after formatting, anything lower significantly reduces this percentage.
 	DefaultMinimumAllocationSizeExt4 = "32Mi"
+	// DefaultMinimumAllocationSizeBtrfs is the default minimum size for a filesystem volume with btrfs formatting.
+	// Btrfs minimum data block size is 4KiB and by default the metadata block size is 16 KiB.
+	DefaultMinimumAllocationSizeBtrfs = "32Mi"
 )
 
 var config struct {
@@ -92,8 +95,9 @@ func init() {
 		"Minimum Allocation Sizing for block storage. Logical Volumes will always be at least this big.")
 	config.controllerServerSettings.MinimumAllocationSettings.Filesystem = make(map[string]driver.Quantity)
 	for filesystem, minimum := range map[string]resource.Quantity{
-		"ext4": resource.MustParse(DefaultMinimumAllocationSizeExt4),
-		"xfs":  resource.MustParse(DefaultMinimumAllocationSizeXFS),
+		"ext4":  resource.MustParse(DefaultMinimumAllocationSizeExt4),
+		"xfs":   resource.MustParse(DefaultMinimumAllocationSizeXFS),
+		"btrfs": resource.MustParse(DefaultMinimumAllocationSizeBtrfs),
 	} {
 		config.controllerServerSettings.MinimumAllocationSettings.Filesystem[filesystem] = driver.NewQuantityFlagVar(fs,
 			fmt.Sprintf("minimum-allocation-%s", filesystem),
--- a/cmd/topolvm-controller/app/root.go
+++ b/cmd/topolvm-controller/app/root.go
@@ -26,8 +26,8 @@ const (
 	// Allows for more than 80% free space after formatting, anything lower significantly reduces this percentage.
 	DefaultMinimumAllocationSizeExt4 = "32Mi"
 	// DefaultMinimumAllocationSizeBtrfs is the default minimum size for a filesystem volume with btrfs formatting.
-	// Btrfs minimum data block size is 4KiB and by default the metadata block size is 16 KiB.
-	DefaultMinimumAllocationSizeBtrfs = "32Mi"
+	// Btrfs minimum data block size is 4KiB, and by default, the metadata block size is 16 KiB.
+	DefaultMinimumAllocationSizeBtrfs = "200Mi"
 )
 
 var config struct {
--- a/cmd/topolvm-controller/app/root.go
+++ b/cmd/topolvm-controller/app/root.go
@@ -26,7 +26,8 @@ const (
 	// Allows for more than 80% free space after formatting, anything lower significantly reduces this percentage.
 	DefaultMinimumAllocationSizeExt4 = "32Mi"
 	// DefaultMinimumAllocationSizeBtrfs is the default minimum size for a filesystem volume with btrfs formatting.
-	// Btrfs minimum data block size is 4KiB, and by default, the metadata block size is 16 KiB.
+	// Btrfs changes its minimum allocation size based on various underlying device block settings and the host OS,
+	// but 200Mi seemed to be safe after some experimentation.
 	DefaultMinimumAllocationSizeBtrfs = "200Mi"
 )
 
--- a/internal/driver/node.go
+++ b/internal/driver/node.go
@@ -230,7 +230,7 @@ func (s *nodeServerNoLocked) nodePublishFilesystemVolume(req *csi.NodePublishVol
 		return status.Errorf(codes.Internal, "target device is already formatted with different filesystem: volume=%s, current=%s, new:%s", req.GetVolumeId(), fsType, mountOption.FsType)
 	}
 
-	mounted, err := filesystem.IsMounted(device, req.GetTargetPath())
+	mounted, err := filesystem.IsMounted(req.GetTargetPath())
 	if err != nil {
 		return status.Errorf(codes.Internal, "mount check failed: target=%s, error=%v", req.GetTargetPath(), err)
 	}
@@ -363,22 +363,11 @@ func (s *nodeServerNoLocked) NodeUnpublishVolume(ctx context.Context, req *csi.N
 func (s *nodeServerNoLocked) nodeUnpublishFilesystemVolume(req *csi.NodeUnpublishVolumeRequest, device string) error {
 	targetPath := req.GetTargetPath()
 
-	mounted, err := filesystem.IsMounted(device, targetPath)
-	if err != nil {
-		return status.Errorf(codes.Internal, "mount check failed: target=%s, error=%v", targetPath, err)
-	}
-	if mounted {
-		if err := s.mounter.Unmount(targetPath); err != nil {
-			return status.Errorf(codes.Internal, "unmount failed for %s: error=%v", targetPath, err)
-		}
+	if err := mountutil.CleanupMountPoint(targetPath, s.mounter, true); err != nil {
+		return status.Errorf(codes.Internal, "unmount failed for %s: error=%v", targetPath, err)
 	}
 
-	if err := os.RemoveAll(targetPath); err != nil {
-		return status.Errorf(codes.Internal, "remove dir failed for %s: error=%v", targetPath, err)
-	}
-
-	err = os.Remove(device)
-	if err != nil && !os.IsNotExist(err) {
+	if err := os.Remove(device); err != nil && !os.IsNotExist(err) {
 		return status.Errorf(codes.Internal, "remove device failed for %s: error=%v", device, err)
 	}
 
--- a/internal/driver/node.go
+++ b/internal/driver/node.go
@@ -230,7 +230,7 @@ func (s *nodeServerNoLocked) nodePublishFilesystemVolume(req *csi.NodePublishVol
 		return status.Errorf(codes.Internal, "target device is already formatted with different filesystem: volume=%s, current=%s, new:%s", req.GetVolumeId(), fsType, mountOption.FsType)
 	}
 
-	mounted, err := filesystem.IsMounted(req.GetTargetPath())
+	mounted, err := s.mounter.IsMountPoint(req.GetTargetPath())
 	if err != nil {
 		return status.Errorf(codes.Internal, "mount check failed: target=%s, error=%v", req.GetTargetPath(), err)
 	}
--- a/internal/filesystem/util.go
+++ b/internal/filesystem/util.go
@@ -12,11 +12,9 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
-	"path/filepath"
 	"strings"
 
 	"golang.org/x/sys/unix"
-	"k8s.io/mount-utils"
 )
 
 const (
@@ -25,36 +23,6 @@ const (
 
 type temporaryer interface {
 	Temporary() bool
-}
-
-// IsMounted returns true if device is mounted on target.
-// The implementation uses /proc/1/mountinfo because some filesystem uses a virtual device.
-func IsMounted(target string) (bool, error) {
-	abs, err := filepath.Abs(target)
-	if err != nil {
-		return false, err
-	}
-	target, err = filepath.EvalSymlinks(abs)
-	if err != nil {
-		return false, err
-	}
-
-	data, err := mount.ParseMountInfo("/proc/1/mountinfo")
-	if err != nil {
-		return false, fmt.Errorf("could not read /proc/1/mountinfo: %v", err)
-	}
-
-	for _, line := range data {
-		if line.MountPoint == target {
-			return true, nil
-		}
-
-		if d, err := filepath.EvalSymlinks(line.MountPoint); err == nil && d == target {
-			return true, nil
-		}
-	}
-
-	return false, nil
 }
 
 // DetectFilesystem returns filesystem type if device has a filesystem.
--- a/internal/filesystem/util.go
+++ b/internal/filesystem/util.go
@@ -30,29 +30,12 @@ type temporaryer interface {
 	Temporary() bool
 }
 
-func isSameDevice(dev1, dev2 string) (bool, error) {
-	if dev1 == dev2 {
-		return true, nil
-	}
+func isDeviceMapperDevice(device string) bool {
+	return strings.HasPrefix(device, "/dev/mapper/") || strings.HasPrefix(device, "/dev/dm-")
+}
 
-	var st1, st2 unix.Stat_t
-	if err := Stat(dev1, &st1); err != nil {
-		// Some filesystems like tmpfs and nfs aren't backed by block device files.
-		// In such case, given device path does not exist,
-		// we regard it is not an error but is false.
-		if os.IsNotExist(err) {
-			return false, nil
-		}
-		return false, fmt.Errorf("stat failed for %s: %v", dev1, err)
-	}
-	if err := Stat(dev2, &st2); err != nil {
-		if os.IsNotExist(err) {
-			return false, nil
-		}
-		return false, fmt.Errorf("stat failed for %s: %v", dev2, err)
-	}
-
-	return st1.Rdev == st2.Rdev, nil
+func majorMinorMatch(a, b unix.Stat_t) bool {
+	return unix.Major(a.Dev) == unix.Major(b.Dev) && unix.Minor(a.Dev) == unix.Minor(b.Dev)
 }
 
 // IsMounted returns true if device is mounted on target.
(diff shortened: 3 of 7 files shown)


jakobmoellerdev:
Looking at the test failures it seems the mount point is not set as expected with btrfs. Anyone has any input on that?

llamerada-jp:
@cupnes 
Is it the same as following issue?
https://github.com/topolvm/topolvm/issues/877

cupnes:
@llamerada-jp 
I think the situation is different from #877. What is failing now is [the pod deploy](https://github.com/topolvm/topolvm/blob/b0a2229dc10d170d31ce783c34958a183afe9836/test/e2e/e2e_test.go#L249). However, #877 is an issue where the pod cannot be deleted.

jakobmoellerdev:
@cupnes From what I can see in the test logs, the device verification fails: https://github.com/topolvm/topolvm/actions/runs/8615413549/job/23610904469?pr=879#step:9:570 Im not sure if this is really a failure as early as pod deployment.

cupnes:
@jakobmoellerdev Yes. Please wait a moment while I investigate.

cupnes:
@jakobmoellerdev 
I changed DefaultMinimumAllocationSizeBtrfs to 110MiB and the problem area passed.

By my investigation, it seems that the minimum device size for btrfs changes dynamically [1]. So, I do not think it is necessary to support a strict minimum allocation size in this PR. We can just write it in the limitation.md for now, with another PR to deal with it if necessary.
[1] https://patchwork.kernel.org/project/linux-btrfs/patch/1403148338-6584-1-git-send-email-quwenruo@cn.fujitsu.com/

jakobmoellerdev:
After adjusting the tests it seems that resizing behaves weirdly. The desired/actual storage for btrfs is off after resizing. Any ideas on what that is caused by?

jakobmoellerdev:
seems that now the pod deletion in btrfs for offline resizing has an issue. On top of that the runtime of the tests is pretty much unbearable. We need to figure out how to fix that otherwise the tests will run 20minutes +

I really think the tests should be refactored...

jakobmoellerdev:
Seems NodeUnpublishVolume for btrfs is broken due to "device or resource busy":

```
 {"level":"info","ts":"2024-04-15T11:41:37Z","logger":"driver.node","msg":"NodeUnpublishVolume called","volume_id":"52e43313-4f6b-4bdf-b3df-973f666ae983","target_path":"/var/lib/kubelet/pods/3fb3f6af-89f0-4fc4-8007-6b8263048f13/volum \u2502
\u2502 es/kubernetes.io~csi/pvc-411430a5-75d7-4aab-ae38-657b31fbe109/mount"}                                                                                                                                                                    \u2502
\u2502 {"level":"error","ts":"2024-04-15T11:41:37Z","msg":"error on grpc call","method":"/csi.v1.Node/NodeUnpublishVolume","error":"rpc error: code = Internal desc = remove dir failed for /var/lib/kubelet/pods/3fb3f6af-89f0-4fc4-8007-6b826 \u2502
\u2502 3048f13/volumes/kubernetes.io~csi/pvc-411430a5-75d7-4aab-ae38-657b31fbe109/mount: error=unlinkat /var/lib/kubelet/pods/3fb3f6af-89f0-4fc4-8007-6b8263048f13/volumes/kubernetes.io~csi/pvc-411430a5-75d7-4aab-ae38-657b31fbe109/mount: de \u2502
\u2502 vice or resource busy","stacktrace":"github.com/topolvm/topolvm/cmd/topolvm-node/app.ErrorLoggingInterceptor\n\t/home/jmoller/topolvm/cmd/topolvm-node/app/run.go:193\ngithub.com/container-storage-interface/spec/lib/go/csi._Node_Node \u2502
\u2502 UnpublishVolume_Handler\n\t/home/jmoller/go/pkg/mod/github.com/container-storage-interface/spec@v1.6.0/lib/go/csi/csi.pb.go:6168\ngoogle.golang.org/grpc.(*Server).processUnaryRPC\n\t/home/jmoller/go/pkg/mod/google.golang.org/grpc@v1 \u2502
\u2502 .58.3/server.go:1374\ngoogle.golang.org/grpc.(*Server).handleStream\n\t/home/jmoller/go/pkg/mod/google.golang.org/grpc@v1.58.3/server.go:1751\ngoogle.golang.org/grpc.(*Server).serveStreams.func1.1\n\t/home/jmoller/go/pkg/mod/google. \u2502
\u2502 golang.org/grpc@v1.58.3/server.go:986"}
```

jakobmoellerdev:
Seems the problem here is the `umount` call does not go through. When debugging into the node and running `umount /var/lib/kubelet/pods/3fb3f6af-89f0-4fc4-8007-6b8263048f13/volumes/kubernetes.io~csi/pvc-411430a5-75d7-4aab-ae38-657b31fbe109/mount` manually then the test recovers

jakobmoellerdev:
Seems like we have a device mapper / mounting issue again. The reason this is not recovering is because for btrfs the following `/proc/mounts` exists:

`/dev/dm-3 /var/lib/kubelet/pods/2be71706-9a37-41fe-91c2-f21220f5b1be/volumes/kubernetes.io~csi/pvc-411430a5-75d7-4aab-ae38-657b31fbe109/mount btrfs rw,relatime,discard=async,space_cache=v2,subvolid=5,subvol=/ 0 0`

as you can see this is under `/dev/dm-3`, I assume this is the same root cause as https://github.com/topolvm/topolvm/issues/877

jakobmoellerdev:
@toshipp @cupnes In light of my recent investigation I have found that we are still using /proc/mounts instead of /proc/1/mountinfo (which would also include Maj,Min and some other neat features). Also when looking through the code I do not see why we would need our own unmount logic if there is a well written unmounter in mount-utils. Im sure there was a reason but could you explain this to me again? If we dont really have any reason to keep this around I would vote for changing this to the upstream unmounting process

toshipp:
@jakobmoellerdev 
mount-util has not been exported during the early development stage of topolvm. We replaced our mount logic with it but missed unmount part. So, I have no objection to using it.

I checked mountinfo and found that the major number is not accurate for lvm devices. It reports 0 but should be 253.

jakobmoellerdev:
Thanks for the info that helps. I now changed the IsMounted check to the upstream version as well which goes against /proc/mounts but has a few more safeguards for fast paths. Now all tests are running green and moun/unmount for all filesystems is working as intended.

cupnes:
@jakobmoellerdev 
Thank you for fixing #877 as well. As for the fix for #877, it looks good. I am on holiday tomorrow and will review the other parts next week.

> I do have one more concern: The test runtime. The resizing tests currently take 4 minutes alone. Just excluding the resizing tests would save 8 minutes right now...

It may not have a big effect, but reducing the PVC size during the resize test may reduce the test time. At the moment we are using several GiB of PVC during resizing tests, but I think a few hundred MiB would be sufficient.

internal/filesystem/util.go
```diff
@@ -174,56 +100,3 @@ func Statfs(path string, buf *unix.Statfs_t) (err error) {
...
-//
-// The bug was fixed by the kernel commit 9f6c61f96f2d97 (since Linux 5.8).
-// Alas, there is no better way to check if the bug is fixed other than to
-// rely on the kernel version returned by uname.
-//
-// Copied from
-// https://github.com/kubernetes/mount-utils/blob/6f4aae5a6ab58574cac605cdd48bf5c0862c047f/mount_helper_unix.go#L204C1-L250
-// *    LICENSE: http://www.apache.org/licenses/LICENSE-2.0
-// *    Copyright The Kubernetes Authors.
-func kernelHasMountinfoBug() bool {
```
cupnes, line 196: There is a comment at the beginning of this file about `kernelHasMountinfoBug()`. Could you update that as well?

cupnes (commented):
I will comment on the points I noticed. The review is not yet complete and will continue tomorrow.

internal/driver/node.go
```diff
@@ -363,22 +363,11 @@ func (s *nodeServerNoLocked) NodeUnpublishVolume(ctx context.Context, req *csi.N
...
-	}
-	if mounted {
-		if err := s.mounter.Unmount(targetPath); err != nil {
-			return status.Errorf(codes.Internal, "unmount failed for %s: error=%v", targetPath, err)
-		}
-	}
-
-	if err := os.RemoveAll(targetPath); err != nil {
-		return status.Errorf(codes.Internal, "remove dir failed for %s: error=%v", targetPath, err)
+	if err := mountutil.CleanupMountPoint(targetPath, s.mounter, true); err != nil {
```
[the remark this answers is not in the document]
jakobmoellerdev: Let me double check but I believe this was by mistake if this happened.

internal/filesystem/util.go
jakobmoellerdev: Good catch, will do.

internal/driver/node.go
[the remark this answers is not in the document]
jakobmoellerdev: The current behavior is 

```go
pathExists, pathErr := PathExists(mountPath)
	if !pathExists && pathErr == nil {
		klog.Warningf("Warning: mount cleanup skipped because path does not exist: %v", mountPath)
		return nil
	}
```
This means that if we have a non existing path we should be fine. Also If there is no link present we will run into another nil case that is covered with IsCorruptedMount. This should be covering the cases we had before.

cmd/topolvm-controller/app/root.go
```diff
@@ -25,6 +25,9 @@ const (
 	// Derived from the usual 4096K blocks, 1024 inode default and journaling overhead,
 	// Allows for more than 80% free space after formatting, anything lower significantly reduces this percentage.
 	DefaultMinimumAllocationSizeExt4 = "32Mi"
+	// DefaultMinimumAllocationSizeBtrfs is the default minimum size for a filesystem volume with btrfs formatting.
+	// Btrfs minimum data block size is 4KiB, and by default, the metadata block size is 16 KiB.
```
llamerada-jp, line 29: There is a difference between the comment and the default value. How about deleting the comment or describing it as an experimental result?

test/e2e/manifests/values/base.yaml
```diff
@@ -65,6 +73,14 @@ storageClasses:
       allowVolumeExpansion: true
       additionalParameters:
         '{{ include "topolvm.pluginName" . }}/device-class': "thin"
+  - name: topolvm-provisioner-thin-btrfs
```
llamerada-jp, line 76: This SC looks unused.

jakobmoellerdev: Good catch! Oversight on my part, I forgot to add snapshot tests for btrfs before.

cmd/topolvm-controller/app/root.go
jakobmoellerdev: changed as suggested, I liked that idea.

internal/driver/node.go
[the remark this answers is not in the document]
cupnes: Thank you for your explanation. I understood.

test/e2e/manifests/values/base.yaml
jakobmoellerdev: Tests added

jakobmoellerdev:
I was able to reduce the test runtime of one filesystem by about 2 minutes now thanks to some missing --now flags on deleting the pods which caused a 30 period grace wait time. The reason we are now taking quite a bit of time on the resize tests is actually related to the [check with df in the pod](https://github.com/topolvm/topolvm/pull/879/files#diff-d5d12b6ea07b0a78c9ad1987e1486c565ecf16839f77a93e91b09475cf6c0e99R789). TBH I have no idea why its taking so long and I will have to do some tracing on where its slow but im assuming it has to do with the async nature of CSI because the volume expansion on lvm was extremely quick.

cupnes (approved):
Thank you. LGTM

jakobmoellerdev:
I also want to give an update on my long running but interesting investigations on the long duration of the resizing and the impact on our test suite (6+ minutes). 

This is in fact caused by ControllerExpandVolume always specifying needing NodeExpandVolume https://github.com/topolvm/topolvm/blob/a967c95da14f80955332a00ebb258e319c6c39ac/internal/driver/controller.go#L690. 

Why does this matter? Because NodeExpandVolume is not immediately called, but instead queued by the kubelets Volume Manager. When investigating I found that we are in fact not delaying the call in TopoLVM at all, rather the kubelet is not calling NodeExpandVolume for minutes, causing the test to stall until the kubelet invokes the CSI call. When investigating deeper I found that https://github.com/kubernetes/enhancements/blob/0e4d5df19d396511fe41ed0860b0ab9b96f46a2d/keps/sig-storage/556-csi-volume-resizing/README.md#expansion-on-kubelet is the design reason for this. However note that in theory, [volume manager should update every 100ms](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/volumemanager/volume_manager.go#L62), which for some reason does not happen. I know that this is where its stuck because of the PVCs conditions while waiting in the df eventually check:

```
│   conditions:                                                                                                                                                │
│   - lastProbeTime: null                                                                                                                                      │
│     lastTransitionTime: "2024-04-23T14:45:33Z"                                                                                                               │
│     message: Waiting for user to (re-)start a pod to finish file system resize of                                                                            │
│       volume on node.                                                                                                                                        │
│     status: "True"                                                                                                                                           │
│     type: FileSystemResizePending       
```

 I still need to verify if there is not an issue in the kubelet that causes delays which we dont want but this is definetly the reason we see the delay, nothing in TopoLVM is working while we are in this state so there is nothing we can do to speed it up.

How could we fix this without fixing the kubelet delay? We always specify NodeExpandVolume because NodeExpandVolume is [our only Code where we resize the Filesystem](https://github.com/topolvm/topolvm/blob/a06750656ba4b9b31ed77205b2b4e3139d9be257/internal/driver/node.go#L518). This is because we do not implement NodeStageVolume and NodeUnstageVolume which could implement the mounting and unmounting seperately from NodePublish/NodeUnpublishvolume which are only run when the pod is scheduled. 

Now this is important because changing this would allow us to run the filesystem resize in the lv controllers resize method, without needing to call NodeExpansionRequired true and thus we would be able to significantly speed up mounting and resizing behavior.

cupnes:
@jakobmoellerdev 
I understood. But I was just wondering is that correct as a CSI specification? I think NodeStageVolume and NodeUnstageVolume are for mounting and unmounting volumes.

jakobmoellerdev:
I think that the main behavior of StageVolume is to prepare it for use by pods. There is no mention of this being intended just for mount/umount in https://github.com/container-storage-interface/spec/blob/master/spec.md#nodestagevolume afaik. Also Ive seen drivers where Stage/Unstage does the necessary work such as activation of volumes and then publish does the actual mount with read or readwrite.

jakobmoellerdev:
Okay this investigation aside, I believe this PR can get merged without issues. @llamerada-jp any concerns?

cupnes:
@jakobmoellerdev 
On the other hand, I think it is a good to report to Kubernetes that the kubelet is not calling NodeExpandVolume for minutes.

jakobmoellerdev:
Agreed, I will take care to put this to the k8s community. Just as a follow-up, in https://github.com/topolvm/topolvm/actions/runs/8829012081/job/24239136537?pr=898#step:9:621 you can see that when I implement a Controller based Resize without NodeExtendVolume, then the test is as fast as the others, further solidifying my findings

llamerada-jp:
Sorry for the delay in reviewing this, I have been busy with other tasks. Good for the most part, and I'm still checking on the details.

llamerada-jp (approved):

jakobmoellerdev:
I just noticed this got merged without a squash. Maybe we should introduce a Github Action to stop merges if there is not one squashed commit?

PR #881, opened 2024-04-10
chore: run ginkgo in parallel execution mode

### LINKED ISSUE #880: Start Parallelizing E2E Test Suite
## What

As discussed [here](https://github.com/topolvm/topolvm/issues/876#issuecomment-2046262759) we noticed that the Ginkgo test suite is not running with `-p` which would significantly speed up test processing. We should embrace parallelism here and try to concurrently run as many tests as possible as we have a lot of waiting on the PVC provisioning.

## How

Setup the `-p` flag with the ginkgo runner as per https://onsi.github.io/ginkgo/#spec-parallelization and then fix all regressions

## Checklist

- [ ] Test all functions

jakobmoellerdev:
fixes #880 

See https://onsi.github.io/ginkgo/#spec-parallelization


jakobmoellerdev:
Looking at the failures it becomes clear that the current e2e Test suite has a few flaws that prohibit us from easily incorporating parallel execution:

- The device classes are setup centrally and not per test so its impossible to decouple checks for available capacity in a parallel fashion (think of one test just succeeding while a second test is creating a lv which influences the capacity)
- The setup is written in a way where namespaces are created statically and then multiple tests are run within a namespace

Ideally we need to rewrite the tests in a way where the deviceClasses are dynamically allocated per test and we have a test that runs exactly once on that deviceClass. Then we can truly parallelize the tests.

I think this may be going a bit far so I want to collect some ideas/feedback here

jakobmoellerdev:
Looking deeper into the E2E suite, it might make sense to rewrite some of these tests from scratch and properly set them up (also not use things like kubectl for applying the resources, but the kube client)

ushitora-anqou:
It seems you are right. One of my colleagues told me that some tests expect multiple nodes, while others delete the nodes. Major code changes would be required to make this PR work.

jakobmoellerdev:
I think looking at this we should decide wether to rewrite the Test suite or to scrap the Issue. I would like to see the rewrite happen but I understand if we think that thats too big of an impact.

ushitora-anqou:
I'd like to suggest closing this issue for now. I think we can add the e2e tests for btrfs without much slowdown if we design the tests carefully. Parallelizing the tests is attractive, but we can deprioritize it.

PR #894, opened 2024-04-23
lvmd: don't call Close() for stdout

ushitora-anqou:
The Go's reference of `StdoutPipe` says:

> Wait will close the pipe after seeing the command exit, so most callers
> need not close the pipe themselves. It is thus incorrect to call Wait
> before all reads from the pipe have completed.

https://pkg.go.dev/os/exec#Cmd.StdoutPipe

So, io.ReadAll should be called instead of Close.

### DIFF
--- a/internal/lvmd/command/lvm_command.go
+++ b/internal/lvmd/command/lvm_command.go
@@ -95,11 +95,9 @@ type commandReadCloser struct {
 	stderr io.ReadCloser
 }
 
+// Close closes stdout and stderr and waits for the command to exit. Close
+// should not be called before all reads from stdout have completed.
 func (p commandReadCloser) Close() error {
-	if err := p.ReadCloser.Close(); err != nil {
-		return err
-	}
-
 	// Read the stderr output after the read has finished since we are sure by then the command must have run.
 	stderr, err := io.ReadAll(p.stderr)
 	if err != nil {


internal/lvmd/command/lvm_command.go
```diff
@@ -96,7 +96,7 @@ type commandReadCloser struct {
 }
 
 func (p commandReadCloser) Close() error {
-	if err := p.ReadCloser.Close(); err != nil {
+	if _, err := io.ReadAll(p); err != nil {
```
jakobmoellerdev, line 99: I would argue that this entire function call would not be necessary as per the documentation and a close error would show up in p.cmd.Wait(). What is the reason for you to choose io.ReadAll? 

That said, great discovery!

ushitora-anqou: You're right, since `callLVMInto` reads all the data in `commandReadClose` before calling `Close`. My concern is that that is currently a hidden precondition of `commandReadCloser.Close()`. If someone accidentally calls `Close` before reading all the data in the future, `Close` will get stuck. So I was inclined to add a redundant `io.ReadAll` here.

But it may be a matter of taste. We can just remove the function call, and add some comments to `Close()`. I'll update my code.

ushitora-anqou: Updated. Waiting for CI to pass.

jakobmoellerdev: Im thinking your reason is actually good behavior and we should avoid a hangup at all costs, even if we do a redundant call. Maybe we should use your version and add a small comment?

ushitora-anqou: After thinking about this for a while, I think your suggestion is better. I'll update my patch again.

ushitora-anqou: Well, after some investigation, I changed my mind (again). Now I believe the current pushed code should be alright. Here's why:

I figured out that I misunderstood the Go's reference of `StdoutPipe`. It doesn't mean that all the data of `stdout` should be read before calling `Wait`, but it does mean that there should be no running goroutines that are trying to read from `stdout`. As proof, the following code runs without any hangs:
```go
package main                                                                                                                   

import (
        "os/exec"
)

func main() {
        cmd := exec.Command("ls")
        _, err := cmd.StdoutPipe()
         if err != nil {
                 return
         }
         cmd.Start()
         cmd.Wait()
}
```
And also [this patch](https://groups.google.com/g/golang-checkins/c/1bJAiVRuo_w) I found while googling says:

> os/exec.Cmd.Wait closes the read end of os/exec.Cmd.StdoutPipe, meaning
that io.ReadAll can return fs.ErrClosed if the child exits too early,
allowing Wait to complete. The StdoutPipe docs already note this sharp
edge.

So in our code, `Close` should be called after all read attempts have finished. This is already guaranteed in the current code, and can be guaranteed in the future by checking the code comments for `Close`, I think.

jakobmoellerdev (approved):

toshipp (approved):

PR #898, opened 2024-04-24
feat: controller only expansion

### LINKED ISSUE #902: TopoLVM should not use the `/dev/topolvm` devices but use lvm2 lv output as source of truth
**Describe the bug**

TopoLVM currently uses `/dev/topolvm` as a base for all its created devices. This goes against the lvm2 usage recommendations:

> [...] Other software and scripts should use the /dev/VolumeGroupName/LogicalVolumeName format to reduce the chance of needing amendment when the software is updated. [...]

We currently do not use this recommendation and thus we have seen bugs occur that are caused by the mismatch of the devices in the Host and the devices in the topolvm-node and lvmd containers. (namely btrfs unmounting leading to a udev refresh)

What TopoLVM does currently is considered a userspace device rename and should be avoided at all costs if not synchronized via udev. We do not use or respect udev of the host system so we should not attempt to rename the devices naively.

This bug is problematic as we currently have live volumes in the wild that use /dev/topolvm so switching to the native lvm path is gonna be hard. However if we do not do this, we are at risk of getting broken everytime a udev rule is triggering a refresh of device mapper paths.

Additionally this bug leads to confusing behavior because we do not use the "path" variable that is returned by lvm2 when interacting with lvmd, causing the codebase to be unnecessarily complex. In addition in currently prohibits workflows like lvmd-based resizing of volume mounts as lvmd is not aware of the correct lvmd path.

**Environments**
- Version: All known
- OS: All known

**To Reproduce**
1. Create a PVC with any StorageClass by TopoLVM
2. Create a Pod consuming it, causing a Mount
3. Open a terminal on the host node and inspect /dev, observe separate entries for /dev/vgname/lvname and /dev/topolvm based on wether viewing the container FS or the Host FS
4. Trigger a udev rule for the device mapper that causes a rename based on /dev/mapper, observe it is not synchronized with the container and deleting the Pod/PVC fails. (this part will be fixed with our unmount behavior getting corrected in https://github.com/topolvm/topolvm/pull/879)

**Expected behavior**
I expect TopoLVM as the defacto most popular lvm CSI driver to use default device conventions for interacting with lvm2 and not to create own devices.

**Additional context**

jakobmoellerdev:
This change is prototyping how using the native LVM device paths reported by device mapper could work together with lvm controlled filesystem resizing in order to run controller-only online/offline volume expansion.

The main benefit of this is significantly faster resizing due to not having to wait for kubelet expansion. Additionally we no longer have to maintain our own devices in /dev/topolvm. (fix #902)

~Note that the current design breaks with existing TopoLVM because it switches from using the TopoLVM device directory to using the lvm owned device paths (so TopoLVM no longer renames the devices).~ Unmounting via lvm2 provided device paths always works even if legacy paths exist.

~I want to drive a POC here so that I can further debug and showcase how much more efficient we could be regarding dealing with device paths, lvm compatibility and the CSI spec.~ I think this approach is feasible and I would like to open it for reviews

Summary of Changes:

- Change all ControllerExpandVolumeResponses to NodeExpansionRequired: false
- Extend lvmd with resize behavior instead of NodeExpansion
- Allow lvmd to temporarily do a mount for resizing of PVCs under /tmp when there is currently no mount for the filesystem, otherwise reuse the mount present.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -176,11 +176,11 @@ image: image-normal image-with-sidecar ## Build topolvm images.
 
 .PHONY: image-normal
 image-normal:
-	docker buildx build --no-cache --load -t $(IMAGE_PREFIX)topolvm:devel --build-arg TOPOLVM_VERSION=$(TOPOLVM_VERSION) --target topolvm .
+	docker buildx build --load -t $(IMAGE_PREFIX)topolvm:devel --build-arg TOPOLVM_VERSION=$(TOPOLVM_VERSION) --target topolvm .
 
 .PHONY: image-with-sidecar
 image-with-sidecar:
-	docker buildx build --no-cache --load -t $(IMAGE_PREFIX)topolvm-with-sidecar:devel --build-arg TOPOLVM_VERSION=$(TOPOLVM_VERSION) --target topolvm-with-sidecar .
+	docker buildx build --load -t $(IMAGE_PREFIX)topolvm-with-sidecar:devel --build-arg TOPOLVM_VERSION=$(TOPOLVM_VERSION) --target topolvm-with-sidecar .
 
 .PHONY: container-structure-test
 container-structure-test: ## Run container-structure-test.
--- a/internal/driver/controller.go
+++ b/internal/driver/controller.go
@@ -689,6 +689,6 @@ func (s controllerServerNoLocked) ControllerExpandVolume(ctx context.Context, re
 
 	return &csi.ControllerExpandVolumeResponse{
 		CapacityBytes:         requestCapacityBytes,
-		NodeExpansionRequired: true,
+		NodeExpansionRequired: false,
 	}, nil
 }
--- a/internal/lvmd/command/lvm.go
+++ b/internal/lvmd/command/lvm.go
@@ -7,6 +7,7 @@ import (
 	"path"
 
 	"github.com/topolvm/topolvm"
+	"github.com/topolvm/topolvm/internal/filesystem"
 )
 
 // ErrNotFound is returned when a VG or LV is not found.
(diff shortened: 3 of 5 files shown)


toshipp:
Since I've not concluded avoiding NodeExpandVolume yet, I'd like to separate this PR into making expansion faster and not using `/dev/topolvm`.

jakobmoellerdev:
How would you be able to achieve that? I am combining these since you need a path in the Controller and I would have to migrate the entirety of /dev/topolvm device logic into the controller first otherwise

toshipp:
IIUC, all logic on `node.go` you wrote is valid if lvmd tells us lv device paths. So we could review the decommission of `/dev/topolvm` first.

jakobmoellerdev:
Understand what you mean now, will work on separating this out, might take a bit

jakobmoellerdev:
@toshipp @peng225 I have now extracted the device changes out into the recently merged PR and rebased on main. The only changes left are now the controller based expansion. PTAL and let me know what you think of it.

toshipp:
@jakobmoellerdev 
Thank you for your update. We'll dig into the problem, but it will take time, IMO. Please wait for a while.

toshipp:
@jakobmoellerdev 
I am investigating the slow expansion issue and suspect kubelet's reconcile mechanism.
I have found a way to speed up e2e testing. Would you like to speed up e2e tests? Or the real environment?

jakobmoellerdev:
The slow expand is because the kubelets desired state of world populator, but my issue is that we should not need to wait for the node if we already interact with lvmd, hence my PR. 

So my desire is to speed up both. Real life expands take as much as testing

toshipp:
Thanks.
I think it is not a good idea to abuse ControllerExpandVolume to resize a filesystem. I found that mutating a pod makes kubelet work faster, it might be a better solution. Maybe this should be solved by the kubernetes project. I am looking into it more.

jakobmoellerdev:
I think your concern is valid. Let me know if you need any input from me to proceed.

toshipp:
@jakobmoellerdev 
I tried to solve this problem by kubelet watching PVC, but the attempt failed because the authors who implemented online resizing of PV had decided not to watch PV/PVC directly, but to use the existing pod resync loop. see https://github.com/kubernetes/community/pull/1535#issuecomment-385146713.
So now I agree with the idea to solve it in topolvm. But as I mentioned before, I think the PR is not good because it abuses ControllerExpandVolume where CSI spec does not expect to resize a filesystem.
I tested another solution to notify kubelet by mutating pods on #956, it worked. I'd like to proceed with this, do you have any comments?

jakobmoellerdev:
`But as I mentioned before, I think the PR is not good because it abuses ControllerExpandVolume where CSI spec does not expect to resize a filesystem.`

I think this assumption is not quite correct. ControllerExpandVolume is abstracted from filesystem management. Thats why one can set NodeVolumeExpansionRequired to true.

Nevertheless I think for code simplicity I agree, we can move forward with "nudging" the pod to do what we want, as it maintains our code paths better. I suggest we add some documentation to that pr and close this one out

jakobmoellerdev:
Closing in favor of #956

PR #101, opened 2020-02-14
Implement Volume expansion functionality

tapih:
This PR implements https://github.com/cybozu-go/topolvm/issues/69.

PR #903, opened 2024-04-25
fix: avoid initial 1s delay and use exponential backoff on spec / status checks for LogicalVolume

### LINKED ISSUE #904: LogicalVolumeService updates wait for at least 1s and do not use backoffs
### Discussed in https://github.com/topolvm/topolvm/discussions/901

<div type='discussions-op-text'>

<sup>Originally posted by **guettli** April 25, 2024</sup>
I am looking at that code, and don't understand it:

```go
// updateSpecSize updates .Spec.Size of LogicalVolume.
func (s *LogicalVolumeService) updateSpecSize(ctx context.Context, volumeID string, size *resource.Quantity) error {
	for {
		select {
		case <-ctx.Done():
			return ctx.Err()
		case <-time.After(1 * time.Second):
		}
```

[source](https://github.com/topolvm/topolvm/blob/c29402b76847462d46ada71b790962dee1e405ce/internal/driver/internal/k8s/logicalvolume_service.go#L354-L355)

If a conflict-error happens, then waiting makes sense. But the code waits before the first attempt.

I guess you have your reasons to do that. Can you help me understand that better?</div>

jakobmoellerdev:
fix #904

### DIFF
--- a/internal/driver/internal/k8s/logicalvolume_service.go
+++ b/internal/driver/internal/k8s/logicalvolume_service.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math"
 	"time"
 
 	"github.com/topolvm/topolvm"
@@ -16,6 +17,8 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/retry"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
@@ -222,7 +225,7 @@ func (s *LogicalVolumeService) DeleteVolume(ctx context.Context, volumeID string
 
 	lv, err := s.GetVolume(ctx, volumeID)
 	if err != nil {
-		if err == ErrVolumeNotFound {
+		if errors.Is(err, ErrVolumeNotFound) {
 			logger.Info("volume is not found", "volume_id", volumeID)
 			return nil
 		}
@@ -238,23 +241,23 @@ func (s *LogicalVolumeService) DeleteVolume(ctx context.Context, volumeID string
 	}
 
 	// wait until delete the target volume
-	for {
-		logger.Info("waiting for delete LogicalVolume", "name", lv.Name)
-		select {
-		case <-ctx.Done():
-			return ctx.Err()
-		case <-time.After(100 * time.Millisecond):
-		}
-
-		err := s.getter.Get(ctx, client.ObjectKey{Name: lv.Name}, new(topolvmv1.LogicalVolume))
-		if err != nil {
+	return wait.Backoff{
+		Duration: 100 * time.Millisecond, // initial backoff
+		Factor:   2,                      // factor for duration increase
+		Jitter:   0.1,
+		Steps:    math.MaxInt, // run for infinity; we assume context gets canceled
+		Cap:      10 * time.Second,
+	}.DelayFunc().Until(ctx, true, false, func(ctx context.Context) (bool, error) {
+		if err := s.getter.Get(ctx, client.ObjectKey{Name: lv.Name}, new(topolvmv1.LogicalVolume)); err != nil {
 			if apierrors.IsNotFound(err) {
-				return nil
+				return true, nil
 			}
 			logger.Error(err, "failed to get LogicalVolume", "name", lv.Name)
-			return err
+			return false, err
 		}
-	}
+		logger.Info("waiting for LogicalVolume to be deleted", "name", lv.Name)
+		return false, nil
+	})
 }
 
 // CreateSnapshot creates a snapshot of existing volume.
@@ -301,48 +304,58 @@ func (s *LogicalVolumeService) CreateSnapshot(ctx context.Context, node, dc, sou
 
 // ExpandVolume expands volume
 func (s *LogicalVolumeService) ExpandVolume(ctx context.Context, volumeID string, requestBytes int64) error {
-	logger.Info("k8s.ExpandVolume called", "volumeID", volumeID, "size", requestBytes)
+	logger := logger.WithValues("volume_id", volumeID, "size", requestBytes)
+	logger.Info("k8s.ExpandVolume called")
+	request := resource.NewQuantity(requestBytes, resource.BinarySI)
 
 	lv, err := s.GetVolume(ctx, volumeID)
 	if err != nil {
 		return err
 	}
 
-	err = s.updateSpecSize(ctx, volumeID, resource.NewQuantity(requestBytes, resource.BinarySI))
+	err = s.updateSpecSize(ctx, volumeID, request)
 	if err != nil {
 		return err
 	}
 
-	// wait until topolvm-node expands the target volume
-	for {
-		logger.Info("waiting for update of 'status.currentSize'", "name", lv.Name)
-		select {
-		case <-ctx.Done():
-			return ctx.Err()
-		case <-time.After(1 * time.Second):
+	return wait.Backoff{
+		Duration: 1 * time.Second, // initial backoff
+		Factor:   2,               // factor for duration increase
+		Jitter:   0.1,
+		Steps:    math.MaxInt, // run for infinity; we assume context gets canceled
+		Cap:      10 * time.Second,
+	}.DelayFunc().Until(ctx, true, false, func(ctx context.Context) (bool, error) {
+		var changedLV topolvmv1.LogicalVolume
+		if err := s.getter.Get(ctx, client.ObjectKey{Name: lv.Name}, &changedLV); err != nil {
+			logger.Error(err, "failed to get LogicalVolume", "name", lv.Name)
+			return false, err
 		}
 
-		var changedLV topolvmv1.LogicalVolume
-		err := s.getter.Get(ctx, client.ObjectKey{Name: lv.Name}, &changedLV)
-		if err != nil {
-			logger.Error(err, "failed to get LogicalVolume", "name", lv.Name)
-			return err
-		}
 		if changedLV.Status.Code != codes.OK {
-			return status.Error(changedLV.Status.Code, changedLV.Status.Message)
+			return false, status.Error(changedLV.Status.Code, changedLV.Status.Message)
 		}
+
 		if changedLV.Status.CurrentSize == nil {
+			logger.Info("waiting for update of 'status.currentSize' "+
+				"to be filled initially", "name", lv.Name)
 			// WA: since Status.CurrentSize is added in v0.4.0. it may be missing.
 			// if the expansion is completed, it is filled, so wait for that.
-			continue
-		}
-		if changedLV.Status.CurrentSize.Value() != changedLV.Spec.Size.Value() {
-			logger.Info("failed to match current size and requested size", "current", changedLV.Status.CurrentSize.Value(), "requested", changedLV.Spec.Size.Value())
-			continue
+			return false, nil
 		}
 
-		return nil
-	}
+		if changedLV.Status.CurrentSize.Cmp(*request) != 0 {
+			logger.Info("waiting for update of 'status.currentSize' to be updated to signal successful expansion",
+				"name", lv.Name,
+				"status.currentSize", changedLV.Status.CurrentSize,
+				"spec.size", changedLV.Spec.Size,
+				"request", request,
+			)
+			return false, nil
+		}
+
+		logger.Info("LogicalVolume successfully expanded")
+		return true, nil
+	})
 }
 
 // GetVolume returns LogicalVolume by volume ID.
@@ -352,56 +365,51 @@ func (s *LogicalVolumeService) GetVolume(ctx context.Context, volumeID string) (
 
 // updateSpecSize updates .Spec.Size of LogicalVolume.
 func (s *LogicalVolumeService) updateSpecSize(ctx context.Context, volumeID string, size *resource.Quantity) error {
-	for {
-		select {
-		case <-ctx.Done():
-			return ctx.Err()
-		case <-time.After(1 * time.Second):
-		}
-
-		lv, err := s.GetVolume(ctx, volumeID)
-		if err != nil {
-			return err
-		}
-
-		lv.Spec.Size = *size
-		if lv.Annotations == nil {
-			lv.Annotations = make(map[string]string)
-		}
-		lv.Annotations[topolvm.GetResizeRequestedAtKey()] = time.Now().UTC().String()
-
-		if err := s.writer.Update(ctx, lv); err != nil {
-			if apierrors.IsConflict(err) {
-				logger.Info("detect conflict when LogicalVolume spec update", "name", lv.Name)
-				continue
+	return wait.ExponentialBackoffWithContext(ctx,
+		retry.DefaultBackoff,
+		func(ctx context.Context) (bool, error) {
+			lv, err := s.GetVolume(ctx, volumeID)
+			if err != nil {
+				return false, err
 			}
-			logger.Error(err, "failed to update LogicalVolume spec", "name", lv.Name)
-			return err
-		}
+			lv.Spec.Size = *size
+			if lv.Annotations == nil {
+				lv.Annotations = make(map[string]string)
+			}
+			lv.Annotations[topolvm.GetResizeRequestedAtKey()] = time.Now().UTC().String()
 
-		return nil
-	}
+			if err := s.writer.Update(ctx, lv); err != nil {
+				if apierrors.IsConflict(err) {
+					logger.Info("detected conflict when trying to update LogicalVolume spec", "name", lv.Name)
+					return false, nil
+				} else {
+					logger.Error(err, "failed to update LogicalVolume spec", "name", lv.Name)
+					return false, err
+				}
+			}
+			return true, nil
+		})
 }
 
 // waitForStatusUpdate waits for logical volume creation/failure/timeout, whichever comes first.
 func (s *LogicalVolumeService) waitForStatusUpdate(ctx context.Context, name string) (string, error) {
-	for {
-		logger.Info("waiting for setting 'status.volumeID'", "name", name)
-		select {
-		case <-ctx.Done():
-			return "", ctx.Err()
-		case <-time.After(100 * time.Millisecond):
-		}
-
+	var volumeID string
+	return volumeID, wait.Backoff{
+		Duration: 1 * time.Second, // initial backoff
+		Factor:   2,               // factor for duration increase
+		Jitter:   0.1,
+		Steps:    math.MaxInt, // run for infinity; we assume context gets canceled
+		Cap:      10 * time.Second,
+	}.DelayFunc().Until(ctx, true, false, func(ctx context.Context) (bool, error) {
 		var newLV topolvmv1.LogicalVolume
-		err := s.getter.Get(ctx, client.ObjectKey{Name: name}, &newLV)
-		if err != nil {
+		if err := s.getter.Get(ctx, client.ObjectKey{Name: name}, &newLV); err != nil {
 			logger.Error(err, "failed to get LogicalVolume", "name", name)
-			return "", err
+			return false, err
 		}
 		if newLV.Status.VolumeID != "" {
-			logger.Info("end k8s.LogicalVolume", "volume_id", newLV.Status.VolumeID)
-			return newLV.Status.VolumeID, nil
+			logger.Info("LogicalVolume successfully provisioned", "volume_id", newLV.Status.VolumeID)
+			volumeID = newLV.Status.VolumeID
+			return true, nil
 		}
 		if newLV.Status.Code != codes.OK {
 			err := s.writer.Delete(ctx, &newLV)
@@ -409,7 +417,8 @@ func (s *LogicalVolumeService) waitForStatusUpdate(ctx context.Context, name str
 				// log this error but do not return this error, because newLV.Status.Message is more important
 				logger.Error(err, "failed to delete LogicalVolume")
 			}
-			return "", status.Error(newLV.Status.Code, newLV.Status.Message)
+			return false, status.Error(newLV.Status.Code, newLV.Status.Message)
 		}
-	}
+		return false, nil
+	})
 }
--- a/internal/driver/node.go
+++ b/internal/driver/node.go
@@ -2,6 +2,7 @@ package driver
 
 import (
 	"context"
+	"errors"
 	"io"
 	"os"
 	"path"
@@ -454,13 +455,12 @@ func (s *nodeServerNoLocked) NodeGetVolumeStats(ctx context.Context, req *csi.No
 func (s *nodeServerNoLocked) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandVolumeRequest) (*csi.NodeExpandVolumeResponse, error) {
 	volumeID := req.GetVolumeId()
 	volumePath := req.GetVolumePath()
-
-	nodeLogger.Info("NodeExpandVolume is called",
-		"volume_id", volumeID,
+	logger := nodeLogger.WithValues("volume_id", volumeID,
 		"volume_path", volumePath,
 		"required", req.GetCapacityRange().GetRequiredBytes(),
-		"limit", req.GetCapacityRange().GetLimitBytes(),
-	)
+		"limit", req.GetCapacityRange().GetLimitBytes())
+
+	logger.Info("NodeExpandVolume is called")
 
 	if len(volumeID) == 0 {
 		return nil, status.Error(codes.InvalidArgument, "no volume_id is provided")
@@ -480,10 +480,7 @@ func (s *nodeServerNoLocked) NodeExpandVolume(ctx context.Context, req *csi.Node
 	}
 
 	if isBlock := req.GetVolumeCapability().GetBlock() != nil; isBlock {
-		nodeLogger.Info("NodeExpandVolume(block) is skipped",
-			"volume_id", volumeID,
-			"target_path", volumePath,
-		)
+		logger.Info("NodeExpandVolume(block) is skipped")
 		return &csi.NodeExpandVolumeResponse{}, nil
 	}
 
@@ -492,7 +489,7 @@ func (s *nodeServerNoLocked) NodeExpandVolume(ctx context.Context, req *csi.Node
 	deviceClass := topolvm.DefaultDeviceClassName
 	if err == nil {
 		deviceClass = lvr.Spec.DeviceClass
-	} else if err != k8s.ErrVolumeNotFound {
+	} else if !errors.Is(err, k8s.ErrVolumeNotFound) {
 		return nil, err
 	}
 	lv, err := s.getLvFromContext(ctx, deviceClass, volumeID)
@@ -517,16 +514,15 @@ func (s *nodeServerNoLocked) NodeExpandVolume(ctx context.Context, req *csi.Node
 	if len(devicePath) == 0 {
 		return nil, status.Errorf(codes.Internal, "filesystem %s is not mounted at %s", volumeID, volumePath)
 	}
+	logger = logger.WithValues("device", devicePath)
 
+	logger.Info("triggering filesystem resize")
 	r := mountutil.NewResizeFs(s.mounter.Exec)
 	if _, err := r.Resize(device, volumePath); err != nil {
 		return nil, status.Errorf(codes.Internal, "failed to resize filesystem %s (mounted at: %s): %v", volumeID, volumePath, err)
 	}
 
-	nodeLogger.Info("NodeExpandVolume(fs) is succeeded",
-		"volume_id", volumeID,
-		"target_path", volumePath,
-	)
+	logger.Info("NodeExpandVolume(fs) is succeeded")
 
 	// `capacity_bytes` in NodeExpandVolumeResponse is defined as OPTIONAL.
 	// If this field needs to be filled, the value should be equal to `.status.currentSize` of the corresponding
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -840,6 +840,18 @@ func testE2E() {
 			_, err = kubectlWithInput(claimYAML, "apply", "-n", ns, "-f", "-")
 			Expect(err).ShouldNot(HaveOccurred())
 
+			By("verifying PVC is updated")
+			Eventually(func() error {
+				var pvc corev1.PersistentVolumeClaim
+				if err := getObjects(&pvc, "pvc", "-n", ns, "topo-pvc"); err != nil {
+					return fmt.Errorf("failed to get PVC. err: %w", err)
+				}
+				if pvc.Status.Capacity.Storage().Cmp(*pvc.Spec.Resources.Requests.Storage()) != 0 {
+					return fmt.Errorf("capacity was not yet updated in status: %s", pvc.Status.Capacity.Storage().String())
+				}
+				return nil
+			}, timeout).Should(Succeed())
+
 			By("confirming that the specified device is resized in the Pod")
 			Eventually(func() error {
 				stdout, err := kubectl("exec", "-n", ns, "ubuntu", "--", "df", "--output=size", "/test1")
@@ -863,7 +875,17 @@ func testE2E() {
 				"reason=VolumeResizeFailed"
 			var events corev1.EventList
 			err = getObjects(&events, "events", "-n", ns, "--field-selector="+fieldSelector)
-			Expect(err).To(BeEquivalentTo(ErrObjectNotFound))
+			wasModified := "the object has been modified; please apply your changes to the latest version and try again"
+			if !errors.Is(err, ErrObjectNotFound) {
+				for _, item := range events.Items {
+					if strings.Contains(item.Message, wasModified) {
+						By("VolumeResizeFailed event was skipped, " +
+							"as it was caused by resizer resource version inconsistency")
+					} else {
+						Fail(fmt.Sprintf("unexpected VolumeResizeFailed event: %s", item.Message))
+					}
+				}
+			}
 
 			By("resizing PVC over vg capacity")
 			claimYAML = []byte(fmt.Sprintf(pvcTemplateYAML, "topo-pvc", "Filesystem", 100*1024, storageClass))
@@ -880,6 +902,16 @@ func testE2E() {
 				case err != nil:
 					return fmt.Errorf("failed to get event. err: %w", err)
 				default:
+					foundValid := false
+					for _, item := range events.Items {
+						if !strings.Contains(item.Message, wasModified) {
+							foundValid = true
+							break
+						}
+					}
+					if !foundValid {
+						return errors.New("no valid failure event found (filtered to not contain conflicts)")
+					}
 					return nil
 				}
 			}).Should(Succeed())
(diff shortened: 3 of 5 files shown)


internal/driver/internal/k8s/logicalvolume_service.go
```diff
@@ -313,36 +312,39 @@ func (s *LogicalVolumeService) ExpandVolume(ctx context.Context, volumeID string
...
 			// WA: since Status.CurrentSize is added in v0.4.0. it may be missing.
 			// if the expansion is completed, it is filled, so wait for that.
-			continue
+			return false, ErrMissingStatusCurrentSize
 		}
+
 		if changedLV.Status.CurrentSize.Value() != changedLV.Spec.Size.Value() {
 			logger.Info("failed to match current size and requested size", "current", changedLV.Status.CurrentSize.Value(), "requested", changedLV.Spec.Size.Value())
-			continue
+			logger.Info("waiting for update of 'status.currentSize'", "name", lv.Name)
```
toshipp, line 341: This message is nice for users. I think the line 340 can be deleted.

internal/driver/internal/k8s/logicalvolume_service.go
```diff
@@ -313,36 +312,39 @@ func (s *LogicalVolumeService) ExpandVolume(ctx context.Context, volumeID string
...
+		Cap:      300 * time.Second, // wait up to 10 seconds per retry, 1, 2, 4, 8, 10 seconds.
+		Steps:    300,               // run at most 10 times
+	}, func(ctx context.Context) (done bool, err error) {
 		var changedLV topolvmv1.LogicalVolume
-		err := s.getter.Get(ctx, client.ObjectKey{Name: lv.Name}, &changedLV)
-		if err != nil {
+		if err := s.getter.Get(ctx, client.ObjectKey{Name: lv.Name}, &changedLV); err != nil {
 			logger.Error(err, "failed to get LogicalVolume", "name", lv.Name)
-			return err
+			return true, err
```
toshipp, line 325:
```suggestion
			return false, err
```

internal/driver/internal/k8s/logicalvolume_service.go
```diff
@@ -313,36 +312,39 @@ func (s *LogicalVolumeService) ExpandVolume(ctx context.Context, volumeID string
...
 		}
+
 		if changedLV.Status.Code != codes.OK {
-			return status.Error(changedLV.Status.Code, changedLV.Status.Message)
+			return true, status.Error(changedLV.Status.Code, changedLV.Status.Message)
```
toshipp, line 329:
```suggestion
			return false, status.Error(changedLV.Status.Code, changedLV.Status.Message)
```

internal/driver/internal/k8s/logicalvolume_service.go
```diff
@@ -352,64 +354,60 @@ func (s *LogicalVolumeService) GetVolume(ctx context.Context, volumeID string) (
...
-		if err := s.writer.Update(ctx, lv); err != nil {
-			if apierrors.IsConflict(err) {
-				logger.Info("detect conflict when LogicalVolume spec update", "name", lv.Name)
-				continue
+	return wait.ExponentialBackoffWithContext(ctx,
+		retry.DefaultBackoff,
+		func(ctx context.Context) (done bool, err error) {
+			lv, err := s.GetVolume(ctx, volumeID)
+			if err != nil {
+				return true, err
```
toshipp, line 362:
```suggestion
				return false, err
```

internal/driver/internal/k8s/logicalvolume_service.go
```diff
@@ -352,64 +354,60 @@ func (s *LogicalVolumeService) GetVolume(ctx context.Context, volumeID string) (
...
+			lv.Spec.Size = *size
+			if lv.Annotations == nil {
+				lv.Annotations = make(map[string]string)
+			}
+			lv.Annotations[topolvm.GetResizeRequestedAtKey()] = time.Now().UTC().String()
+
+			if err := s.writer.Update(ctx, lv); err != nil {
+				if apierrors.IsConflict(err) {
+					logger.Info("detected conflict when trying to update LogicalVolume spec", "name", lv.Name)
+					return false, err
```
toshipp, line 373: Do we need to retry this? I assume ExponentialBackoffWithContext exits and returns an error because the err is not nil.

internal/driver/internal/k8s/logicalvolume_service.go
```diff
@@ -352,64 +354,60 @@ func (s *LogicalVolumeService) GetVolume(ctx context.Context, volumeID string) (
...
+				} else {
+					logger.Error(err, "failed to update LogicalVolume spec", "name", lv.Name)
+					return true, err
```
toshipp, line 376:
```suggestion
					return false, err
```

internal/driver/internal/k8s/logicalvolume_service.go
```diff
@@ -352,64 +354,60 @@ func (s *LogicalVolumeService) GetVolume(ctx context.Context, volumeID string) (
...
+		Cap:      10 * time.Second, // wait up to 10 seconds per retry, 1, 2, 4, 8, 10 seconds.
+		Steps:    10,               // run at most 10 times
+	}, func(ctx context.Context) (done bool, err error) {
 		var newLV topolvmv1.LogicalVolume
-		err := s.getter.Get(ctx, client.ObjectKey{Name: name}, &newLV)
-		if err != nil {
+		if err := s.getter.Get(ctx, client.ObjectKey{Name: name}, &newLV); err != nil {
 			logger.Error(err, "failed to get LogicalVolume", "name", name)
-			return "", err
+			return true, err
```
toshipp, line 396:
```suggestion
			return false, err
```

internal/driver/internal/k8s/logicalvolume_service.go
```diff
@@ -352,64 +354,60 @@ func (s *LogicalVolumeService) GetVolume(ctx context.Context, volumeID string) (
...
+			return true, nil
 		}
 		if newLV.Status.Code != codes.OK {
 			err := s.writer.Delete(ctx, &newLV)
 			if err != nil {
 				// log this error but do not return this error, because newLV.Status.Message is more important
 				logger.Error(err, "failed to delete LogicalVolume")
 			}
-			return "", status.Error(newLV.Status.Code, newLV.Status.Message)
+			return true, status.Error(newLV.Status.Code, newLV.Status.Message)
```
toshipp, line 409:
```suggestion
			return false, status.Error(newLV.Status.Code, newLV.Status.Message)
```

internal/driver/internal/k8s/logicalvolume_service.go
```diff
@@ -352,64 +354,60 @@ func (s *LogicalVolumeService) GetVolume(ctx context.Context, volumeID string) (
...
 		}
-	}
+		return false, ErrMissingStatusVolumeID
```
toshipp, line 411: We are waiting for Status.VolumeID is set, don't return an error.

toshipp (changes requested):

internal/driver/internal/k8s/logicalvolume_service.go
jakobmoellerdev: Just to make sure, previously this did not result in a for loop iteration because the return condition triggered. Do we want to switch to retries here as well?

internal/driver/internal/k8s/logicalvolume_service.go
```diff
@@ -313,36 +312,39 @@ func (s *LogicalVolumeService) ExpandVolume(ctx context.Context, volumeID string
...
-	// wait until topolvm-node expands the target volume
-	for {
-		logger.Info("waiting for update of 'status.currentSize'", "name", lv.Name)
-		select {
-		case <-ctx.Done():
-			return ctx.Err()
-		case <-time.After(1 * time.Second):
-		}
-
+	return wait.ExponentialBackoffWithContext(ctx, wait.Backoff{
```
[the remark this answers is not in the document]
jakobmoellerdev: I agree actually. I dont think the k8s libraries have a way to specify "no step limit though" so Im just gonna run with MAX_INT. Alternatively we could set this to an arbitrarily high number so that we avoid leaks in case the context is never closed (highly unlikely but possible)

internal/driver/internal/k8s/logicalvolume_service.go
jakobmoellerdev: Also the bool variable indicates done, so if we set this to false, then it would keep retrying (different from today)

internal/driver/internal/k8s/logicalvolume_service.go
jakobmoellerdev: See my message above, is retrying this instead of failing fast desired?

internal/driver/internal/k8s/logicalvolume_service.go
jakobmoellerdev: Youre absolutely right, I was assuming the bool would be the deterministic flag for retry, but if err is set it also aborts. need to adjust the PR.

internal/driver/internal/k8s/logicalvolume_service.go
jakobmoellerdev: See my message above, is retrying this instead of failing fast desired?

internal/driver/internal/k8s/logicalvolume_service.go
jakobmoellerdev: See my message above, is retrying this instead of failing fast desired?

internal/driver/internal/k8s/logicalvolume_service.go
jakobmoellerdev: See my message above, is retrying this instead of failing fast desired?

internal/driver/internal/k8s/logicalvolume_service.go
jakobmoellerdev: See my message above, is retrying this instead of failing fast desired?

internal/driver/internal/k8s/logicalvolume_service.go
```diff
@@ -238,23 +241,23 @@ func (s *LogicalVolumeService) DeleteVolume(ctx context.Context, volumeID string
...
-		if err != nil {
-			if apierrors.IsNotFound(err) {
-				return nil
+	return wait.ExponentialBackoffWithContext(ctx,
+		wait.Backoff{
+			Duration: 100 * time.Millisecond, // initial backoff
+			Factor:   2,                      // factor for duration increase
+			Jitter:   0.1,
+			Steps:    math.MaxInt, // run for infinity; we assume context gets canceled
+		}, func(ctx context.Context) (done bool, err error) {
```
daichimukai, line 250:
```suggestion
		}, func(ctx context.Context) (bool, error) {
```

Unused variables.

internal/driver/internal/k8s/logicalvolume_service.go
```diff
@@ -352,64 +358,59 @@ func (s *LogicalVolumeService) GetVolume(ctx context.Context, volumeID string) (
...
-		case <-time.After(100 * time.Millisecond):
-		}
-
+	var volumeID string
+	return volumeID, wait.ExponentialBackoffWithContext(ctx, wait.Backoff{
+		Duration: 1 * time.Second, // initial backoff
+		Factor:   2,               // factor for duration increase
+		Jitter:   0.1,
+		Steps:    math.MaxInt, // run for infinity; we assume context gets canceled
+	}, func(ctx context.Context) (done bool, err error) {
```
daichimukai, line 395:
```suggestion
	}, func(ctx context.Context) (bool, error) {
```

internal/driver/internal/k8s/logicalvolume_service.go
```diff
@@ -313,36 +316,39 @@ func (s *LogicalVolumeService) ExpandVolume(ctx context.Context, volumeID string
...
-			return ctx.Err()
-		case <-time.After(1 * time.Second):
-		}
-
+	return wait.ExponentialBackoffWithContext(ctx, wait.Backoff{
+		Duration: 1 * time.Second, // initial backoff
+		Factor:   2,               // factor for duration increase
+		Jitter:   0.1,
+		Steps:    math.MaxInt, // run for infinity; we assume context gets canceled
+	}, func(ctx context.Context) (done bool, err error) {
```
daichimukai, line 324:
```suggestion
	}, func(ctx context.Context) (bool, error) {
```

jakobmoellerdev:
I believe I spotted a side effect of this change which is kind of strange. There is currently a [failure in the resize test while checking for failure events](https://github.com/topolvm/topolvm/blob/d9cd7547b85557a3c2104d9c48dd7d5d11a1f8b7/test/e2e/e2e_test.go#L849). However, when isolating this test to run alone the failure doesnt occur, making me believe this is actually a regression from a different test.

jakobmoellerdev:
I think something is completely off here. The tests flake hard and im trying to understand why. This looks like a race condition somewhere, still trying to identify whats happening

toshipp:
I'll check it next week, but responses may be delayed due to the holiday period.

jakobmoellerdev:
I have noticed an interesting error message that is concerning to me as it leads me to believe that there might be an issue on the kubelet patch:

```
│ Name:             topo-pvc.17c9d3a398b23494                                                                                                                                                                                                                                   │
│ Namespace:        e2etest-xgwhounmef                                                                                                                                                                                                                                          │
│ Labels:           <none>                                                                                                                                                                                                                                                      │
│ Annotations:      <none>                                                                                                                                                                                                                                                      │
│ API Version:      v1                                                                                                                                                                                                                                                          │
│ Count:            1                                                                                                                                                                                                                                                           │
│ Event Time:       <nil>                                                                                                                                                                                                                                                       │
│ First Timestamp:  2024-04-26T12:17:12Z                                                                                                                                                                                                                                        │
│ Involved Object:                                                                                                                                                                                                                                                              │
│   API Version:       v1                                                                                                                                                                                                                                                       │
│   Kind:              PersistentVolumeClaim                                                                                                                                                                                                                                    │
│   Name:              topo-pvc                                                                                                                                                                                                                                                 │
│   Namespace:         e2etest-xgwhounmef                                                                                                                                                                                                                                       │
│   Resource Version:  1509                                                                                                                                                                                                                                                     │
│   UID:               d447f77e-fcb4-49f7-acea-73d26b524755                                                                                                                                                                                                                     │
│ Kind:                Event                                                                                                                                                                                                                                                    │
│ Last Timestamp:      2024-04-26T12:17:12Z                                                                                                                                                                                                                                     │
│ Message:             Mark PVC "e2etest-xgwhounmef/topo-pvc" as file system resize required failed: can't patch status of  PVC e2etest-xgwhounmef/topo-pvc with Operation cannot be fulfilled on persistentvolumeclaims "topo-pvc": the object has been modified; please apply │
│  your changes to the latest version and try again                                                                                                                                                                                                                             │
│ Metadata:                                                                                                                                                                                                                                                                     │
│   Creation Timestamp:  2024-04-26T12:17:12Z                                                                                                                                                                                                                                   │
│   Resource Version:    1524                                                                                                                                                                                                                                                   │
│   UID:                 0199e840-7f9a-422e-b44e-c3014f8f5c0e                                                                                                                                                                                                                   │
│ Reason:                VolumeResizeFailed 
```
This comes from the kubelets mismatch of desired state of world and actual state of world and is an issue in our tests because we test for VolumeExpansionFailed events. In this case the PVC is updated quite fast so the Event is triggered but we have an immediate success after.

test/e2e/e2e_test.go
```diff
@@ -851,7 +863,17 @@ func testE2E() {
 			"reason=VolumeResizeFailed"
 		var events corev1.EventList
 		err = getObjects(&events, "events", "-n", ns, "--field-selector="+fieldSelector)
-		Expect(err).To(BeEquivalentTo(ErrObjectNotFound))
+		wasModified := "the object has been modified; please apply your changes to the latest version and try again"
```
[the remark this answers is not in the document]
jakobmoellerdev: Even though this test is problematic right now I believe it is useful. That is because every failed run of the csi-resizer will be reported in such an event. Especially this is useful because retries are spotted here that should not happen in the first place. E.g. here, no one would expect any errors because there is nothing that should trigger a retry.

If we change our stance to say all failures are okay, then thats fine for me.

charts/topolvm/templates/controller/deployment.yaml
```diff
@@ -194,6 +194,7 @@ spec:
             - --leader-election-namespace={{ .Release.Namespace }}
             {{ end }}
             - --http-endpoint=:9810
+            - --timeout=300s
```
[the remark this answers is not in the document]
jakobmoellerdev: This is so that ControllerExpandVolume gets a 5min timeout. I initially did this for debugging and will remove it again, however I suspect this will lead to retries on the ControllerExpandVolume call. When reading through the code again, it becomes apparent that the first time we enter ControllerExpandVolume, we wait for LogicalVolume to be resized, after this we do not.

internal/driver/internal/k8s/logicalvolume_service.go
```diff
@@ -238,23 +241,23 @@ func (s *LogicalVolumeService) DeleteVolume(ctx context.Context, volumeID string
...
-		if err != nil {
-			if apierrors.IsNotFound(err) {
-				return nil
+	return wait.ExponentialBackoffWithContext(ctx,
+		wait.Backoff{
+			Duration: 100 * time.Millisecond, // initial backoff
+			Factor:   2,                      // factor for duration increase
+			Jitter:   0.1,
+			Steps:    math.MaxInt, // run for infinity; we assume context gets canceled
+		}, func(ctx context.Context) (bool, error) {
```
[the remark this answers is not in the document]
jakobmoellerdev: I removed caps and maximum backoff before abort after consideration with @toshipp. IMO too long backoff time cannot be avoided with wait.Backoff because if we reach the Cap it will get cancelled. I can only write our own backoff logic for this case or we accept a maximum timeout.

internal/driver/internal/k8s/logicalvolume_service.go
toshipp: not desired, but I assume you know, the value is not meaningless if an error returns, so I prefer the zero value.

internal/driver/internal/k8s/logicalvolume_service.go
toshipp: same as [this](https://github.com/topolvm/topolvm/pull/903#discussion_r1584516775)

internal/driver/internal/k8s/logicalvolume_service.go
```diff
@@ -352,64 +367,58 @@ func (s *LogicalVolumeService) GetVolume(ctx context.Context, volumeID string) (
...
+			}
+			lv.Annotations[topolvm.GetResizeRequestedAtKey()] = time.Now().UTC().String()
+
+			if err := s.writer.Update(ctx, lv); err != nil {
+				if apierrors.IsConflict(err) {
+					logger.Info("detected conflict when trying to update LogicalVolume spec", "name", lv.Name)
+				} else {
+					logger.Error(err, "failed to update LogicalVolume spec", "name", lv.Name)
+				}
+				return false, nil
```
toshipp, line 389: This breaks the previous behavior when the error is not Conflict. Would you do this on another PR if we change this?

internal/driver/internal/k8s/logicalvolume_service.go
toshipp: same as https://github.com/topolvm/topolvm/pull/903#discussion_r1584516775

internal/driver/internal/k8s/logicalvolume_service.go
jakobmoellerdev: Changed back to using 2 different returns (conflict returns false, nil and retries and other errors exit the loop early)

internal/driver/node.go
```diff
@@ -528,17 +525,15 @@ func (s *nodeServerNoLocked) NodeExpandVolume(ctx context.Context, req *csi.Node
...
 	if _, err := r.Resize(device, volumePath); err != nil {
 		return nil, status.Errorf(codes.Internal, "failed to resize filesystem %s (mounted at: %s): %v", volumeID, volumePath, err)
 	}
 
-	nodeLogger.Info("NodeExpandVolume(fs) is succeeded",
-		"volume_id", volumeID,
-		"target_path", volumePath,
-	)
-
+	logger.Info("NodeExpandVolume(fs) is succeeded")
```
toshipp, line 536: add newline.

internal/driver/internal/k8s/logicalvolume_service.go
```diff
@@ -238,23 +241,23 @@ func (s *LogicalVolumeService) DeleteVolume(ctx context.Context, volumeID string
...
+			Factor:   2,                      // factor for duration increase
+			Jitter:   0.1,
+			Steps:    math.MaxInt, // run for infinity; we assume context gets canceled
+		}, func(ctx context.Context) (bool, error) {
+			if err := s.getter.Get(ctx, client.ObjectKey{Name: lv.Name}, new(topolvmv1.LogicalVolume)); err != nil {
+				if apierrors.IsNotFound(err) {
+					return true, nil
+				}
+				logger.Error(err, "failed to get LogicalVolume", "name", lv.Name)
+				return false, nil
```
toshipp, line 256: Should we return err to keep the previous behavior?

internal/driver/internal/k8s/logicalvolume_service.go
```diff
@@ -301,48 +304,60 @@ func (s *LogicalVolumeService) CreateSnapshot(ctx context.Context, node, dc, sou
...
 			logger.Error(err, "failed to get LogicalVolume", "name", lv.Name)
-			return err
+			return false, err
 		}
+
+		if changedLV.Spec.Size.Cmp(*request) != 0 {
+			logger.Info("waiting for update of 'spec.size' to propagate "+
+				"to signal requested expansion", "name", lv.Name)
+			return false, nil
+		}
```
toshipp, line 337: To be clear, why is this needed? We updated the spec [here](https://github.com/topolvm/topolvm/pull/903/files#diff-f3c95da9af2be35fd4a88902d32407e715af7b28b89c590c97f9d3e5f8d96453R316), I assume this is redundant.

internal/driver/internal/k8s/logicalvolume_service.go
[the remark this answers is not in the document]
toshipp: Oh, I didn't know the behavior. I also think no-cap is not good, but I couldn't find other useful functions.

test/e2e/e2e_test.go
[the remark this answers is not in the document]
toshipp: This test ensures no event occurs before the test expects events.
https://github.com/topolvm/topolvm/pull/903/files#diff-d5d12b6ea07b0a78c9ad1987e1486c565ecf16839f77a93e91b09475cf6c0e99L856-R895

BTW, can we check if the resize fails by checking the PVC status?
https://github.com/topolvm/topolvm/pull/903/files#diff-d5d12b6ea07b0a78c9ad1987e1486c565ecf16839f77a93e91b09475cf6c0e99R878-R881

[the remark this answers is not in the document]
jakobmoellerdev: Sure it would be possible to check for PersistentVolumeClaimControllerResizeFailed or PersistentVolumeClaimNodeResizeFailed conditions to indicate PVC resize failure

internal/driver/internal/k8s/logicalvolume_service.go
jakobmoellerdev: Usually this is avoided in controllers because it is recommended to just return and wait for the watcher to update after updating a resource in the reconcile loop, but we dont have that.

toshipp: Get it. BTW, we already know the request size, so why don't you use the `request` variable directly [here](https://github.com/topolvm/topolvm/pull/903/files/51d0ca59d9a2fcf59eaefbb63b6882181047e839#diff-f3c95da9af2be35fd4a88902d32407e715af7b28b89c590c97f9d3e5f8d96453L339)?

test/e2e/e2e_test.go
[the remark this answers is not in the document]
toshipp: I'd like to write it the way you suggested. What do you think? @daichimukai

internal/driver/internal/k8s/logicalvolume_service.go
jakobmoellerdev: I think ultimately it does not matter if I use request because above I checked that spec.Size == request. If its important to you I can adjust it.

test/e2e/e2e_test.go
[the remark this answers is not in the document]
jakobmoellerdev: added the test case as suggested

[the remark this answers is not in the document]
toshipp: Thanks to the new checking code, we can now remove the event check codes, which are redundant and difficult to understand the intention to check resize failure. Could you remove them?

internal/driver/internal/k8s/logicalvolume_service.go
toshipp: I mean lines from 334 to 337 and lines from 351 to 356 can be merged. We know `changedLV.Status.CurrentSize` will eventually equal `request`, could we write the code as such?

test/e2e/e2e_test.go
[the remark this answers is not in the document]
jakobmoellerdev: The event check is meant to check for explicit failures after going over capacity. The event checking code is tested before that, when a succesful resize is expected. We do not expect any failure events during this, which is why we have this test. I suggest we keep the test function in for that reason.

internal/driver/internal/k8s/logicalvolume_service.go
jakobmoellerdev: Got what you mean now, changed accordingly, youre completely right.

internal/driver/internal/k8s/logicalvolume_service.go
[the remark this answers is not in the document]
daichimukai: How about this code? I think this effectively do an exponential backoff with the limited backoff time and unlimited retries.

```go
wait.Backoff{
	Duration: 100 * time.Millisecond,
	Factor:   2,
	Jitter:   0.1,
	Steps:    math.MaxInt,
	Cap:      5 * time.Minute,
}.DelayFunc().Until(ctx, false, false, func(ctx context.Context) (bool, error) {
	// check conditions
})
```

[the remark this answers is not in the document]
jakobmoellerdev: This will stop retrying once 5 minutes are hit as duration. See https://github.com/kubernetes/apimachinery/blob/c9c3e94f52f06444120a09bfbdb11bec38a5c8e1/pkg/util/wait/backoff.go#L111 for the concrete logic that sets steps to 0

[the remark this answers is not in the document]
daichimukai: Seriously? If I understand correctly, it is just a logic of ExponentialBackoffWithContext to stop retrying as Backoff.Steps becomes zero. Generally it's not the case.

Test Code:

```go
func TestExponentialBackoff(t *testing.T) {
	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
	defer cancel()

	var prev time.Time
	f := func(ctx context.Context) (bool, error) {
		if !prev.IsZero() {
			fmt.Printf("%s\n", time.Since(prev))
		}
		prev = time.Now()
		return false, nil
	}

	wait.Backoff{
		Duration: 100 * time.Millisecond,
		Factor:   2,
		Jitter:   0.1,
		Steps:    math.MaxInt,
		Cap:      1 * time.Second,
	}.DelayFunc().Until(ctx, true, false, f)
}
```

Output:

```
=== RUN   TestExponentialBackoff
400ns
110.298409ms
201.412217ms
424.604589ms
872.976143ms
1.019066153s
1.014166001s
1.077151559s
1.074724855s
1.039145566s
1.038260597s
1.038210446s
1.033878205s
--- PASS: TestExponentialBackoff (10.00s)
```

NOTE: The first retry after 400ns seems to be a bug that is already fixed in the master branch of k8s ([commit](https://github.com/kubernetes/kubernetes/commit/79f9bfdc8e5f5805c98741fd10e57db95c3c604e)).

test/e2e/e2e_test.go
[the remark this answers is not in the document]
toshipp: I think what we want to check is that the PV resizing succeeds. This can be achieved by checking no events, but it is an implicit condition, IMO. Therefore, I prefer to use explicit checking and delete implicit one.

This is a side note, we ignore events showing retry on conflicts [here](https://github.com/topolvm/topolvm/pull/903/files#diff-d5d12b6ea07b0a78c9ad1987e1486c565ecf16839f77a93e91b09475cf6c0e99R877). If there are such events, [this condition](https://github.com/topolvm/topolvm/pull/903/files#diff-d5d12b6ea07b0a78c9ad1987e1486c565ecf16839f77a93e91b09475cf6c0e99R896-R907) is accidentally met.

internal/driver/internal/k8s/logicalvolume_service.go
```diff
@@ -301,48 +304,58 @@ func (s *LogicalVolumeService) CreateSnapshot(ctx context.Context, node, dc, sou
...
-		if changedLV.Status.CurrentSize.Value() != changedLV.Spec.Size.Value() {
-			logger.Info("failed to match current size and requested size", "current", changedLV.Status.CurrentSize.Value(), "requested", changedLV.Spec.Size.Value())
-			continue
+
+		if changedLV.Status.CurrentSize.Cmp(*request) != 0 {
+			logger.Info("waiting for update of 'status.currentSize' to be updated to signal successful expansion",
+				"name", lv.Name,
+				"status.currentSize", changedLV.Status.CurrentSize,
+				"spec.size", changedLV.Spec.Size,
+				"status.currentSize", changedLV.Status.CurrentSize,
```
toshipp, line 350: It would be redundant with line 348.

internal/driver/internal/k8s/logicalvolume_service.go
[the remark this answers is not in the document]
jakobmoellerdev: You are absolutely correct, I completely missed that you switched from ExponentialBackoffWithContext to DelayFunc. I will switch accordingly. Thanks for the correction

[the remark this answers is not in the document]
jakobmoellerdev: Except for the spec update i now switched everything to the capped delay with a maximum of 10s. The spec update with conflict checking I left with the default backoff because we do not expect long wait times on this and should error if we have update contention.

test/e2e/e2e_test.go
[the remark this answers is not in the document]
jakobmoellerdev: I think we have 2 scenarios: successful resize, and failinig resize due to lack of capacity.

For successful resize, having no failure events might be seen as critical or not. Of course in an eventually consistent system we are fine with failures every now and then but if we have a system in which we can ensure optimal conditions, then testing that no intermediate failure event occurs makes sense imo.

For failing resize, similarly it makes sense to verify that we have events checked that report this failure on top of the failing PVC status.

That being said, Im also fine with removing after all if you think thats not enough of a reason, not married to this idea.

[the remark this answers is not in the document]
toshipp: I found the checks were introduced in [this commit](https://github.com/topolvm/topolvm/pull/101/commits/9f84e16f214d94db0f67fae58f098a74bff7e347) and asked the authors about the intention. The answer is that the first check is a precondition of the failing resize to ensure no events occur and the last is a postcondition to confirm events happen. They also told me that checking events is just a tool to check the failing resize, alternatives could be used.

I feel checking the status is more readable, so I'd like to use it. However, I found this method can not be used because  external-resizer sets the status only when the RecoverVolumeExpansionFailure feature gate is enabled (default false now).
https://github.com/kubernetes-csi/external-resizer/blob/master/pkg/controller/controller.go#L336
You can see [expandAndRecover](https://github.com/kubernetes-csi/external-resizer/blob/master/pkg/controller/expand_and_recover.go#L28) is called, and it calls [callResizeOnPlugin](https://github.com/kubernetes-csi/external-resizer/blob/master/pkg/controller/expand_and_recover.go#L156), and then it calls [markControllerExpansionFailed](https://github.com/kubernetes-csi/external-resizer/blob/master/pkg/controller/resize_status.go#L90) which updates the status.

But, why did our test succeed? It is because we missed `.Should(Succeed())` [here](https://github.com/topolvm/topolvm/pull/903/files#diff-d5d12b6ea07b0a78c9ad1987e1486c565ecf16839f77a93e91b09475cf6c0e99R928).

So, we should retain the event check code and remove the status check.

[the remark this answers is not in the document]
jakobmoellerdev: Thanks for going in detail on this and explaining that for me. Switched back to events for now with the exception for the Collision to avoid flakes.

internal/driver/internal/k8s/logicalvolume_service.go
```diff
@@ -352,64 +365,60 @@ func (s *LogicalVolumeService) GetVolume(ctx context.Context, volumeID string) (
...
-		logger.Info("waiting for setting 'status.volumeID'", "name", name)
-		select {
-		case <-ctx.Done():
-			return "", ctx.Err()
-		case <-time.After(100 * time.Millisecond):
-		}
-
+	var volumeID string
+	return volumeID, wait.Backoff{
+		Duration: 1 * time.Second, // initial backoff
```
toshipp, line 398: The previous code waits 100ms, do you have any reason to change it?

internal/driver/internal/k8s/logicalvolume_service.go
```diff
@@ -352,64 +365,60 @@ func (s *LogicalVolumeService) GetVolume(ctx context.Context, volumeID string) (
...
-		}
-		lv.Annotations[topolvm.GetResizeRequestedAtKey()] = time.Now().UTC().String()
-
-		if err := s.writer.Update(ctx, lv); err != nil {
-			if apierrors.IsConflict(err) {
-				logger.Info("detect conflict when LogicalVolume spec update", "name", lv.Name)
-				continue
+	return wait.ExponentialBackoffWithContext(ctx,
+		retry.DefaultBackoff,
+		func(ctx context.Context) (bool, error) {
```
toshipp, line 370: Also, do we need to write it as `wait.Backoff{...}.DelayFunc().Until`?

internal/driver/internal/k8s/logicalvolume_service.go
jakobmoellerdev: The previous code executes its first status check after 100ms, not immediately. Here the first check is run immediately. I think waiting 100ms for a status update here is overly expensive because LogicalVolumes will likely not be updated that fast (I think 1-2 seconds is most realistic). Thats why I changed it. Feel free to let me know if I should adjust back.

internal/driver/internal/k8s/logicalvolume_service.go
jakobmoellerdev: For the Spec size we only retry for conflicts so we should adopt the K8s default timings for dealing with conflicts or abort if we cannot resolve in time as this would mean that another controller is fighting with topolvm over updates. In this case waiting until context cancelation is not wanted I believe.

internal/driver/internal/k8s/logicalvolume_service.go
toshipp: Got it. I don't think we have hard requirements to this value, it's OK.

toshipp (approved):
LGTM!

toshipp:
@jakobmoellerdev 
If you wish to squash some commits, please tell us.

jakobmoellerdev:
I will squash

jakobmoellerdev:
I have the fix for external resizer not sending events in merged now, so once we update to the 1.30 release of resizer, the event test will no longer have to contain the special check.

toshipp:
I'm glad to hear that. I assume [this](https://github.com/kubernetes-csi/external-resizer/pull/392) is yours. Great work! Thank you @jakobmoellerdev

toshipp (approved):

daichimukai (approved):

PR #910, opened 2024-05-10
Support k8s 1.29

toshipp:
Fix: #793

PR #917, opened 2024-05-16
fix: reallow building for s390x

jakobmoellerdev:
Reapplies the cast from 70b37b16075bfd5af3c4cf9d4373c3bd1a19138f that was removed by 5e84659232096f23e73b18c7753817b4a09b3f0a by accident

In downstream we use this to build TopoLVM for s390x and it should not interfere with other architectures


llamerada-jp (approved):

toshipp (commented):
@jakobmoellerdev 
The PR itself looks good to me.

I want to avoid regressions like this, it would be better to add a compile test to CI.
How do you think about adding `make build-topolvm GOARCH=s390x` [here](https://github.com/topolvm/topolvm/blob/main/.github/workflows/main.yaml#L20)?

toshipp:
Another thought, we will be able to build s390x on the multi-platform docker image, but I think it is not a good idea because the build time is already too long, and I assume users who want to use s390x tend to use vendor build binaries.

jakobmoellerdev:
I support that idea, s390x should not be the default compile target as its user base is extremely limited, even for us, but if someone uses it, its fundamental in their architecture.

Regarding where to add it, im open to add it to the tests, but that will increase the test. In general, on the image build time, I think it has been mentioned before but TopoLVM disables the docker build cache and could be optimized a bit, maybe that is worth targeting?

toshipp:
@jakobmoellerdev 
Do you mean you want to add s390x support to our multi-platform build target?
https://github.com/topolvm/topolvm/blob/main/Makefile#L207
I thought that simply adding compilation tests to the main workflow would be sufficient.

We use apt in Dockerfile, the upstream repository may be updated even if we change nothing, so we used `--no-cache`. But it is ok to add an option to enable caching, IMO.

toshipp:
Sorry, I misunderstood what you said.
I think we should ensure to build the binary but do not have to provide official images for the architectures because they don't use our binaries (I assume they use enterprise-supported binaries).

jakobmoellerdev:
Check uncommited is showing a failure in lvmd.pb.go but im 99% certain that this is a cache issue since the last k8s version bump PR. @toshipp do you mind taking a look and telling me what to do with that?

toshipp:
I updated protoc version on https://github.com/topolvm/topolvm/pull/910, but I missed updating `pkg/lvmd/proto/lvmd.pb.go` and `pkg/lvmd/proto/lvmd_grpc.pb.go`
Could you generate them as follows?

```
rm pkg/lvmd/proto/lvmd.pb.go
rm pkg/lvmd/proto/lvmd_grpc.pb.go
make generate
```

toshipp (approved):

llamerada-jp (approved):

Issue #907, opened 2024-05-03
Docs: Please elaborate `topolvm.io/webhook=ignore`


guettli:
**What should the feature do:**

In the future, the reader of [getting started docs](https://github.com/topolvm/topolvm/blob/main/docs/getting-started.md#install-instructions) should understand this:

> TopoLVM uses webhooks. To work webhooks properly, add a label to the target namespace. We also recommend to use a dedicated namespace.

```
kubectl label namespace topolvm-system topolvm.io/webhook=ignore
kubectl label namespace kube-system topolvm.io/webhook=ignore
```
---

Please elaborate the markdown-file and explain why this is needed.

`webhook=ignore` looks like disabling webhooks.

---
Thank you very much for topolvm!

cupnes:
@guettli 
If the namespace is not labeled `topolvm.io/webhook=ignore`, mutating webhooks of `topolvm-controller` are executed for pods and PVCs. These webhooks are for resources managed by TopoLVM. Therefore, `topolvm-system` and `kube-system` are labeled with `topolvm.io/webhook=ignore` to prevent those webhooks from being executed.
I will add the above explanation to [getting started docs](https://github.com/topolvm/topolvm/blob/main/docs/getting-started.md#install-instructions).

guettli:
What would happen, if you don't set the label on the namespace?

Why do the mutating webhooks of topolvm-controller not handle that? Then it would be a bit easier to configure topolvm.

cupnes:
@guettli 
If `topolvm.io/webhook=ignore` isn't set for the namespace, Pods and PVCs in that namespace are handled by mutating webhooks. In the case of Pods, this means [adding annotations about capacity, etc](https://github.com/topolvm/topolvm/blob/d7d81c8ee62d2532da801bce9f5e5fc509c76a18/internal/hook/mutate_pod.go#L44-L101). In the case of PVCs, this means [adding finalizers, etc](https://github.com/topolvm/topolvm/blob/d7d81c8ee62d2532da801bce9f5e5fc509c76a18/internal/hook/mutate_persistentvolumeclaim.go#L38-L83).

Whether mutating webhooks in topolvm-controller works or not should be configured by the user. At first glance, a possible implementation is that the webhooks handler should do nothing for the `topolvm-system` or `kube-system` namespace, but in some environments, the Kubernetes system may not be a `kube-system` namespace.

guettli:
[...]

PR #920, opened 2024-05-20
docs: add a note explaining the need for `topolvm.io/webhook=ignore`

cupnes:
In getting started, it is not clear why `topolvm.io/webhook=ignore` is needed. Add a note explaining it.

Closes: https://github.com/topolvm/topolvm/issues/907

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -24,6 +24,9 @@ kubectl label namespace topolvm-system topolvm.io/webhook=ignore
 kubectl label namespace kube-system topolvm.io/webhook=ignore
 ```
 
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


ushitora-anqou (changes requested):
I feel that this description is a bit obscure. How about something like this?


> The label `topolvm.io/webhook=ignore` is set here to prevent the triggering of TopoLVM's mutating webhooks when pods and pvcs are created in the topolvm-system and kube-system namespaces. These webhooks depend on the topolvm-controller. Therefore, the webhooks should not be activated during TopoLVM's startup process to avoid causing it to become stuck.

cupnes:
@ushitora-anqou 
Thank you. It looks good, I will use it.

ushitora-anqou (approved):

pluser (approved):

PR #924, opened 2024-06-07
Remove --container option

toshipp:
nsenter is used to call lvmd command on the container environment because lvm command should be called on the root namespace. To enable nsenter, we introduced --container option to lvmd. On the host, nsenter is no-op, therefore, we can omit the option.


jakobmoellerdev (commented):
I did verify that for root this is a noop, however for non root this results in elevation to root. Im assuming the expectation is that lvmd always runs privileged? If so this PR is okay for me

peng225 (approved):

toshipp:
A process that runs lvm needs to have CAP_SYS_ADMIN. We can assume that.
https://github.com/torvalds/linux/blob/83a7eefedc9b56fe7bfeff13b6c7356688ffa670/drivers/md/dm-ioctl.c#L2040

jakobmoellerdev (approved):

llamerada-jp (approved):

PR #931, opened 2024-07-03
Add pprof profiling support to all binaries.

naemono:
## What is this change?

This adds the ability to expose the [pprof](https://pkg.go.dev/net/http/pprof) profiler for all of the topolvm binaries.

### Background

We are seeing a bit of memory growth over time with lvmd, and wanted to run some profiling to see if there's some sort of memory leak, but don't have that ability. This would add that functionality.

### DIFF
--- a/charts/topolvm/templates/controller/deployment.yaml
+++ b/charts/topolvm/templates/controller/deployment.yaml
@@ -64,6 +64,9 @@ spec:
             {{- if .Values.controller.nodeFinalize.skipped }}
             - --skip-node-finalize
             {{- end }}
+            {{- if .Values.controller.profiling.bindAddress }}
+            - --profiling-bind-address={{ .Values.controller.profiling.bindAddress }}
+            {{- end }}
           {{- if or .Values.useLegacy .Values.env.topolvm_controller }}
           env:
             {{- if .Values.useLegacy }}
--- a/charts/topolvm/values.yaml
+++ b/charts/topolvm/values.yaml
@@ -110,6 +110,10 @@ scheduler:
   #    ssd: 1
   #    hdd: 10
 
+  profiling:
+    # scheduler.profiling.bindAddress -- Enables pprof profiling server. If empty, profiling is disabled.
+    bindAddress: ""
+
   options:
     listen:
       # scheduler.options.listen.host -- Host used by Probe.
@@ -212,6 +216,10 @@ lvmd:
 
   # lvmd.initContainers -- Additional initContainers for the lvmd service.
   initContainers: []
+
+  profiling:
+    # lvmd.profiling.bindAddress -- Enables pprof profiling server. If empty, profiling is disabled.
+    bindAddress: ""
 
 # CSI node service
 node:
@@ -345,6 +353,10 @@ node:
   # node.labels -- Additional labels to be added to the Daemonset.
   labels: {}
 
+  profiling:
+    # node.profiling.bindAddress -- Enables pprof profiling server. if empty profiling is disabled.
+    bindAddress: ""
+
   # node.initContainers -- Additional initContainers for the node service.
   initContainers: []
 
@@ -463,6 +475,10 @@ controller:
   podLabels: {}
   # controller.labels -- Additional labels to be added to the Deployment.
   labels: {}
+
+  profiling:
+    # controller.profiling.bindAddress -- Enables pprof profiling server. If empty, profiling is disabled.
+    bindAddress: ""
 
   # controller.initContainers -- Additional initContainers for the controller service.
   initContainers: []
--- a/cmd/lvmd/app/root.go
+++ b/cmd/lvmd/app/root.go
@@ -2,9 +2,11 @@ package app
 
 import (
 	"context"
+	"errors"
 	"flag"
 	"fmt"
 	"net"
+	"net/http"
 	"os"
 	"os/signal"
 	"path/filepath"
@@ -15,6 +17,7 @@ import (
 	"github.com/topolvm/topolvm"
 	"github.com/topolvm/topolvm/internal/lvmd"
 	"github.com/topolvm/topolvm/internal/lvmd/command"
+	"github.com/topolvm/topolvm/internal/profiling"
 	"github.com/topolvm/topolvm/pkg/lvmd/proto"
 	lvmdTypes "github.com/topolvm/topolvm/pkg/lvmd/types"
 	"google.golang.org/grpc"
@@ -26,9 +29,10 @@ import (
 )
 
 var (
-	cfgFilePath string
-	lvmPath     string
-	zapOpts     zap.Options
+	cfgFilePath          string
+	lvmPath              string
+	zapOpts              zap.Options
+	profilingBindAddress string
 )
 
 // rootCmd represents the base command when called without any subcommands
@@ -51,13 +55,13 @@ volume group.
 	},
 }
 
-func subMain(ctx context.Context) error {
+func subMain(parentCtx context.Context) error {
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&zapOpts)))
-	logger := log.FromContext(ctx)
+	logger := log.FromContext(parentCtx)
 
 	command.SetLVMPath(lvmPath)
 
-	if err := loadConfFile(ctx, cfgFilePath); err != nil {
+	if err := loadConfFile(parentCtx, cfgFilePath); err != nil {
 		return err
 	}
 
@@ -65,7 +69,7 @@ func subMain(ctx context.Context) error {
 		return err
 	}
 
-	vgs, err := command.ListVolumeGroups(ctx)
+	vgs, err := command.ListVolumeGroups(parentCtx)
 	if err != nil {
 		logger.Error(err, "error while retrieving volume groups")
 		return err
@@ -79,7 +83,7 @@ func subMain(ctx context.Context) error {
 		}
 
 		if dc.Type == lvmdTypes.TypeThin {
-			_, err = vg.FindPool(ctx, dc.ThinPoolConfig.Name)
+			_, err = vg.FindPool(parentCtx, dc.ThinPoolConfig.Name)
 			if err != nil {
 				logger.Error(err, "Thin pool not found:", "thinpool", dc.ThinPoolConfig.Name)
 				return err
@@ -105,8 +109,18 @@ func subMain(ctx context.Context) error {
 	proto.RegisterLVServiceServer(grpcServer, lvmd.NewLVService(dcm, ocm, notifier))
 	grpc_health_v1.RegisterHealthServer(grpcServer, lvmd.NewHealthService())
 
-	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
+	ctx, stop := signal.NotifyContext(parentCtx, os.Interrupt, syscall.SIGTERM)
 	defer stop()
+
+	var pprofServer *http.Server
+	if profilingBindAddress != "" {
+		pprofServer = profiling.NewProfilingServer(profilingBindAddress)
+		go func() {
+			if err := pprofServer.ListenAndServe(); !errors.Is(err, http.ErrServerClosed) {
+				logger.Error(err, "pprof server error")
+			}
+		}()
+	}
 
 	go func() {
 		ticker := time.NewTicker(10 * time.Minute)
@@ -114,6 +128,11 @@ func subMain(ctx context.Context) error {
 			select {
 			case <-ctx.Done():
 				ticker.Stop()
+				if pprofServer != nil {
+					if err := pprofServer.Shutdown(parentCtx); err != nil {
+						logger.Error(err, "failed to shutdown pprof server")
+					}
+				}
 				grpcServer.GracefulStop()
 				return
 			case <-ticker.C:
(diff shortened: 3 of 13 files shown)


naemono:
## Testing:

Tested locally  in minikube

```
❯ helm upgrade -i -n topolvm-system topolvm ../../charts/topolvm -f manifests/values/local.yaml
❯ kc get pod -n topolvm-system
NAME                                               READY   STATUS    RESTARTS      AGE
topolvm-cert-manager-657b7864b7-hhw2k              1/1     Running   0             9m13s
topolvm-cert-manager-cainjector-57fbb46b78-7k6nq   1/1     Running   0             9m13s
topolvm-cert-manager-webhook-85bff86bcc-zcj7w      1/1     Running   0             9m13s
topolvm-controller-664c66c87c-v46j4                5/5     Running   0             34s
topolvm-lvmd-0-t4qh8                               1/1     Running   0             33s
topolvm-node-58smh                                 3/3     Running   1 (31s ago)   33s

❯ diff manifests/values/base.yaml manifests/values/local.yaml 
2,4c2,4
<   repository: topolvm
<   tag: dev
<   pullPolicy: Never
---
>   repository: docker.io/personal/topolvm-with-sidecar
>   tag: devel
>   pullPolicy: Always
11a12,13
>   profiling:
>     port: 6060
46a49,50
>   profiling:
>     port: 6060
49a54,55
>   profiling:
>     port: 6060

```

internal/profiling/server.go
```diff
@@ -0,0 +1,11 @@
+package profiling
+
+import (
+	"fmt"
+	"net/http"
+	_ "net/http/pprof"
```
jakobmoellerdev, line 6: please use a concrete import here, not the side effects based one, as this will automatically start pprof on the default handler (https://go.dev/src/net/http/pprof/pprof.go look into the init function). if you want to do it manually:

https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/manager/internal.go#L312-L327 see here for an example

cmd/lvmd/app/root.go
```diff
@@ -108,6 +109,10 @@ func subMain(ctx context.Context) error {
 	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
 	defer stop()
 
+	if config.ProfilingPort > 0 {
+		go profiling.StartProfiling(config.ProfilingPort)
```
jakobmoellerdev, line 113: This goroutine leaks and is not controlled by anything. Its preferrable to add it via controller-runtimes manager.Add()

jakobmoellerdev (changes requested):
I think this PR in general is a great idea, thank you for your submission. However I have a few concerns:

1. Why do you not choose to use the existing pprof implementation at https://github.com/topolvm/topolvm/blob/ebfa7d02ad403a3a60fa9031a2d1796959e0dba8/cmd/topolvm-controller/app/run.go#L74-L89 via [PprofBindAddress](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.18.4/pkg/manager/manager.go#L237-L242)?
2. The goroutine you start uses a generic http handler and also is not tracked via the manager, making shutdown leak the goroutine

Last but not least, it would be great if you could share your memory observations over time. The leak findings are insanely useful and we will do our best to remediate

jakobmoellerdev:
Another thing of note. If you use the images from topolvm directly, please note that we strip debugging information: https://github.com/topolvm/topolvm/blob/2096240b0b18dcfb4e4656259c5619176a2b102e/Makefile#L163 

It would be cool if we could add pprof when debugging symbols are enabled, but Ill leave it to you if you want to implement that or not

cmd/lvmd/app/root.go
toshipp: Agreed.

naemono:
> 1. Why do you not choose to use the existing pprof implementation at

I wasn't aware of this implementation in `controller-runtime`. I will migrate this to use that instead. Thanks for the pointer.

cmd/lvmd/app/root.go
naemono: I'll get these non-controller-runtime implementations moved over to something that is able to be cancelled properly via a context/termination signal.. :+1:

naemono:
> Last but not least, it would be great if you could share your memory observations over time. The leak findings are insanely useful and we will do our best to remediate

I will share more information when I have the details. At this time I don't have a ton to go on, just observation of memory growth over time that seems specific to the lvmd process.

naemono:
Ok, the current changes use the `controller-runtime`'s pprof integration when available, otherwise it handled shutdown gracefully via `context` and doesn't leak goroutines. 

As for the note/question about stripping debugging information, were you suggesting detecting the debugging symbol's existence at runtime, and only serving the pprof endpoint when the debugging symbols exist? In my testing, even without the debugging symbols the pprof endpoint is rather useful. Let me know if you want some changes in this area. Thanks much for the quick review!

cmd/lvmd/app/config.go
```diff
@@ -18,10 +18,13 @@ type Config struct {
...
 	// LvcreateOptionClasses are classes that define options for the lvcreate command
 	LvcreateOptionClasses []*lvmdTypes.LvcreateOptionClass `json:"lvcreate-option-classes"`
+	// ProfilingPort is the port number to expose profiling. If 0, profiling is disabled.
+	ProfilingPort int `json:"profiling-port"`
 }
 
 var config = &Config{
-	SocketName: topolvm.DefaultLVMdSocket,
+	SocketName:    topolvm.DefaultLVMdSocket,
+	ProfilingPort: 0,
```
jakobmoellerdev, line 27: this is the zero value and you can just leave it out of the definition of the struct

cmd/lvmd/app/config.go
```diff
@@ -18,10 +18,13 @@ type Config struct {
 	DeviceClasses []*lvmdTypes.DeviceClass `json:"device-classes"`
 	// LvcreateOptionClasses are classes that define options for the lvcreate command
 	LvcreateOptionClasses []*lvmdTypes.LvcreateOptionClass `json:"lvcreate-option-classes"`
+	// ProfilingPort is the port number to expose profiling. If 0, profiling is disabled.
+	ProfilingPort int `json:"profiling-port"`
```
jakobmoellerdev, line 22: how about changing this to a bind address instead so that people can not only control the port but also the network interface?

internal/profiling/server.go
```diff
@@ -0,0 +1,24 @@
...
+	"fmt"
+	"net/http"
+	"net/http/pprof"
+)
+
+// NewProfilingServer creates a new HTTP server for profiling.
+func NewProfilingServer(port int) *http.Server {
+	mux := http.NewServeMux()
+	srv := http.Server{
+		Addr:    fmt.Sprintf(":%d", port),
```
jakobmoellerdev, line 13: if you forward this entire address as argument instead of the port it allows people to use a different (e.g. non exposed) network for debugging in production environments without security risk.

cmd/topolvm-node/app/run.go
```diff
@@ -72,9 +72,10 @@ func subMain(ctx context.Context) error {
 	}
 
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
-		Scheme:         scheme,
-		Metrics:        metricsServerOptions,
-		LeaderElection: false,
+		Scheme:           scheme,
+		Metrics:          metricsServerOptions,
+		LeaderElection:   false,
+		PprofBindAddress: fmt.Sprintf(":%d", config.profilingPort),
```
jakobmoellerdev, line 78: similarly here, just forward the entire pprof bind address as an arg :)

jakobmoellerdev (changes requested):
Mostly LGTM, just want to make sure we expose the entire bind address, not just the port. Thanks for taking care!

cmd/lvmd/app/root.go
```diff
@@ -108,13 +111,30 @@ func subMain(ctx context.Context) error {
 	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
 	defer stop()
 
+	var pprofServer *http.Server
+	if config.ProfilingPort > 0 {
+		go func() {
+			pprofServer = profiling.NewProfilingServer(config.ProfilingPort)
```
toshipp, line 117: The assignment should be out of the go statement because the shutdown code might access `pprofServer` before this.

cmd/topolvm-scheduler/app/root.go
```diff
@@ -99,13 +103,28 @@ func subMain(parentCtx context.Context) error {
 	ctx, stop := signal.NotifyContext(parentCtx, os.Interrupt, syscall.SIGTERM)
 	defer stop() // stop() should be called before wg.Wait() to stop the goroutine correctly.
 
+	var pprofServer *http.Server
+	if config.ProfilingPort > 0 {
+		go func() {
+			pprofServer = profiling.NewProfilingServer(config.ProfilingPort)
```
toshipp, line 109: Also move this.

toshipp (changes requested):

cmd/lvmd/app/config.go
naemono: All implementations have been moved to using a bind address instead.

cmd/lvmd/app/root.go
```diff
@@ -108,13 +111,30 @@ func subMain(ctx context.Context) error {
...
+
 	go func() {
 		ticker := time.NewTicker(10 * time.Minute)
 		for {
 			select {
 			case <-ctx.Done():
 				ticker.Stop()
 				grpcServer.GracefulStop()
+				if pprofServer != nil {
+					shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), 5*time.Second)
```
jakobmoellerdev, line 132: This is the wrong background context and also the shutdown should be bound to the base context

toshipp: Since the `ctx` has been closed, we can not use it, IMO.

charts/topolvm/templates/controller/deployment.yaml
```diff
@@ -64,6 +64,7 @@ spec:
             {{- if .Values.controller.nodeFinalize.skipped }}
             - --skip-node-finalize
             {{- end }}
+            - --profiling-port={{ .Values.controller.profiling.port | default 0 }}
```
toshipp, line 67: We should update this. Must check other places.

toshipp (changes requested):

cmd/lvmd/app/root.go
toshipp: Hmm, we already call `grpcServer.GracefulStop`, the `subMain` may exit before shutting down pprofServer. It would be better to care about this case or close the server immediately.

charts/topolvm/templates/controller/deployment.yaml
naemono: :man_facepalming: I didn't test this after the last update or I would have caught this. I'll get this fixed. Thanks.

naemono: Ok, these have all been updated, and tested with the helm chart in minikube.

cmd/lvmd/app/root.go
naemono: @jakobmoellerdev I believe I understand what you're asking for now, and have made the changes. Lmk, thanks.

jakobmoellerdev (approved):
LGTM, thanks so much for the submission again! :)

toshipp (changes requested):

charts/topolvm/values.yaml
```diff
@@ -110,6 +110,10 @@ scheduler:
   #    ssd: 1
   #    hdd: 10
 
+  profiling:
+    # scheduler.profiling.bindAddress -- Enables pprof profiling server. Empty, or '0' disables profiling.
```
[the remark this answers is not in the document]
naemono: You are correct, no longer true. removed.

cmd/lvmd/app/config.go
```diff
@@ -18,6 +18,8 @@ type Config struct {
 	DeviceClasses []*lvmdTypes.DeviceClass `json:"device-classes"`
 	// LvcreateOptionClasses are classes that define options for the lvcreate command
 	LvcreateOptionClasses []*lvmdTypes.LvcreateOptionClass `json:"lvcreate-option-classes"`
+	// ProfilingBindAddress is the bind address to expose pprof profiling. If empty, profiling is disabled.
+	ProfilingBindAddress string `json:"profiling-bind-address"`
```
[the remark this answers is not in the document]
naemono: I was unaware of this option. I have removed this, and added a conditional cli flag.

charts/topolvm/values.yaml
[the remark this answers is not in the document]
toshipp: Could you please update the comments in this file?

toshipp (changes requested):
mostly good. I'll approve this after you update values.yaml.
Also, these commits can be squashed, can you do that?

naemono:
> mostly good. I'll approve this after you update values.yaml. Also, these commits can be squashed, can you do that?

Values file updated, and I'm working to squash these commits. :+1:

jakobmoellerdev (dismissed):

naemono:
@jakobmoellerdev It appears the helm readme hadn't been updated with `helm-docs`. I've run this update and will check back the ci tasks.

jakobmoellerdev (approved):

toshipp (commented):
@naemono 
Please squash two commits.

> Add pprofl profiling support to all binaries.

I suspect `pprofl` is a typo of `pprof`. If true, please fix the commit message.

toshipp (approved):
LGTM. Thank you for your contribution!

llamerada-jp:
Hi @naemono ,

> We are seeing a bit of memory growth over time with lvmd, and wanted to run some profiling to see if there's some sort of memory leak, but don't have that ability. This would add that functionality.

FYI, our team also found lvmd memory growth. I don't know if this is the same issue as your report.

The issue we found was that as we repeatedly created and deleted PVs in TopoLVM, the metrics `container_memory_usage_bytes{namespace="topolvm-system", container="lvmd"}` kept increasing. After investigation, this was not lvmd, but rather an increase in the lvm cache. Therefore, the following metrics were not increasing
`container_memory_usage_bytes{namespace="topolvm-system", container="lvmd"} - container_memory_cache{namespace="topolvm-system", container="lvmd"}`

Once we've reported this issue to LVM mailing list. However, we don't have enought time to dig into this problem for now.
https://lore.kernel.org/linux-lvm/ZjO_tW3_5ZNhnHFm@redhat.com/T/

PR #943, opened 2024-07-30
Set lvmd log verbosity based on state modification or reporting

### LINKED ISSUE #941: Setting log verbosity to lvmd pod errors out
<!--
```
Defaulted container "lvmd" out of: lvmd, topolvm-metrics, volume-group-manager (init)
Error: unknown flag: --v
Usage:
unknown flag: --v
  lvmd [flags]
  lvmd [command]

```  -->

I want to contribute to this project to improve verbosity of the logs, which would help users to set verbosity logs to their desired level.  I want to get the opinions of the maintainers first. 
For example, following logs is too noisy:
```
{"level":"info","ts":"2024-07-19T04:17:03Z","msg":"invoking LVM command","args":["fullreport","--reportformat","json","--units","b","--nosuffix","--configreport","vg","-o","vg_name,vg_uuid,vg_size,vg_free","--configreport","lv","-o","lv_uuid,lv_name,lv_full_name,lv_path,lv_size,lv_kernel_major,lv_kernel_minor,origin,origin_size,pool_lv,lv_tags,lv_attr,vg_name,data_percent,metadata_percent,pool_lv","--configreport","pv","-o,","--configreport","pvseg","-o,","--configreport","seg","-o,"]}
{"level":"info","ts":"2024-07-19T04:17:03Z","msg":"invoking LVM command","args":["fullreport","--reportformat","json","--units","b","--nosuffix","--configreport","vg","-o","vg_name,vg_uuid,vg_size,vg_free","--configreport","lv","-o","lv_uuid,lv_name,lv_full_name,lv_path,lv_size,lv_kernel_major,lv_kernel_minor,origin,origin_size,pool_lv,lv_tags,lv_attr,vg_name,data_percent,metadata_percent,pool_lv","--configreport","pv","-o,","--configreport","pvseg","-o,","--configreport","seg","-o,"]}
```

### PR DESCRIPTION
**Pull request by Sunnatillo**
This commit adds verbosity to invoking command log. 

Fix #941. (edited by @ushitora-anqou )

### DIFF
--- a/internal/lvmd/command/lvm_command.go
+++ b/internal/lvmd/command/lvm_command.go
@@ -15,7 +15,9 @@ import (
 )
 
 const (
-	nsenter = "/usr/bin/nsenter"
+	nsenter                   = "/usr/bin/nsenter"
+	verbosityLVMStateUpdate   = 1
+	verbosityLVMStateNoUpdate = 4
 )
 
 var (
@@ -31,13 +33,13 @@ func SetLVMPath(path string) {
 
 // callLVM calls lvm sub-commands and prints the output to the log.
 func callLVM(ctx context.Context, args ...string) error {
-	return callLVMInto(ctx, nil, args...)
+	return callLVMInto(ctx, nil, verbosityLVMStateUpdate, args...)
 }
 
 // callLVMInto calls lvm sub-commands and decodes the output via JSON into the provided struct pointer.
 // if the struct pointer is nil, the output will be printed to the log instead.
-func callLVMInto(ctx context.Context, into any, args ...string) error {
-	output, err := callLVMStreamed(ctx, args...)
+func callLVMInto(ctx context.Context, into any, logVerbosity int, args ...string) error {
+	output, err := callLVMStreamed(ctx, logVerbosity, args...)
 	if err != nil {
 		return fmt.Errorf("failed to execute command: %v", err)
 	}
@@ -60,12 +62,12 @@ func callLVMInto(ctx context.Context, into any, args ...string) error {
 // callLVMStreamed calls lvm sub-commands and returns the output as a ReadCloser.
 // The caller is responsible for closing the ReadCloser, which will cause the command to complete.
 // Not calling close on this method will result in a resource leak.
-func callLVMStreamed(ctx context.Context, args ...string) (io.ReadCloser, error) {
+func callLVMStreamed(ctx context.Context, logVerbosity int, args ...string) (io.ReadCloser, error) {
 	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithCallDepth(1))
 	cmd := commandOnRootNS(lvm, args...)
 	cmd.Env = os.Environ()
 	cmd.Env = append(cmd.Env, "LC_ALL=C")
-	return runCommand(ctx, cmd)
+	return runCommand(ctx, logVerbosity, cmd)
 }
 
 // commandOnRootNS calls cmd with args on root NS.
@@ -78,7 +80,7 @@ func commandOnRootNS(cmd string, args ...string) *exec.Cmd {
 // runCommand runs the command and returns the stdout as a ReadCloser that also Waits for the command to finish.
 // After the Close command is called the cmd is closed and the resources are released.
 // Not calling close on this method will result in a resource leak.
-func runCommand(ctx context.Context, cmd *exec.Cmd) (io.ReadCloser, error) {
+func runCommand(ctx context.Context, logVerbosity int, cmd *exec.Cmd) (io.ReadCloser, error) {
 	stdout, err := cmd.StdoutPipe()
 	if err != nil {
 		return nil, err
@@ -89,7 +91,7 @@ func runCommand(ctx context.Context, cmd *exec.Cmd) (io.ReadCloser, error) {
 		return nil, err
 	}
 
-	log.FromContext(ctx).Info("invoking command", "args", cmd.Args)
+	log.FromContext(ctx).V(logVerbosity).Info("invoking command", "args", cmd.Args)
 	if err := cmd.Start(); err != nil {
 		_ = stdout.Close()
 		_ = stderr.Close()
--- a/internal/lvmd/command/lvm_command_test.go
+++ b/internal/lvmd/command/lvm_command_test.go
@@ -23,7 +23,7 @@ func Test_lvm_command(t *testing.T) {
 	}
 	t.Run("simple lvm version should succeed with stream", func(t *testing.T) {
 		ctx := log.IntoContext(context.Background(), testr.New(t))
-		dataStream, err := callLVMStreamed(ctx, "version")
+		dataStream, err := callLVMStreamed(ctx, verbosityLVMStateNoUpdate, "version")
 		if err != nil {
 			t.Fatal(err, "version should succeed")
 		}
@@ -79,7 +79,7 @@ func Test_lvm_command(t *testing.T) {
 		fakeDeviceName := "/dev/does-not-exist"
 
 		ctx := log.IntoContext(context.Background(), testr.New(t))
-		dataStream, err := callLVMStreamed(ctx, "vgcreate", "test-vg", fakeDeviceName)
+		dataStream, err := callLVMStreamed(ctx, verbosityLVMStateUpdate, "vgcreate", "test-vg", fakeDeviceName)
 		if err != nil {
 			t.Fatal(err, "vgcreate should not fail instantly as read didn't finish")
 		}
--- a/internal/lvmd/command/lvm_lvs.go
+++ b/internal/lvmd/command/lvm_lvs.go
@@ -123,7 +123,7 @@ func getLVReport(ctx context.Context, name string) (map[string]lv, error) {
 		"--reportformat",
 		"json",
 	}
-	err := callLVMInto(ctx, res, args...)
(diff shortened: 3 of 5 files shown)

### CODE REVIEW

**Comment by ushitora-anqou** [comment]
Could you adjust the logging verbosity only when the command being executed doesn't change the LVM's state, as @llamerada-jp mentioned [here](https://github.com/topolvm/topolvm/issues/941#issuecomment-2242432123)? I think we could add a new argument to the `runCommand` function to indicate whether the command will update the LVM's state or not.

**Comment by Sunnatillo** [comment]
> Could you adjust the logging verbosity only when the command being executed doesn't change the LVM's state, as @llamerada-jp mentioned [here](https://github.com/topolvm/topolvm/issues/941#issuecomment-2242432123)? I think we could add a new argument to the `runCommand` function to indicate whether the command will update the LVM's state or not.

Could you give me a hint how to know if LVM state is changed? Some comparison?  
Are we expected to get it from cmd output?

**Comment by Sunnatillo** [comment]
> Could you adjust the logging verbosity only when the command being executed doesn't change the LVM's state, as @llamerada-jp mentioned [here](https://github.com/topolvm/topolvm/issues/941#issuecomment-2242432123)? I think we could add a new argument to the `runCommand` function to indicate whether the command will update the LVM's state or not.

I have added new implementation when function is called from create, resize, remove chain it will set `shouldLog`  variable to `true`, when it is just listing LVS it will set `shouldLog` to `false`. 
Kindly check.

File: internal/lvmd/command/lvm_command.go [hunk 1]
```diff
@@ -78,7 +79,7 @@ func commandOnRootNS(cmd string, args ...string) *exec.Cmd {
 // runCommand runs the command and returns the stdout as a ReadCloser that also Waits for the command to finish.
 // After the Close command is called the cmd is closed and the resources are released.
 // Not calling close on this method will result in a resource leak.
-func runCommand(ctx context.Context, cmd *exec.Cmd) (io.ReadCloser, error) {
+func runCommand(ctx context.Context, shouldLog bool, cmd *exec.Cmd) (io.ReadCloser, error) {
```
jakobmoellerdev, line 82: how about you change the signature to a desired log level?

jakobmoellerdev (commented):
hmm this feels a little strange to me.

I think we should change this implementation to be a bit more idiomatic.

For example, I would expect log levels to be declared on a package / function level. Waiting for other opinions here

ushitora-anqou:
Overall, it looks good to me. Personally, I would prefer a name such as `doesUpdateLVMState` rather than `shouldLog`.  Also, I suggest defining and using some constants like these:

```go
const (
  UpdateLVMState = true
  DoesNotUpdateLVMState = false
)
```

I'd like to hear @llamerada-jp's opinions as well.

llamerada-jp:
I would prefer a method that is easy for developers and reviewers to understand. Using constants, as suggested by @ushitora-anqou, is a good idea. However, it might simplify the logic if log levels were constants instead of true/false values. Additionally, if we decide to subdivide the logs in the future, it would be easier to identify where changes need to be made.

Sunnatillo:
> hmm this feels a little strange to me.
> 
> I think we should change this implementation to be a bit more idiomatic.
> 
> For example, I would expect log levels to be declared on a package / function level. Waiting for other opinions here

Thank you for your review. I checked the function chain grouped them into three. As you can see on the picture functions that are group one and group three does not change the LVM state on group two functions change the LVM state. I used lower most function from each group to set verbosity variable. IMHO, passing verbosity variable from each function would be too much. But we can discuss it. 

Topolvm draw io

Sunnatillo:
@llamerada-jp, @ushitora-anqou. Thank you for your reviews. I have changed the code according to your suggestions

internal/lvmd/command/lvm_command_test.go
```diff
@@ -23,7 +23,7 @@ func Test_lvm_command(t *testing.T) {
 	}
 	t.Run("simple lvm version should succeed with stream", func(t *testing.T) {
 		ctx := log.IntoContext(context.Background(), testr.New(t))
-		dataStream, err := callLVMStreamed(ctx, "version")
+		dataStream, err := callLVMStreamed(ctx, 4, "version")
```
ushitora-anqou, line 26: Let's use `verbosityLVMStateNoUpdate` here.

internal/lvmd/command/lvm_command_test.go
```diff
@@ -79,7 +79,7 @@ func Test_lvm_command(t *testing.T) {
 		fakeDeviceName := "/dev/does-not-exist"
 
 		ctx := log.IntoContext(context.Background(), testr.New(t))
-		dataStream, err := callLVMStreamed(ctx, "vgcreate", "test-vg", fakeDeviceName)
+		dataStream, err := callLVMStreamed(ctx, 1, "vgcreate", "test-vg", fakeDeviceName)
```
ushitora-anqou, line 82: Let's use `verbosityLVMStateUpdate` here.

internal/lvmd/command/lvm_state_json.go
```diff
@@ -46,7 +46,8 @@ func getLVMState(ctx context.Context) ([]vg, []lv, error) {
 		"--configreport", "pvseg", "-o,",
 		"--configreport", "seg", "-o,",
 	}
-	streamed, err := callLVMStreamed(ctx, append([]string{"fullreport"}, args...)...)
+	logVerbosity := verbosityLVMStateNoUpdate
+	streamed, err := callLVMStreamed(ctx, logVerbosity, append([]string{"fullreport"}, args...)...)
```
ushitora-anqou, line 50:
```suggestion
	streamed, err := callLVMStreamed(ctx, verbosityLVMStateNoUpdate, append([]string{"fullreport"}, args...)...)
```

internal/lvmd/command/lvm_vgs.go
```diff
@@ -53,7 +53,8 @@ func getVGReport(ctx context.Context, name string) (vg, error) {
 	args := []string{
 		"vgs", name, "-o", "vg_uuid,vg_name,vg_size,vg_free", "--units", "b", "--nosuffix", "--reportformat", "json",
 	}
-	err := callLVMInto(ctx, res, args...)
+	logVerbosity := verbosityLVMStateUpdate
+	err := callLVMInto(ctx, res, logVerbosity, args...)
```
ushitora-anqou, line 57:
```suggestion
	err := callLVMInto(ctx, res, verbosityLVMStateUpdate, args...)
```

internal/lvmd/command/lvm_lvs.go
```diff
@@ -123,7 +123,8 @@ func getLVReport(ctx context.Context, name string) (map[string]lv, error) {
 		"--reportformat",
 		"json",
 	}
-	err := callLVMInto(ctx, res, args...)
+	logVerbosity := verbosityLVMStateNoUpdate
+	err := callLVMInto(ctx, res, logVerbosity, args...)
```
ushitora-anqou, line 127:
```suggestion
	err := callLVMInto(ctx, res, verbosityLVMStateNoUpdate, args...)
```

ushitora-anqou (changes requested):

internal/lvmd/command/lvm_command_test.go
Sunnatillo: I have addressed all your suggestions. Thanks

internal/lvmd/command/lvm_command_test.go
Sunnatillo: done

internal/lvmd/command/lvm_state_json.go
Sunnatillo: done

internal/lvmd/command/lvm_vgs.go
Sunnatillo: done

internal/lvmd/command/lvm_lvs.go
Sunnatillo: done

internal/lvmd/command/lvm_vgs.go
```diff
@@ -53,7 +53,7 @@ func getVGReport(ctx context.Context, name string) (vg, error) {
 	args := []string{
 		"vgs", name, "-o", "vg_uuid,vg_name,vg_size,vg_free", "--units", "b", "--nosuffix", "--reportformat", "json",
 	}
-	err := callLVMInto(ctx, res, args...)
+	err := callLVMInto(ctx, res, verbosityLVMStateUpdate, args...)
```
ushitora-anqou, line 56: This should be `verbosityLVMStateNoUpdate`, because `vgs` won't change the LVM's state.

ushitora-anqou (changes requested):

internal/lvmd/command/lvm_vgs.go
Sunnatillo: done

Sunnatillo: done

llamerada-jp (approved):

Sunnatillo:
@ushitora-anqou, could you please make approving review if it the changes fine for you? Change request is blocking the merge

satoru-takeuchi:
@Sunnatillo Today is a national holiday in Japan. I guess he'll get back to you tomorrow.

ushitora-anqou (approved):
Sorry for the late reply. (As @satoru-takeuchi kindly commented, I was on my days off.) The code now looks good to me. Thank you very much for your contirbution!

PR #947, opened 2024-08-12
Introduce prometheus metrics for lvmd process.

naemono:
## What is this change?

This is a bit of a follow-up to #931 , as we're wanting the `go_memstats*` information from prometheus for the lvmd pods to compare to some pprof dumps that we are analyzing, but found that the lvmd pods weren't exposing any metrics, in comparison to the other processes, which all expose these metrics (controller/node/etc).

### Open Question

Would your team prefer these flags to be similar to the other processes, such as `lvmd.metrics.enabled: bool`? I didn't do it this way as lvmd isn't using `controller-runtime`, so I wasn't sure if this made more sense. Let me know as I'll be happy to update it.

### DIFF
--- a/cmd/lvmd/app/root.go
+++ b/cmd/lvmd/app/root.go
@@ -10,9 +10,12 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"sync"
 	"syscall"
 	"time"
 
+	"github.com/go-logr/logr"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/spf13/cobra"
 	"github.com/topolvm/topolvm"
 	"github.com/topolvm/topolvm/internal/lvmd"
@@ -33,6 +36,7 @@ var (
 	lvmPath              string
 	zapOpts              zap.Options
 	profilingBindAddress string
+	metricsBindAddress   string
 )
 
 // rootCmd represents the base command when called without any subcommands
@@ -112,15 +116,7 @@ func subMain(parentCtx context.Context) error {
 	ctx, stop := signal.NotifyContext(parentCtx, os.Interrupt, syscall.SIGTERM)
 	defer stop()
 
-	var pprofServer *http.Server
-	if profilingBindAddress != "" {
-		pprofServer = profiling.NewProfilingServer(profilingBindAddress)
-		go func() {
-			if err := pprofServer.ListenAndServe(); !errors.Is(err, http.ErrServerClosed) {
-				logger.Error(err, "pprof server error")
-			}
-		}()
-	}
+	wg, pprofServer, metricsServer := startMetricsAndProfilingServers(logger)
 
 	go func() {
 		ticker := time.NewTicker(10 * time.Minute)
@@ -133,7 +129,13 @@ func subMain(parentCtx context.Context) error {
 						logger.Error(err, "failed to shutdown pprof server")
 					}
 				}
+				if metricsServer != nil {
+					if err := metricsServer.Shutdown(parentCtx); err != nil {
+						logger.Error(err, "failed to shutdown metrics server")
+					}
+				}
 				grpcServer.GracefulStop()
+				wg.Wait()
 				return
 			case <-ticker.C:
 				notifier()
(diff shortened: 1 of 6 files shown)


cmd/lvmd/app/root.go
```diff
@@ -122,6 +124,21 @@ func subMain(parentCtx context.Context) error {
...
+	var metricsServer *http.Server
+	if metricsBindAddress != "" {
+		mux := http.NewServeMux()
+		mux.Handle("/metrics", promhttp.Handler())
+		metricsServer = &http.Server{
+			Addr:    metricsBindAddress,
+			Handler: mux,
+		}
+		go func() {
+			if err := metricsServer.ListenAndServe(); !errors.Is(err, http.ErrServerClosed) {
```
jakobmoellerdev, line 136: What happens if the metrics Server doesnt shut down in time of a SIGTERM signal? I think we could bind this go routine to a context if possible. (Similar to the pprof endpoint below with a Shutdown loagic on the parent Ctx). Since this is quite some boilerplate it might make sense to start extracting these starts out into separate functions and using something like a wait group to orchestrate them all together.

In the current form, the start method is getting a little too long and its becoming hard to read due to different start / stop styles

jakobmoellerdev (changes requested):
This is a great addition. Thanks!

I suggest you add a test to https://github.com/topolvm/topolvm/blob/main/test/e2e/metrics_test.go to make sure these metrics are always exposed correctly.

llamerada-jp:
> Would your team prefer these flags to be similar to the other processes, such as lvmd.metrics.enabled: bool?

lvmd.metrics.enabled was added following PR. The objection of this PR is supporting PodMonitor for Prometheus Operator. I think many people will be happy if we make it the same way as this PR.

https://github.com/topolvm/topolvm/pull/373

The topolvm-node has metrics-bind-address command line argument, but this cannot be specified from the helm, and the pod definition uses a fixed 8080 port. Is there a problem with the same setup in lvmd? I would be happy to be able to change both if that is a problem.

When you turn on embedded mode, topolvm-node and lvmd run in the same process. At that time, metrics from lvmd may not be needed, so you should turn metrics off.

llamerada-jp:
> Would your team prefer these flags to be similar to the other processes, such as lvmd.metrics.enabled: bool?

The `lvmd.metrics.enabled` flag was added following a PR. The objective of this PR is to support PodMonitor for Prometheus Operator. I believe many people would appreciate it if we implemented it in the same manner as this PR.

https://github.com/topolvm/topolvm/pull/373

The topolvm-node has a metrics-bind-address command line argument, but this cannot be specified from Helm, and the pod definition uses a fixed 8080 port. Would there be any issues with having the same setup for lvmd? I would be happy to make changes to both if that would resolve any concerns.

When you enable embedded mode, topolvm-node and lvmd run in the same process. In such cases, metrics from lvmd may not be necessary, so it would be advisable to turn off the metrics.

toshipp (commented):
>Would your team prefer these flags to be similar to the other processes, such as lvmd.metrics.enabled: bool?

I agree with @llamerada-jp's opinion. Other components don't provide the parameter for their metric port, they open metrics HTTP handler always.

naemono:
@toshipp @llamerada-jp Sorry for the delays. I've updated this to be consistent with the other components, and will do some testing/verification, and also added an integration test for lvmd metrics.

@jakobmoellerdev I'm working through your suggestions now and will update as soon as I can.

cmd/lvmd/app/root.go
naemono: @jakobmoellerdev Ok, I've (at least) made the main startup method for lvmd shorter from my recent changes, but I hesitated to start extracting out pieces for starting these common http servers as I feel I'd be copying around the `sync.WaitGroup`, which seems to be an anti-pattern. The only other way I see to do this is to return a func from some other start* func, which is then called in a goroutine, using the waitgroup.

example
```
wd.Add(1)
fn := startHTTPServer(server)
go func() {
  defer wg.Done()
  fn()
}()
```

Any thoughts on the approach I've taken, vs what you were thinking? Thanks for any input before I go further down this path.

llamerada-jp (approved):

toshipp (commented):
seems good.

jakobmoellerdev (approved):

PR #956, opened 2024-08-27
make resizefs faster

toshipp:
This fixes the issue originally reported  by #898

### DIFF
--- a/constants.go
+++ b/constants.go
@@ -76,6 +76,9 @@ const PVCFinalizer = pluginName + "/pvc"
 // LegacyPVCFinalizer is a legacy finalizer of PVC.
 const LegacyPVCFinalizer = legacyPluginName + "/pvc"
 
+// LastResizeFSRequestedAtKey is the annotation key that represents the timestamp of the resizefs request to kubelet.
+const LastResizeFSRequestedAtKey = pluginName + "/last-resizefs-requested-at"
+
 // DefaultCSISocket is the default path of the CSI socket file.
 const DefaultCSISocket = "/run/topolvm/csi-topolvm.sock"
 
--- a/docs/topolvm-controller.md
+++ b/docs/topolvm-controller.md
@@ -1,10 +1,7 @@
 # topolvm-controller
 
-`topolvm-controller` provides a CSI controller service.  It also works as
-a custom Kubernetes controller to cleanup stale resources.
-
-Specifically, `topolvm-controller` watches `Node` resource deletion to
-cleanup `PersistentVolumeClaim` on the deleting Nodes.
+`topolvm-controller` provides a CSI controller service.
+It also works as a custom Kubernetes controller for additional tasks.
 
 ## CSI Controller Features
 
@@ -164,10 +161,9 @@ At step 4, the StatefulSet pod is not deleted if the PVC finalizer does not exis
 
 ### The Controller for Nodes
 
-The controller adds `topolvm.io/node` finalizer.
+The controller is to cleanup all PVCs and LogicalVolumes associating to the deleting Node.
 
-When a Node is being deleted, the controller deletes all PVCs and LogicalVolumes for TopoLVM
-on the deleting node. 
+It adds the `topolvm.io/node` finalizer to run the cleanup task.
 
 This node finalize procedure may be skipped with the `--skip-node-finalize` flag. 
 When this is true, the PVCs and the LogicalVolume CRs from a deleted node must be
@@ -175,20 +171,28 @@ deleted manually by a cluster administrator.
 
 ### The Controller for PersistentVolumeClams
 
-When a PVC for TopoLVM is being deleted, the controller waits for other
-finalizers to be completed.  Once it becomes the last finalizer, it removes
-the finalizer to immediately delete PVC then deletes pending pods referencing
-the deleted PVC, if any.
+The controller accomplishes two tasks.
+
+1. Delete Pods using PVCs under deletion.
+When a PVC for TopoLVM is being deleted, the controller deletes pods referencing
+the PVC, if any. This is repeated until other finalizers to be completed.
+Once it becomes the last finalizer, it removes the finalizer to immediately delete the PVC.
+
+2. Speed up resizing a PVC filesystem by nudging the kubelet.
+kubelet watches Pods rather than PVCs periodically to resize the filesystem,
+therefore the filesystem resizing may be delayed.
+To avoid this, the controller will notify kubelet by setting
+the `topolvm.io/last-resizefs-requested-at` annotation with the current time to the Pod.
 
 Command-line flags
 ------------------
 
-| Name                   | Type   | Default                                 | Description                                                                  |
-| ---------------------- |--------|-----------------------------------------|------------------------------------------------------------------------------|
-| `cert-dir`             | string | `/tmp/k8s-webhook-server/serving-certs` | Directory for `tls.crt` and `tls.key` files.                                 |
-| `csi-socket`           | string | `/run/topolvm/csi-topolvm.sock`         | UNIX domain socket of `topolvm-controller`.                                  |
-| `metrics-bind-address` | string | `:8080`                                 | Listen address for Prometheus metrics.                                       |
-| `secure-metrics-server`| bool   | `false`                                 | Secures the metrics server.                                                  |
-| `leader-election-id`   | string | `topolvm`                               | ID for leader election by controller-runtime.                                |
-| `webhook-addr`         | string | `:9443`                                 | Listen address for the webhook endpoint.                                     |
-| `skip-node-finalize`   | bool   | `false`                                 | When true, skips automatic cleanup of PhysicalVolumeClaims on Node deletion. |
+| Name                    | Type   | Default                                 | Description                                                                  |
+| ----------------------- | ------ | --------------------------------------- | ---------------------------------------------------------------------------- |
+| `cert-dir`              | string | `/tmp/k8s-webhook-server/serving-certs` | Directory for `tls.crt` and `tls.key` files.                                 |
+| `csi-socket`            | string | `/run/topolvm/csi-topolvm.sock`         | UNIX domain socket of `topolvm-controller`.                                  |
+| `metrics-bind-address`  | string | `:8080`                                 | Listen address for Prometheus metrics.                                       |
+| `secure-metrics-server` | bool   | `false`                                 | Secures the metrics server.                                                  |
+| `leader-election-id`    | string | `topolvm`                               | ID for leader election by controller-runtime.                                |
+| `webhook-addr`          | string | `:9443`                                 | Listen address for the webhook endpoint.                                     |
+| `skip-node-finalize`    | bool   | `false`                                 | When true, skips automatic cleanup of PhysicalVolumeClaims on Node deletion. |
--- a/internal/controller/persistentvolumeclaim_controller.go
+++ b/internal/controller/persistentvolumeclaim_controller.go
@@ -32,16 +32,7 @@ func NewPersistentVolumeClaimReconciler(client client.Client, apiReader client.R
 //+kubebuilder:rbac:groups=core,resources=persistentvolumeclaims,verbs=get;list;watch;update
 //+kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch;delete
 
-// Reconcile finalize PVC
-//
-// This was originally created as a workaround for the following issue.
-// https://github.com/kubernetes/kubernetes/pull/93457
-//
-// Because the issue was fixed, the PVC reconciler was once removed by PR #536.
-// However, it turned out that the PVC finalizer was also useful to
-// resolve the issue that a pod created from StatefulSet persists in the PENDING state
-// when PVC is deleted for some reasons like node deletion.
-// Thus, the reconciler was revived by PR #620.
+// Reconcile PVC
 func (r *PersistentVolumeClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	log := crlog.FromContext(ctx)
 	// your logic here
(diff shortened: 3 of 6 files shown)


internal/controller/persistentvolumeclaim_controller.go
```diff
@@ -66,16 +57,57 @@ func (r *PersistentVolumeClaimReconciler) Reconcile(ctx context.Context, req ctr
...
+	// Manually annotating pods is a way to speed up the process.
+	for _, cond := range pvc.Status.Conditions {
+		if cond.Type == corev1.PersistentVolumeClaimFileSystemResizePending && cond.Status == corev1.ConditionTrue {
+			pods, err := r.getPodsByPVC(ctx, pvc)
+			if err != nil {
+				log.Error(err, "unable to fetch PodList for a PVC", "pvc", pvc.Name, "namespace", pvc.Namespace)
+				return err
+			}
+			for _, pod := range pods {
+				pod.Annotations["topolvm.io/resizefs-notify"] = time.Now().String()
```
jakobmoellerdev, line 86: I think we should set this annotation with a docs entry in addition to the time to make it clear why we do this. apart from that I think its a good idea and we should keep it

toshipp: Do you mean adding a docs annotation to pods?

jakobmoellerdev: I was thinking of changing the current annotation because its not used anywhere else and serves only as marker. Also I think we should add it as part of our normal docs

toshipp: IC. This is a draft, I'm going to work on it more. When I'm done, I'll assign you as a reviewer.

constants.go
```diff
@@ -76,6 +76,9 @@ const PVCFinalizer = pluginName + "/pvc"
 // LegacyPVCFinalizer is a legacy finalizer of PVC.
 const LegacyPVCFinalizer = legacyPluginName + "/pvc"
 
+// ResizeFSRequestedAtKey is the annotation key that represents the timestamp of the resizefs request to kubelet.
+const ResizeFSRequestedAtKey = pluginName + "/resizefs-requested-at"
```
jakobmoellerdev, line 80:
```suggestion
const LastResizeFSRequestedAtKey = pluginName + "/last-resizefs-requested-at"
```
I think the last prefix can be nice here because it will be confusing if ops think that the resize is currently ongoing due to this annotation

docs/topolvm-controller.md
```diff
@@ -164,31 +161,38 @@ At step 4, the StatefulSet pod is not deleted if the PVC finalizer does not exis
 
 ### The Controller for Nodes
 
-The controller adds `topolvm.io/node` finalizer.
+The controller is to cleanup all PVCs and LogicalVolumes associating to the deleting Node.
 
-When a Node is being deleted, the controller deletes all PVCs and LogicalVolumes for TopoLVM
-on the deleting node. 
+It adds `topolvm.io/node` finalizer to run the cleanup task.
```
jakobmoellerdev, line 166:
```suggestion
It adds the `topolvm.io/node` finalizer to run the cleanup task.
```

docs/topolvm-controller.md
```diff
@@ -164,31 +161,38 @@ At step 4, the StatefulSet pod is not deleted if the PVC finalizer does not exis
...
 When this is true, the PVCs and the LogicalVolume CRs from a deleted node must be
 deleted manually by a cluster administrator.
 
 ### The Controller for PersistentVolumeClams
 
-When a PVC for TopoLVM is being deleted, the controller waits for other
-finalizers to be completed.  Once it becomes the last finalizer, it removes
-the finalizer to immediately delete PVC then deletes pending pods referencing
-the deleted PVC, if any.
+The controller is to accomplish two tasks.
```
jakobmoellerdev, line 174:
```suggestion
The controller accomplishes two tasks.
```

docs/topolvm-controller.md
```diff
@@ -164,31 +161,38 @@ At step 4, the StatefulSet pod is not deleted if the PVC finalizer does not exis
...
+
+The first is to delete Pods using deleting PVCs.
```
jakobmoellerdev, line 176:
```suggestion
1. Delete Pods using PVCs under deletion.
```

docs/topolvm-controller.md
```diff
@@ -164,31 +161,38 @@ At step 4, the StatefulSet pod is not deleted if the PVC finalizer does not exis
...
+When a PVC for TopoLVM is being deleted, the controller deletes pods referencing
+the PVC, if any. This is repeated until other finalizers to be completed.
+Once it becomes the last finalizer, it removes the finalizer to immediately delete the PVC.
+
+The second is to speed up resizing a filesystem by kubelet.
```
jakobmoellerdev, line 181:
```suggestion
2. Speed up resizing a PVC filesystem by nudging the kubelet.
```

docs/topolvm-controller.md
```diff
@@ -164,31 +161,38 @@ At step 4, the StatefulSet pod is not deleted if the PVC finalizer does not exis
...
+kubelet watches Pods rather than PVCs periodically to resize the filesystem,
+therefore the filesystem resizing may be delayed.
+To avoid this, the controller will notify kubelet by setting
+the `topolvm.io/resizefs-requested-at` annotation with the current time to the Pod.
```
jakobmoellerdev, line 185:
```suggestion
the `topolvm.io/last-resizefs-requested-at` annotation with the current time to the Pod.
```

internal/controller/persistentvolumeclaim_controller.go
```diff
@@ -66,16 +58,81 @@ func (r *PersistentVolumeClaimReconciler) Reconcile(ctx context.Context, req ctr
...
+		err := r.notifyKubeletToResizeFS(ctx, log, pvc)
+		return ctrl.Result{}, err
+	}
+
+	result, err := r.deletePodsUsingDeletingPVC(ctx, log, pvc)
+	if result.Requeue || err != nil {
+		return result, err
+	}
+
+	controllerutil.RemoveFinalizer(pvc, topolvm.PVCFinalizer)
```
jakobmoellerdev, line 76:
```suggestion
	updated := controllerutil.RemoveFinalizer(pvc, topolvm.PVCFinalizer)
```
you can use updated to conditionally remove the finalizer, should (almost) never happen, but still good practice

internal/controller/persistentvolumeclaim_controller.go
```diff
@@ -66,16 +58,81 @@ func (r *PersistentVolumeClaimReconciler) Reconcile(ctx context.Context, req ctr
 		}, nil
 	}
 
-	// Skip if the PVC is not deleted or PVC does not have TopoLVM's finalizer.
-	if pvc.DeletionTimestamp == nil {
+	// Skip if PVC does not have TopoLVM's finalizer.
+	if !controllerutil.ContainsFinalizer(pvc, topolvm.PVCFinalizer) {
```
jakobmoellerdev, line 62: maybe a quick note here on when the finalizer is set would be good.

internal/controller/persistentvolumeclaim_controller.go
```diff
@@ -66,16 +58,81 @@ func (r *PersistentVolumeClaimReconciler) Reconcile(ctx context.Context, req ctr
...
+	return nil
 
+needResizing:
+	pods, err := r.getPodsByPVC(ctx, pvc)
+	if err != nil {
+		log.Error(err, "unable to fetch PodList for a PVC", "pvc", pvc.Name, "namespace", pvc.Namespace)
+		return err
+	}
+	for _, pod := range pods {
+		resizeFSRequestedAt, err := time.Parse(time.RFC3339, pod.Annotations[topolvm.ResizeFSRequestedAtKey])
```
jakobmoellerdev, line 107: this can end up in a nil pointer if pod.Annotations is nil I think you need to move line 112-114 above this

internal/controller/persistentvolumeclaim_controller.go
```diff
@@ -66,16 +58,81 @@ func (r *PersistentVolumeClaimReconciler) Reconcile(ctx context.Context, req ctr
...
+
+	if err := r.client.Update(ctx, pvc); err != nil {
+		log.Error(err, "failed to remove finalizer", "name", pvc.Name)
+		return ctrl.Result{}, err
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *PersistentVolumeClaimReconciler) notifyKubeletToResizeFS(ctx context.Context, log logr.Logger, pvc *corev1.PersistentVolumeClaim) error {
```
jakobmoellerdev, line 86:
```suggestion
func (r *PersistentVolumeClaimReconciler) notifyKubeletToResizeFS(ctx context.Context, pvc *corev1.PersistentVolumeClaim) error {
logger := crlog.FromContext(ctx)
```

internal/controller/persistentvolumeclaim_controller.go
```diff
@@ -66,16 +58,81 @@ func (r *PersistentVolumeClaimReconciler) Reconcile(ctx context.Context, req ctr
...
+			condTransitionAt = cond.LastTransitionTime.Time
+			goto needResizing
+		}
 	}
+	return nil
 
+needResizing:
+	pods, err := r.getPodsByPVC(ctx, pvc)
+	if err != nil {
+		log.Error(err, "unable to fetch PodList for a PVC", "pvc", pvc.Name, "namespace", pvc.Namespace)
```
jakobmoellerdev, line 103:
```suggestion
		log.Error(err, "unable to fetch PodList for a PVC")
```
the controller logger should already include a resource reference to the PVC because thats the object it reconciles

internal/controller/persistentvolumeclaim_controller_test.go
```diff
@@ -555,4 +555,98 @@ var _ = Describe("PersistentVolumeClaimController controller", func() {
...
+		err = k8sClient.Create(ctx, &pvc)
+		Expect(err).NotTo(HaveOccurred())
+
+		By("Checking a precondition that the Pod doesn't have annotation")
+		Eventually(func(g Gomega) {
+			var got corev1.Pod
+			err := k8sClient.Get(ctx, client.ObjectKey{Namespace: pod.Namespace, Name: pod.Name}, &got)
+			g.Expect(err).NotTo(HaveOccurred())
+			g.Expect(got.Annotations).ShouldNot(HaveKey("topolvm.io/resizefs-requested-at"))
+		}).Should(Succeed())
```
jakobmoellerdev, line 623:
```suggestion
		}).WithContext(ctx).Should(Succeed())
```
I think its best practice to pass the context to all eventuallies: https://onsi.github.io/ginkgo/#using-speccontext-with-gomegas-eventually

jakobmoellerdev (commented):
Mostly looks good, just minor comments. Approach is great!

constants.go
toshipp: I referred to [another annotation](https://github.com/topolvm/topolvm/blob/main/constants.go#L55), but I agree with your opinion.
The existing one is added only so that it can be renamed.
I'll rename them as your suggestion.

internal/controller/persistentvolumeclaim_controller.go
toshipp: It would be better.
btw, I didn't know nil map can be read :)
https://go.dev/play/p/_LEM2BUmBPj

internal/controller/persistentvolumeclaim_controller.go
toshipp: I wrote it as you suggested, but I felt it was redundant.
Could you tell me why you prefer it?

constants.go
toshipp: Renaming the existing one strays off this, I'll make another PR.

internal/controller/persistentvolumeclaim_controller.go
jakobmoellerdev: I think its best practice nowadays to only pass context and retrieve logger from context to have less parameter bloat, really just a nit that I picked up over the years. Feel free to change back if you feel strongly about it

internal/controller/persistentvolumeclaim_controller.go
jakobmoellerdev: Interestingly I didnt know either. Also it seems like the time.Parse thus doesnt crash but just returns an error. https://go.dev/play/p/ARuTL3Gfw7M Well I never stop learning :D

internal/controller/persistentvolumeclaim_controller.go
toshipp: Ok. I adapt it to the style.

internal/controller/persistentvolumeclaim_controller.go
```diff
@@ -66,19 +57,88 @@ func (r *PersistentVolumeClaimReconciler) Reconcile(ctx context.Context, req ctr
...
+		log.Error(err, "unable to fetch PodList for a PVC")
+		return err
+	}
+	for _, pod := range pods {
+		if pod.Annotations == nil {
+			pod.Annotations = map[string]string{}
+		}
+
+		lastresizeFSRequestedAt, err := time.Parse(time.RFC3339Nano, pod.Annotations[topolvm.LastResizeFSRequestedAtKey])
+		if err != nil && lastresizeFSRequestedAt.After(condTransitionAt) {
```
peng225, line 113: When `err != nil` is true, `lastresizeFSRequestedAt` will be zero value.
If so, why is it compared with `condTransitionAt`?

internal/controller/persistentvolumeclaim_controller.go
```diff
@@ -66,19 +57,88 @@ func (r *PersistentVolumeClaimReconciler) Reconcile(ctx context.Context, req ctr
...
+			continue
+		}
```
peng225, line 115: The case where `err != nil` is true and `lastresizeFSRequestedAt.After(condTransitionAt)` is false is ignored.
I think we need an error handling.

internal/controller/persistentvolumeclaim_controller.go
```diff
@@ -66,19 +57,88 @@ func (r *PersistentVolumeClaimReconciler) Reconcile(ctx context.Context, req ctr
...
+
+	result, err := r.deletePodsUsingDeletingPVC(ctx, pvc)
+	if result.Requeue || err != nil {
+		return result, err
+	}
+
+	updated := controllerutil.RemoveFinalizer(pvc, topolvm.PVCFinalizer)
+	if updated {
+		if err := r.client.Update(ctx, pvc); err != nil {
+			log.Error(err, "failed to remove finalizer", "name", pvc.Name)
```
peng225, line 78:
```suggestion
			log.Error(err, "failed to remove finalizer")
```

For the same reason as 78868e29723289704fdf173fe0380865819d4176.

internal/controller/persistentvolumeclaim_controller.go
toshipp: My bad. I intended to compare them when the parsing successes.

internal/controller/persistentvolumeclaim_controller.go
toshipp: In this case, someone or a bug may set a wrong value to the annotation. Therefore, we can not trust the value, simply add a new one. This is the same logic as the happy path.

peng225 (dismissed):

internal/controller/persistentvolumeclaim_controller.go
toshipp: I found we checked the PVC has a finalizer on line 62, the condition must be true. I leave it as before.

toshipp:
@jakobmoellerdev @peng225 
I squashed some commits. Could you recheck it?

jakobmoellerdev (approved):
LGTM, great results!

peng225 (approved):

PR #960, opened 2024-09-03
Add ObjectSelector Configuration for Webhooks

dsvetl:
## Summary

This pull request introduces the capability to configure `objectSelector` for both `podMutatingWebhook` and `pvcMutatingWebhook` in the Helm chart. This enhancement allows users to specify which Pods or PVCs should be targeted by the mutating webhooks based on specific labels, improving the flexibility and control of webhook operations.

## Key Changes

- **Pod MutatingWebhook**:
  - Added an `objectSelector` configuration to the YAML template, enabling users to define specific labels that Pods must match for the webhook to take action.
  - Updated `values.yaml` to include the `objectSelector` for `podMutatingWebhook`, with the default set to an empty object.

- **PVC MutatingWebhook**:
  - Added an `objectSelector` configuration to the YAML template, allowing users to specify labels that PVCs must match for the webhook to take action.
  - Updated `values.yaml` to include the `objectSelector` for `pvcMutatingWebhook`, with the default set to an empty object.

## Documentation

For more details on how `objectSelector` works and how to configure it, please refer to the Kubernetes documentation on [Matching Requests: ObjectSelector](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-objectselector).

## Additional Information

- This update maintains backward compatibility by setting the `objectSelector` defaults to empty objects, ensuring that existing configurations will not be affected unless explicitly modified by the user.

### DIFF
--- a/charts/topolvm/values.yaml
+++ b/charts/topolvm/values.yaml
@@ -698,12 +698,14 @@ webhook:
     # webhook.podMutatingWebhook.enabled -- Enable Pod MutatingWebhook.
     enabled: false
     # webhook.podMutatingWebhook.objectSelector -- Labels required on Pods for webhook action.
+    # WARNING: Modifying objectSelector can affect TopoLVM Pod scheduling. Proceed with caution.
     objectSelector: {}
       # webhook: topolvm
   pvcMutatingWebhook:
     # webhook.pvcMutatingWebhook.enabled -- Enable PVC MutatingWebhook.
     enabled: true
     # webhook.pvcMutatingWebhook.objectSelector -- Labels required on PVCs for webhook action.
+    # WARNING: Modifying objectSelector can affect TopoLVM PVC management. Proceed with caution.
     objectSelector: {}
       # webhook: topolvm
 
--- a/charts/topolvm/values.yaml
+++ b/charts/topolvm/values.yaml
@@ -697,9 +697,15 @@ webhook:
   podMutatingWebhook:
     # webhook.podMutatingWebhook.enabled -- Enable Pod MutatingWebhook.
     enabled: false
+    # webhook.podMutatingWebhook.objectSelector -- Labels required on Pods for webhook action.
+    objectSelector: {}
+      # webhook: topolvm
   pvcMutatingWebhook:
     # webhook.pvcMutatingWebhook.enabled -- Enable PVC MutatingWebhook.
     enabled: true
+    # webhook.pvcMutatingWebhook.objectSelector -- Labels required on PVCs for webhook action.
+    objectSelector: {}
+      # webhook: topolvm
 
 # Container Security Context
 # ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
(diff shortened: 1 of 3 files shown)


daichimukai:
Hi @dsvetl,

Thank you for your PR. I have a question and a suggestion:

1. What are the concrete use cases for this feature?
2. The Pod mutating webhook of TopoLVM is one of the core parts of scheduling. I'd like to emphasize that manually editing the `objectSelector` must be done with sufficient knowledge of what is being done, as it might break TopoLVM's functionality.

dsvetl:
I completely agree with your concern regarding the critical nature of the Pod MutatingWebhook for TopoLVM's scheduling. Misconfiguring the objectSelector could indeed interfere with the proper scheduling of Pods using TopoLVM.

I see this feature as an advanced use case of `MutatingWebhookConfiguration`\`s for environments with multiple StorageClasses. It allows administrators to prevent unnecessary webhook calls from affecting other controllers. In my scenario, the objectSelector can be used to reduce the number of calls to the topolvm-controller for PVCs that are not managed by TopoLVM, ensuring that the webhooks only operate on relevant resources.

charts/topolvm/values.yaml
```diff
@@ -697,9 +697,15 @@ webhook:
   podMutatingWebhook:
     # webhook.podMutatingWebhook.enabled -- Enable Pod MutatingWebhook.
     enabled: false
+    # webhook.podMutatingWebhook.objectSelector -- Labels required on Pods for webhook action.
```
daichimukai, line 700: As I previously mentioned in https://github.com/topolvm/topolvm/pull/960#issuecomment-2328087818, we'd like to emphasize that manually editing this field requires close attention. Could you add a warning comment?

charts/topolvm/values.yaml
```diff
@@ -697,9 +697,15 @@ webhook:
...
+    objectSelector: {}
+      # webhook: topolvm
   pvcMutatingWebhook:
     # webhook.pvcMutatingWebhook.enabled -- Enable PVC MutatingWebhook.
     enabled: true
+    # webhook.pvcMutatingWebhook.objectSelector -- Labels required on PVCs for webhook action.
```
daichimukai, line 706: Same here. Although the TopoLVM's PVC mutating webhook currently does not play such a critical role, most users won't need to change this field.

daichimukai (changes requested):

charts/topolvm/values.yaml
dsvetl: Yes, I've added warning and put reference into README.md

charts/topolvm/values.yaml
dsvetl: Yes, I've added warning and put reference into README.md

pluser (changes requested):
The comments in the Helm chart and the description in README.md must match. It would be good if you could check.

dsvetl:
> The comments in the Helm chart and the description in README.md must match. It would be good if you could check.

Thanks for pointing, I found that it is should be generated via `helm-docs` and fixed.

pluser (approved):

daichimukai (approved):
LGTM

PR #961, opened 2024-09-06
fix: lvmd: use 0 as free bytes of thin pool if usage exceeds total capacity

ushitora-anqou:
In the current implementation, the free capacity of a thin pool is calcualted by subtracting its usage from the overprovisioned total capacity.  Since both usage and capacity are represented as unsigned integers, the result wraps around if the usage somehow exceeds the capacity.

This update addresses the issue by comparing the two values before performing the subtraction.

See also: https://github.com/topolvm/topolvm/discussions/958

### DIFF
--- a/internal/lvmd/util.go
+++ b/internal/lvmd/util.go
@@ -1,0 +1,11 @@
+package lvmd
+
+import "math"
+
+func calcThinPoolFreeBytes(overProvisionRatio float64, tpuSizeBytes, tpuVirtualBytes uint64) uint64 {
+	virtualPoolSize := uint64(math.Floor(overProvisionRatio * float64(tpuSizeBytes)))
+	if virtualPoolSize <= tpuVirtualBytes {
+		return 0
+	}
+	return virtualPoolSize - tpuVirtualBytes
+}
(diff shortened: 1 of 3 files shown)


internal/lvmd/util.go
```diff
@@ -0,0 +1,11 @@
+package lvmd
+
+import "math"
+
+func calcThinPoolFreeBytes(overProvisionRatio float64, tpuSizeBytes, tpuVirtualBytes uint64) uint64 {
+	poolSize := uint64(math.Floor(overProvisionRatio * float64(tpuSizeBytes)))
```
peng225, line 6: I wonder if `poolSize` really represents the pool size.
IIUC, `tpuSizeBytes` is the pool size, so we need another name for`uint64(math.Floor(overProvisionRatio * float64(tpuSizeBytes)))`.

How do you think about changing it to something like `virtualPoolSize`, `effectivePoolSize`, or `overprovisionedPoolSize`?

ushitora-anqou: Thank you for your review. I've just fixed this, so could you review this PR again?

pluser (approved):

peng225 (approved):

PR #969, opened 2024-10-03
Fix empty line

jurim76:


jurim76:
Seems like the chart has a lot of empty lines everywhere

Issue #983, opened 2024-10-28
Old version in cert-manager sub chart 1.7.0


heinzelmannm:
**Describe the bug**
In the subchart for installing cert-manager a very old version will be deployed. The current release is 1.16.1

**Environments**

**To Reproduce**
1. Enable the cert-manager subchart in the values file

**Expected behavior**
A more current version should be installed.

**Additional context**

toshipp:
closed via #985

PR #985, opened 2024-10-29
  Very old version of cert-manager in the subchart - updated to v1.12  Update for Issue. #983

heinzelmannm:
This replaces #984

Very old version of cert-manager in the subchart - updated to v1.16.1

Update for Issue. https://github.com/topolvm/topolvm/issues/983

### DIFF
--- a/charts/topolvm/Chart.yaml
+++ b/charts/topolvm/Chart.yaml
@@ -12,7 +12,7 @@ sources:
 dependencies:
   - name: cert-manager
     repository: https://charts.jetstack.io
-    version: 1.7.0
+    version: ~1.12.x
     condition: cert-manager.enabled
 
 annotations:
--- a/charts/topolvm/Chart.lock
+++ b/charts/topolvm/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: cert-manager
   repository: https://charts.jetstack.io
-  version: v1.7.0
-digest: sha256:2d72eb34ee783a5a3e710366d7c50cdaaf9197613b3904907798e7a13dd39f0f
-generated: "2022-01-28T07:01:39.942953089Z"
+  version: v1.12.13
+digest: sha256:92b3de762998e5b933e95852d94dda91d998eb7b8bd4fef45bbdfd4ec344940c
+generated: "2024-11-06T09:47:35.354991961+01:00"


daichimukai:
@heinzelmannm 
You don't need to recreate a pull request every time you revise your commits. Just use force push so we can easily keep track of the PR status and previous discussions. Thanks.

heinzelmannm:
> @heinzelmannm You don't need to recreate a pull request every time you revise your commits. Just use force push so we can easily keep track of the PR status and previous discussions. Thanks.

@daichimukai thanks for the hint. I will keep that in the mind next time!

peng225:
@heinzelmannm 
Thank you for your PR.
We have some concerns about upgrading cert-manager to the latest version.
I'll run some tests to check the upgrade behavior, so I'd appreciate it if you could wait a little while.

heinzelmannm:
> @heinzelmannm Thank you for your PR. We have some concerns about upgrading cert-manager to the latest version. I'll run some tests to check the upgrade behavior, so I'd appreciate it if you could wait a little while.

@peng225 Thanks :) for me it's fine and to be honest, from my side the focus was about a fresh installation which is quite more easy than upgrading if cert-manager is already deployed in v1.7.0

If I can help, please let me know.

charts/topolvm/Chart.yaml
```diff
@@ -12,7 +12,7 @@ sources:
 dependencies:
   - name: cert-manager
     repository: https://charts.jetstack.io
-    version: 1.7.0
+    version: 1.16.1
```
peng225, line 15: We have decided to use the LTS version of cert-manager.
ref. https://cert-manager.io/docs/releases/#currently-supported-releases

We have also decided to follow [the best practice of the helm chart](https://helm.sh/docs/chart_best_practices/dependencies/#versions) and use the version range instead of the exact version.

Could you change the code as follows?
```suggestion
    version: ~1.12.x
```

cf. https://github.com/Masterminds/semver?tab=readme-ov-file#tilde-range-comparisons-patch

peng225 (changes requested):

charts/topolvm/Chart.yaml
heinzelmannm: @peng225 looks good to me, I've updated the commit. What bothers me a little about version 12 is the known error (https://cert-manager.io/docs/releases/release-notes/release-notes-1.12/#known-issues) but since we are working with multiple secrets here, that should be fine.

ushitora-anqou: Could you run `helm dependency update charts/topolvm/` to upgrade `charts/topolvm/Chart.lock`? Also, it would be great if you could squash your commits into one. (If you're not sure how to do that, you can check out https://stackoverflow.com/a/5201642, for example.) Thanks!

heinzelmannm: @ushitora-anqou OK, I've updated the Chart.lock, too and squashed the commits into one, thank you for that hint!

peng225:
@heinzelmannm 
Could you update the title of this PR? It says "updated to v1.16.1", but it should be "updated to v1.12".

heinzelmannm:
> @heinzelmannm Could you update the title of this PR? It says "updated to v1.16.1", but it should be "updated to v1.12".

@peng225: Sorry for the circumstances for this little change. I've updated the title .

peng225 (approved):

ushitora-anqou (approved):

ushitora-anqou:

PR #1005, opened 2025-01-08
test: don't skip tests using root unless SKIP_TESTS_USING_ROOT is set

ushitora-anqou:
Related to #999

In the current implementation, some tests are skipped when they run without root privilege, so developers can overlook them easily.

This commit resolves the problem above by introducing SKIP_TESTS_USING_ROOT environment variable. The behaviour of the tests is changed as follows:

- If the tests are started with root privilege, they will run.
- If the tests are started without root privilege and without SKIP_TESTS_USING_ROOT=1, they will fail.
- If the tests are started without root privilege but with SKIP_TESTS_USING_ROOT=1, they will be skipped.

### DIFF
--- a/internal/lvmd/testutils/test_utils.go
+++ b/internal/lvmd/testutils/test_utils.go
@@ -6,7 +6,9 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+	"testing"
 
+	"github.com/topolvm/topolvm/internal/testutils"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
@@ -76,4 +78,8 @@ func CleanLoopbackVG(name string, loops []string, files []string) error {
 	}
 
 	return nil
+}
+
+func RequireRoot(t *testing.T) {
+	testutils.RequireRoot(t)
 }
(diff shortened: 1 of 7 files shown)


internal/lvmd/testutils/test_utils.go
```diff
@@ -77,3 +79,7 @@ func CleanLoopbackVG(name string, loops []string, files []string) error {
 
 	return nil
 }
+
+func RequireRoot(t *testing.T) {
```
jakobmoellerdev, line 83: Why is it necessary to pass this require root here? Cant we just reuse the internal package directly?

jakobmoellerdev (commented):
Looks okay to me except for the double require root call.

internal/lvmd/testutils/test_utils.go
ushitora-anqou: In theory, we could, but both `internal/lvmd/testutils` and `internal/testutils` are packaged under the same name, `testutils`. To use them separately in the `internal/lvmd/*_test.go` files, we'd have to rename one of them in the import statement. That's a bit of a hassle, so I decided to include `RequireRoot` here instead.

pluser (approved):

daichimukai (approved):

PR #1014, opened 2025-02-12
Remove checking some k8s.io dependencies

toshipp:
Since controller-runtime depends on them, there is no need to check themselves.


pluser (approved):
The possibility of these dependencies being removed in the future is very low. So I think it's good. 👍

llamerada-jp (approved):

PR #1020, opened 2025-03-12
Helm Chart: allow additional containers to be defined in the chart

beschoenen:
The PR allows for additional containers to be added to all DaemonSets and Deployments in the chart.

### DIFF
--- a/charts/topolvm/templates/controller/deployment.yaml
+++ b/charts/topolvm/templates/controller/deployment.yaml
@@ -258,6 +258,10 @@ spec:
           {{- with .Values.env.liveness_probe }}
           env: {{ toYaml . | nindent 12 }}
           {{- end }}
+
+        {{- with .Values.controller.additionalContainers }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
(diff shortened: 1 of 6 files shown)


charts/topolvm/templates/controller/deployment.yaml
```diff
@@ -258,6 +258,10 @@ spec:
           {{- with .Values.env.liveness_probe }}
           env: {{ toYaml . | nindent 12 }}
           {{- end }}
+
+        {{- if .Values.controller.additionalContainers }}
```
ushitora-anqou, line 262: Could you use `with` instead of `if` for readability, like other fields?

ushitora-anqou (changes requested):

charts/topolvm/templates/controller/deployment.yaml
beschoenen: done

toshipp (dismissed):

ushitora-anqou (approved):

toshipp (approved):

PR #1025, opened 2025-03-21
docs: add a note about nodeName for pod

llamerada-jp:
We have found a limitation that the pod cannot be started when nodeName is specified. This behavior is due to kube scheduler, and cannot be fixed in TopoLVM. Therefore, we will create faq.md and add a note.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -7,6 +7,11 @@ assignees: ''
 
 ---
 
+<!-- Please check the limitations and FAQ before posting a bug report.
+  https://github.com/topolvm/topolvm/blob/main/docs/limitations.md
+  https://github.com/topolvm/topolvm/blob/main/docs/faq.md
+  -->
+
 **Describe the bug**
 A clear and concise description of what the bug is.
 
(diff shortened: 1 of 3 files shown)


docs/limitations.md
```diff
@@ -98,4 +99,31 @@ In the past, we used Ubuntu 18.04 as a base image and used older xfsprogs whenev
...
 This was originally introduced to support changes for cloud-providers where storage-class attributes might change ([see the PR for implementation details](https://github.com/kubernetes-csi/external-provisioner/pull/699)) during the restore process, however this doesn't apply for TopoLVM.
 
-Thus, if a pod consumes a restored/cloned PV having a different storage class from the original PV, this pod will not get scheduled if the StorageClass contents differ from the source StorageClass (e.g. by using a different device class).
\ No newline at end of file
+Thus, if a pod consumes a restored/cloned PV having a different storage class from the original PV, this pod will not get scheduled if the StorageClass contents differ from the source StorageClass (e.g. by using a different device class).
+
+## The Pod does not start when nodeName is specified  in the Pod spec
```
daichimukai, line 104:
```suggestion
## The Pod does not start when nodeName is specified in the Pod spec
```

daichimukai (commented):
In my understanding, this is more of a natural consequence of the specifications of Kubernetes itself rather than a limitation of TopoLVM. I believe that addressing this in an FAQ or similar resource would be appropriate. What do you think?

pluser (approved):

daichimukai (approved):

pluser (approved):

Issue #1039, opened 2025-04-25
Taking a snapshot of byte-precision sized PVC fails


qJkee:
**Describe the bug**
When trying to create a snapshot with exact bytes specified, it fails with error
`rpc error: code = OutOfRange desc = requested size 1136238592 is smaller than source logical volume: 1136656384` during this [check](https://github.com/topolvm/topolvm/blob/main/internal/lvmd/lvservice.go#L239). Probably this is happens because of LVM which rounds up the storage, because i see the following message in the logs
`Rounding up size to full physical extent <1.06 GiB`

**To Reproduce**
1. Create PVC
```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: simple-pvc
spec:
  volumeMode: Filesystem
  storageClassName: topolvm-provisioner
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 1136234735
```

2. Create pod
```
apiVersion: v1
kind: Pod
...
  volumes:
  - name: mypvc
    persistentVolumeClaim:
      claimName: simple-pvc
```

3. Create Snapshot
```
apiVersion: snapshot.storage.k8s.io/v1
kind: VolumeSnapshot
metadata:
  name: simple-snap
spec:
  volumeSnapshotClassName: topolvm-provisioner
  source:
    persistentVolumeClaimName: simple-pvc
```
**Expected behavior**
Snapshot succeed

**Additional context**
When creating PVC with size `1136234735`, LV gets created with size `1136656384`, probably because of LVM rounding

pluser:
@qJkee Sorry for the late reply.
I have successfully reproduced it in my local environment.
I can assume that this bug is caused by a rounding error. We are currently working on identifying the exact problem and how to fix it, but it is likely to take a long time to fix due to the wide range of affected areas.

As a work-around, there is a means to target PVCs with capacities (e.g., 4MB units) where rounding does not occur.

molpako:
@qJkee Thank you for reporting. This should be fixed in the above PR.

PR #1061, opened 2025-06-20
fix: return actual lv size from ControllerExpandVolume

molpako:
Follow-up to #1052: add missing ControllerExpandVolume changes

- related: https://github.com/topolvm/topolvm/issues/1039


internal/driver/controller.go
```diff
@@ -638,13 +638,14 @@ func (s controllerServerNoLocked) ControllerExpandVolume(ctx context.Context, re
 		currentSize = &lv.Spec.Size
 	}
 
+	_, nodeExpansionRequired := req.VolumeCapability.GetAccessType().(*csi.VolumeCapability_Mount)
```
molpako, line 641: I chose a variable name that emphasizes its intended use rather than a state-like name such as isMount. Please let me know if you have any feedback.

ushitora-anqou (approved):

llamerada-jp (approved):

PR #1075, opened 2025-07-17
test/e2e: add a script to run minikube e2e tests on multipass

ushitora-anqou:


pluser:
Because it includes version number, it may be a good idea to update the version number during maintenance.
https://github.com/topolvm/topolvm/blob/bf1fc85672a2525753496d3f9f387e05be99956a/docs/maintenance.md?plain=1#L24-L29

peng225 (dismissed):

PR #1081, opened 2025-08-06
Avoid unnecessary VG scan by adding Status.Path

sirpaleet:
Avoid unnecessary VG scan by adding Status.Path

A problem has been noticed to appear with a high number of PVC/mounts. NodePublishVolume calls take a long time to complete because of a scan for the volume.

In node.go NodePublishVolume uses getLvFromContext, which does an unnecessary volume group scan. This can be avoided by simply reading the path from lvr.Status.Path. If it has not gotten a value, the old way is used. Doing this meant adding a new Status value "Path" which gets its value already when it is available in CreateLV. To keep code cohesion, also NodeExpandVolume was edited to match.

Feel free to suggest how this could be made more efficient.

### DIFF
--- a/internal/controller/logicalvolume_controller.go
+++ b/internal/controller/logicalvolume_controller.go
@@ -257,6 +257,7 @@ func (r *LogicalVolumeReconciler) createLV(ctx context.Context, log logr.Logger,
 
 		lv.Status.VolumeID = volume.Name
 		lv.Status.CurrentSize = resource.NewQuantity(volume.SizeBytes, resource.BinarySI)
+		lv.Status.Path = volume.Path
 		lv.Status.Code = codes.OK
 		lv.Status.Message = ""
 		return nil
--- a/internal/driver/internal/k8s/logicalvolume_service.go
+++ b/internal/driver/internal/k8s/logicalvolume_service.go
@@ -343,6 +343,30 @@ func (s *LogicalVolumeService) updateSpecSize(ctx context.Context, volumeID stri
 		})
 }
 
+// UpdateStatusPath updates .Status.Path of LogicalVolume.
+func (s *LogicalVolumeService) UpdateStatusPath(ctx context.Context, lv *topolvmv1.LogicalVolume, path string) error {
+	logger.Info("Updating Status.Path", "path", path)
+	return wait.ExponentialBackoffWithContext(ctx,
+		retry.DefaultBackoff,
+		func(ctx context.Context) (bool, error) {
+			lv.Status.Path = path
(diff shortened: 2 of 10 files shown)


sirpaleet:
Since this PR would add a new flag/variable to the topolvm-node init, does someone know maybe what parts of documentation / in the helm chart need to be edited as well? The reason why this was added to begin with is because the VG's / DeviceClass's default value is not added to the topolvm-node but exists in the lvmd service. Adding an extra variable to define this for the node would make the mount process faster since it would not need to be scanned (which is a problem with high amounts of mounts). -- Any suggestions?

sirpaleet:
Okay, thanks for the input! I'll convert this back to Draft and open when I have another solution proposed.

internal/driver/internal/k8s/logicalvolume_service.go
```diff
@@ -343,6 +343,30 @@ func (s *LogicalVolumeService) updateSpecSize(ctx context.Context, volumeID stri
 		})
 }
 
+// UpdateStatusPath updates .Status.Path of LogicalVolume.
+func (s *LogicalVolumeService) UpdateStatusPath(ctx context.Context, lv *topolvmv1.LogicalVolume, path string) error {
```
sirpaleet, line 347: This is a fallback way to updating the Status.Path.

internal/controller/logicalvolume_controller.go
```diff
@@ -257,6 +257,7 @@ func (r *LogicalVolumeReconciler) createLV(ctx context.Context, log logr.Logger,
 
 		lv.Status.VolumeID = volume.Name
 		lv.Status.CurrentSize = resource.NewQuantity(volume.SizeBytes, resource.BinarySI)
+		lv.Status.Path = volume.Path
```
sirpaleet, line 260: Giving Status.Path value in (LogicalVolumeReconciler) createLV.

sirpaleet:
Opened again, please review :) + also added some comments and questions.

llamerada-jp:
I'm sorry for the misunderstanding. The cache mechanism using CR itself increases considerations and complexity. So I would like to avoid it.
I don't want to implement a cache function in TopoLVM itself, but if you need to resolve performance issues, we suggest implementing a cache in the lvm command using sh to speed things up. If you want to implement a custom lvm for this purpose, lvm-command-prefix may be useful.

llamerada-jp:
@sirpaleet 
This PR will be closed for now. When submitting code for a different mechanism, we would appreciate it if you could create a new PR.

Issue #1101, opened 2025-10-01
Thin pool uses hard threshold to stop volume provisioning


kiemlicz:
**What should the feature do:**

Allow the thin pool to create LVs "endlessly".

`func (p *thinPoolAdapter) Free(ctx context.Context) (uint64, error)` [internal/lvmd/pool.go:42](https://github.com/topolvm/topolvm/blob/main/internal/lvmd/pool.go#L42):
Uses [`Usage` method](https://github.com/topolvm/topolvm/blob/main/internal/lvmd/command/lvm.go#L417), and [`FreeBytes` from `internal/lvmd/command/lvm.go:291` ](https://github.com/topolvm/topolvm/blob/main/internal/lvmd/command/lvm.go#L291-300).

The `FreeBytes` is not taking into account the `Data%`, but instead a logic based on configurable `overprovisionRatio`.

I've checked the [design](https://github.com/topolvm/topolvm/blob/main/docs/proposals/thin-volumes.md#capacity-awareness) and seems to me that actual use case for `overprovisionRatio` parameter is updating the capacity for scheduling decisions?

I'm wondering if is it possible to make the `overprovisioningRatio` parameter optional?
As of capacity calculus: take the `Data%` into account and place the actual pool usage.    
Additionally as a user I can use affinity rules to spread my workload among nodes.  

Given the change is acceptable I can work on this PR.

**What is use case behind this feature:**

- I would expect the LVM-thinpool-backed storage to behave, well... like the LVM. The LVM doesn't introduce virtual overprovisioning parameter, it allows to create thin LVs "endlessly" (as long they don't exhaust the underlying data storage)
- as a TopoLVM user I have a ton of almost empty LVs and I can't schedule new ones (snapshots of rarely changing LV). The [`capacity awareness`](https://github.com/topolvm/topolvm/blob/main/docs/proposals/thin-volumes.md#capacity-awareness) works in my disadvantage here.

molpako:
Thank you for raising this.

Just to confirm, is the issue that, even though thin snapshots only consume the changed blocks, they still appear to consume their full virtual size (i.e., the provisioned size within the Thin Pool) for scheduling purposes?

Also, does a volume have low usage compared to its virtual size?

kiemlicz:
> Just to confirm, is the issue that, even though thin snapshots only consume the changed blocks, they still appear to consume their full virtual size (i.e., the provisioned size within the Thin Pool) for scheduling purposes?

[...]

PR #1113, opened 2025-10-30
Making overprovisioning parameter optional

kiemlicz:
[Context](https://github.com/topolvm/topolvm/issues/1101)

Decided to start with simple change: 
- make the overprovisioning parameter optional
- adjust the free space calculation to take into account lack of overprovisioning
- added test-case, aligned existing ones

Please let me know if this is the acceptable direction, I'll update the docs then

### DIFF
--- a/internal/lvmd/command/lvm.go
+++ b/internal/lvmd/command/lvm.go
@@ -288,15 +288,24 @@ type ThinPoolUsage struct {
 	SizeBytes       uint64
 }
 
-func (t ThinPoolUsage) FreeBytes(overprovisionRatio float64) (uint64, error) {
-	if overprovisionRatio < 1.0 {
-		return 0, errors.New("overprovision ratio must be >= 1.0")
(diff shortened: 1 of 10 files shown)


molpako (changes requested):
Thanks for creating the PR.

OverProvisionRatio must be just a value. While making it optional could be efficient, we need to prioritize backward compatibility. 

Instead, please consider a mode-switch parameter. It may be redundant, but we can accept it.

molpako (changes requested):
Thank you for addressing this. Please take a look at my comments.

kiemlicz:
(don't close)

peng225 (commented):
Just to be clear, I'll review this PR after you've addressed all of @molpako's comments.

kiemlicz:
not stale

kiemlicz:
not stale

kiemlicz:
not stale, almost there

kiemlicz:
not

PR #1118, opened 2025-11-07
charts: add ignoreNamespaces for webhooks

### LINKED ISSUE #1115: Remove the dependency on `kube-system` label `topolvm.io/webhook=ignore`

Remove the need for labeling `kube-system` with `topolvm.io/webhook=ignore`. Always treat `kube-system` as having that label.

I haven't found a way to configure labels on system/default k8s namespaces. The only way to do it, therefore, is to patch the Namespace manually. I would like a declarative solution.

ushitora-anqou:
Fix #1115.

### DIFF
--- a/charts/topolvm/templates/mutatingwebhooks.yaml
+++ b/charts/topolvm/templates/mutatingwebhooks.yaml
@@ -21,6 +21,12 @@ webhooks:
       - key: {{ include "topolvm.pluginName" . }}/webhook
         operator: NotIn
         values: ["ignore"]
+      {{- with .Values.webhook.podMutatingWebhook.ignoreNamespaces }}
+      - key: kubernetes.io/metadata.name
+        operator: NotIn
+        values:
+          {{- toYaml . | nindent 10 }}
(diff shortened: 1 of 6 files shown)


charts/topolvm/templates/mutatingwebhooks.yaml
```diff
@@ -21,6 +21,14 @@ webhooks:
       - key: {{ include "topolvm.pluginName" . }}/webhook
         operator: NotIn
         values: ["ignore"]
+      {{- with .Values.webhook.podMutatingWebhook.ignoreNamespaces }}
+      - key: kubernetes.io/metadata.name
+        operator: NotIn
+        values:
+          {{- range $v := .}}
```
molpako: nit: It might render correctly even without the space, but it's better to add it for consistency with the others.
```suggestion
          {{- range $v := . }}
```

skoya76 (approved):

molpako (approved):

skoya76 (approved):

Issue #1126, opened 2025-12-08
Make filesystem successfully even if a fake signature exists


toshipp:
## What

When creating a new PV, TopoLVM creates an LVM lv and formats it with the appropriate filesystem specified by the StorageClass. We found that it may fail because the lv would contain a jmicron_raid_member that is not a real one. This is a similar problem to the one resolved by #1033.

## How

If the filesystem is not formatted, we can always format it, even if there is a fake signature. To avoid verification failure for a fake signature by mkfs, erase them all by wipefs or zero-filling the signature before mkfs [here](https://github.com/topolvm/topolvm/blob/7616df5f4d333342215ff2887a38c738f41fa68f/internal/driver/node.go#L233-L237).

## Checklist

- [x] Finish implementation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

PR #1127, opened 2025-12-10
run wipefs before mkfs to clear stray signatures

skoya76:
NodePublishVolume now calls wipefs -a when no filesystem is detected, before SafeFormatAndMount runs mkfs. This clears stray signatures (e.g., jmicron_raid_member) that would otherwise make mkfs refuse to format and cause mount to fail (refs #1126).

### DIFF
--- a/internal/driver/node.go
+++ b/internal/driver/node.go
@@ -7,7 +7,6 @@ import (
 	"io"
 	"os"
 	"path/filepath"
-	"strconv"
 	"strings"
 	"sync"
 
@@ -233,13 +232,9 @@ func (s *nodeServerNoLocked) nodePublishFilesystemVolume(req *csi.NodePublishVol
 	if !mounted {
 		mountFunc := s.mounter.Mount
 		if len(fsType) == 0 {
-			// NOTE: temporary instrumentation to reproduce the mount failure reported in
-			// https://github.com/topolvm/topolvm/issues/1126. We inject a fake jmicron-like
-			// signature into the last sector when the device is treated as unformatted, then
-			// wipe signatures before mkfs. Remove this after validation.
-			if err := s.injectFakeJmicronSignature(lv.GetPath(), req.GetVolumeId()); err != nil {
-				return err
-			}
+			// wipefs to clear stray signatures (e.g., JMicron marker in the
+			// last sector per #1126). Behavior was verified in PR 1127. If you
+			// need the verification code, see that PR.
 			if out, err := s.mounter.Exec.Command("wipefs", "-a", lv.GetPath()).CombinedOutput(); err != nil {
 				return status.Errorf(codes.Internal, "wipefs failed: volume=%s, output=%s, error=%v", req.GetVolumeId(), string(out), err)
 			}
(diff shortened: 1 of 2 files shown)


skoya76:
No new test case was added for this change because reproducing issue #1126 in the CI environment is difficult. Instead, the correctness of the fix was verified through the following steps:

- First, commit `ec9987e7269a3690194be06c9cc35232c44650f7` modified the code to write "JM" to the last sector immediately before calling `mkfs`. This successfully reproduced the error reported in #1126 . The CI log confirms that the presence of "JM" causes the expected `FailedMount` error ([ref](https://github.com/topolvm/topolvm/actions/runs/20117587463/job/57730611466#step:9:1894)).
- Next, while retaining the changes from the first step, commit `43d193620c8e0da4a9406e229cdbfcfb91e892bf` added the execution of `wipefs -a` immediately before `mkfs`. The CI log confirms that even with "JM" written to the last sector, it is properly cleaned up, and the CI passes. This demonstrates that `wipefs -a` is effective and the implementation is correct.
- Finally, commit `f5ee81c714e6f6c88fc909a67b2b87e2e866afb9` removed the temporary verification code to finalize the changes.

These steps demonstrate that this change successfully prevents `mkfs` from failing when a "JM" signature exists in the last sector.

cupnes (approved):

peng225 (commented):
https://github.com/topolvm/topolvm/pull/1127#issuecomment-3645543273
> The CI log confirms that the presence of "JM" causes the expected FailedMount error ([ref](https://github.com/topolvm/topolvm/actions/runs/20117587463/job/57730611466#step:9:1894)).

When I open the link above, I can see the following log:
```
    Warning  FailedMount  7m8s (x6 over 7m23s)  kubelet            MountVolume.SetUp failed for volume "certs" : secret "topolvm-mutatingwebhook" not found
```

I wonder if this is really the log of the problem caused by "JM" marker because it says "secret not found". Could you explain a bit more about your intention for linking to this log?

skoya76:
> [#1127 (comment)](https://github.com/topolvm/topolvm/pull/1127#issuecomment-3645543273)
> 
> > The CI log confirms that the presence of "JM" causes the expected FailedMount error ([ref](https://github.com/topolvm/topolvm/actions/runs/20117587463/job/57730611466#step:9:1894)).
> 
> When I open the link above, I can see the following log:
> 
> ```
>     Warning  FailedMount  7m8s (x6 over 7m23s)  kubelet            MountVolume.SetUp failed for volume "certs" : secret "topolvm-mutatingwebhook" not found
> ```
> 
> I wonder if this is really the log of the problem caused by "JM" marker because it says "secret not found". Could you explain a bit more about your intention for linking to this log?


The link I provided was misleading. I accidentally pointed to a “secret not found” error, which isn’t directly related to this issue. I intended to show logs tied to a PVC FailedMount, but the CI runs don’t expose mkfs/blkid/wipefs failures directly.

That said, here’s how I’m interpreting the results: when I injected JM only, every test case ended up with the PVC pod never reaching Running (the scheduling test timed out after 300 seconds), which strongly suggests mkfs was failing behind the scenes. After adding wipefs -a to clear JM, the same tests succeeded. In other words, “with JM it consistently stalls, with JM + wipefs it passes,” and that behavior change is the basis for the fix.

peng225 (approved):

cupnes (approved):

PR #1138, opened 2026-01-13
feat: release helm charts to ghcr.io

xvzf:
Signed-off-by: Matthias Riegler <me@m4tbit.de>

### DIFF
--- a/.github/workflows/helm-release.yaml
+++ b/.github/workflows/helm-release.yaml
@@ -16,6 +16,13 @@ jobs:
           git config user.name "$GITHUB_ACTOR"
           git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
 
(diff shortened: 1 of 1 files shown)


xvzf:
Hi @llamerada-jp - simply following the best practices. Quote from [helm docs](https://helm.sh/de/docs/topics/registries/): _It is recommended to use container registries with [OCI](https://www.opencontainers.org/) support to store and share chart packages._

I noticed we're using Helm OCI charts for all of our GitOps driven deployments, just topolvm has been missing. If the maintainer team does not expect a benefit from this, it's fine as well!

llamerada-jp:
@xvzf 
We considered our own motivations and requirements for using OCI, but we couldn't identify any clear reasons. For now, we would like to hold off on using OCI.

PR #1146, opened 2026-01-26
Fix volume expansion timeout for LVM extent-aligned sizes

qJkee:
Volume expansion was timing out when the requested size didn't align to LVM extent boundaries (typically 4MB). The expansion succeeded at the LVM level, but the CSI driver waited indefinitely because it expected an exact size match.

LVM rounds up volume sizes to extent boundaries. For example, a request for 203Mi becomes 204Mi (51 extents × 4MB). The previous comparison
 would never return true when LVM rounded up.

Changed the comparison from  (exact match) to  (at least as large as requested) to correctly handle LVM extent rounding.

Fixes topolvm#1144

### DIFF
--- a/internal/driver/internal/k8s/logicalvolume_service.go
+++ b/internal/driver/internal/k8s/logicalvolume_service.go
@@ -295,8 +295,8 @@ func (s *LogicalVolumeService) ExpandVolume(ctx context.Context, volumeID string
 			return false, nil
 		}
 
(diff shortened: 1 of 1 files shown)


qJkee:
@molpako 
Done, sorry for delay

qJkee:
@jakobmoellerdev maybe you can give 2nd approval?

PR #1149, opened 2026-02-09
Wait for controller evacuating from deleted nodes to stabilize CSI sa…

toshipp:
…nity test

We delete nodes to make the test use the remaining node. In this time, the controller would be on deleted nodes, it will be rescheduled to the node. But we did not wait for it before, the test cannot connect to the controller, so it causes flakiness.


toshipp:
We have already extended the default timeout to 5 minutes.
https://github.com/topolvm/topolvm/blob/main/test/e2e/suite_test.go#L33

PR #1165, opened 2026-04-08
Expose extraArgs for the csi-provisioner in the Helm chart

hamza-boudouche:
Expose extraArgs for the csi-provisioner in the Helm chart

To be able to pass custome cli args that are not exposed via the values or set in the chart currently.

### DIFF
--- a/charts/topolvm/templates/controller/deployment.yaml
+++ b/charts/topolvm/templates/controller/deployment.yaml
@@ -166,6 +166,10 @@ spec:
             - --enable-capacity
             - --capacity-ownerref-level=2
             {{- end }}
+          {{- $csiProvisionerArgs := default (list) .Values.controller.csiProvisioner.args }}
+          {{- if gt (len $csiProvisionerArgs) 0 }}
+          args: {{ toYaml $csiProvisionerArgs | nindent 12 }}
+          {{- end }}
           ports:
             - containerPort: 9809
               name: csi-provisioner
--- a/charts/topolvm/templates/controller/deployment.yaml
+++ b/charts/topolvm/templates/controller/deployment.yaml
@@ -157,6 +157,21 @@ spec:
             - /csi-provisioner
             - --csi-address=/run/topolvm/csi-topolvm.sock
             - --feature-gates=Topology=true
+            {{- if gt (int .Values.controller.csiProvisioner.workerThreads) 0 }}
+            - --worker-threads={{ .Values.controller.csiProvisioner.workerThreads }}
+            {{- end }}
+            {{- if gt (int .Values.controller.csiProvisioner.kubeAPIBurst) 0 }}
+            - --kube-api-burst={{ .Values.controller.csiProvisioner.kubeAPIBurst }}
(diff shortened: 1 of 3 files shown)


charts/topolvm/templates/controller/deployment.yaml
```diff
@@ -157,6 +157,21 @@ spec:
...
+            {{- if gt (int .Values.controller.csiProvisioner.kubeAPIBurst) 0 }}
+            - --kube-api-burst={{ .Values.controller.csiProvisioner.kubeAPIBurst }}
+            {{- end }}
+            {{- if gt (int .Values.controller.csiProvisioner.kubeAPICapacityBurst) 0 }}
+            - --kube-api-capacity-burst={{ .Values.controller.csiProvisioner.kubeAPICapacityBurst }}
+            {{- end }}
+            {{- if gt (float64 .Values.controller.csiProvisioner.kubeAPICapacityQPS) 0.0 }}
+            - --kube-api-capacity-qps={{ .Values.controller.csiProvisioner.kubeAPICapacityQPS }}
+            {{- end }}
+            {{- if gt (int .Values.controller.csiProvisioner.kubeAPIQPS) 0 }}
```
molpako, line 172: Would you mind using float for `KubeAPIQPS` here?

It is handled as a float32 in csi-provisioner.

https://github.com/kubernetes-csi/csi-lib-utils/blob/f7e0e2eddd101b04ea573db657a93b0429928fed/standardflags/flags.go#L59

molpako (changes requested):

toshipp (commented):
I have a concern about adding many options, as it would be hard to maintain. similar to https://github.com/topolvm/topolvm/pull/1166#pullrequestreview-4080219776

hamza-boudouche:
> I have a concern about adding many options, as it would be hard to maintain. similar to [#1166 (review)](https://github.com/topolvm/topolvm/pull/1166#pullrequestreview-4080219776)

Same [reply](https://github.com/topolvm/topolvm/pull/1166#issuecomment-4212777297) I made on the other comment is valid here.

charts/topolvm/templates/controller/deployment.yaml
hamza-boudouche: I can do that, but let's wait for the [comment](https://github.com/topolvm/topolvm/pull/1165#pullrequestreview-4080238034) from @toshipp to be sorted out, because if we go with that proposed route, then the individual args won't be explicitly exposed in the helm chart.

hamza-boudouche:
> I have a concern about adding many options, as it would be hard to maintain. similar to [#1166 (review)](https://github.com/topolvm/topolvm/pull/1166#pullrequestreview-4080219776)

done

charts/topolvm/templates/controller/deployment.yaml
hamza-boudouche: The individual values are not exposed anymore, but an option to specify extraArgs is exposed instead (as proposed by @toshipp), which means we don't have to worry about types.

PR #1166, opened 2026-04-08
Add option to specify extraArgs for the csi-resizer in the Helm chart

hamza-boudouche:
Add option to specify extraArgs for the csi-resizer in the Helm chart

to be able to customize args that are not set or exposed by the Helm chart

### DIFF
--- a/charts/topolvm/templates/controller/deployment.yaml
+++ b/charts/topolvm/templates/controller/deployment.yaml
@@ -212,6 +212,10 @@ spec:
             - --leader-election-namespace={{ .Release.Namespace }}
             {{- end }}
             - --http-endpoint=:9810
+          {{- $csiResizerArgs := default (list) .Values.controller.csiResizer.args }}
+          {{- if gt (len $csiResizerArgs) 0 }}
+          args: {{ toYaml $csiResizerArgs | nindent 12 }}
+          {{- end }}
           ports:
             - containerPort: 9810
               name: csi-resizer
--- a/charts/topolvm/templates/controller/deployment.yaml
+++ b/charts/topolvm/templates/controller/deployment.yaml
@@ -207,6 +207,7 @@ spec:
           command:
             - /csi-resizer
             - --csi-address=/run/topolvm/csi-topolvm.sock
+            - --handle-volume-inuse-error={{ .Values.controller.csiResizer.handleVolumeInUseError }}
             {{- if .Values.controller.leaderElection.enabled }}
             - --leader-election
             - --leader-election-namespace={{ .Release.Namespace }}
--- a/charts/topolvm/values.yaml
+++ b/charts/topolvm/values.yaml
@@ -432,6 +432,10 @@ controller:
   # controller.args -- Arguments to be passed to the command.
   args: []
 
+  csiResizer:
+    # controller.csiResizer.args -- Arguments to be passed to the csi-resizer.
+    args: []
+
   storageCapacityTracking:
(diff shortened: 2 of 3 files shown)


toshipp (commented):
I understand your motivation, but I don't want to introduce helm options for each command option.
Is it sufficient to add the args option like [this](https://github.com/topolvm/topolvm/blob/main/charts/topolvm/templates/controller/deployment.yaml#L86)?

hamza-boudouche:
> I understand your motivation, but I don't want to introduce helm options for each command option. Is it sufficient to add the args option like [this](https://github.com/topolvm/topolvm/blob/main/charts/topolvm/templates/controller/deployment.yaml#L86)?

I can do that too, but:
-  in the case of the csi-resizer, we're already passing some args to it, so I assume we'd need a  `controller.csiResizer.extraArgs` Helm value instead (maybe that's what you meant already), and we'll need to filter out the args [we already pass to it](https://github.com/topolvm/topolvm/blob/main/charts/topolvm/templates/controller/deployment.yaml#L209-L214) so that they're not duplicated (so there is some small amount of complexity involved still :) )
- if we do it like this we won't be able to check if the extraArgs passed by a user are valid, so if they pass a non existing or false arg and it breaks they will be on their own.

I have no strong preference for either approaches, if you feel like this way is better taking in consideration the above points then I can change it.

toshipp:
Other commands already use both the command field for some options and the args field. So I prefer adding the args option for consistency and simplicity.
As you said, users may specify wrong option combinations, but Helm has no strong type system and validation capabilities, it is up to users.

hamza-boudouche:
> Other commands already use both the command field for some options and the args field. So I prefer adding the args option for consistency and simplicity. As you said, users may specify wrong option combinations, but Helm has no strong type system and validation capabilities, it is up to users.

should be good now

charts/topolvm/values.yaml
```diff
@@ -432,6 +432,10 @@ controller:
   # controller.args -- Arguments to be passed to the command.
   args: []
 
+  csiResizer:
+    # controller.csiResizer.extraArgs -- Additional arguments appended to the built-in csi-resizer arguments. Users are responsible for ensuring they are valid csi-resizer args.
+    extraArgs: []
```
toshipp: Other containers use the name `args`. For consistency, could you rename it? Also, simplify the comment like 
"Arguments to be passed to the csi-resizer command."

toshipp (changes requested):

charts/topolvm/values.yaml
hamza-boudouche: done

charts/topolvm/templates/controller/deployment.yaml
```diff
@@ -212,6 +212,10 @@ spec:
             - --leader-election-namespace={{ .Release.Namespace }}
             {{- end }}
             - --http-endpoint=:9810
+          {{- $csiResizerArgs := default (list) .Values.controller.csiResizer.args }}
+          {{- if gt (len $csiResizerArgs) 0 }}
+          args: {{ toYaml $csiResizerArgs | nindent 12 }}
+          {{- end }}
```
daichimukai, line 218:
```suggestion
          {{- with .Values.controller.csiResizer.args }}
          args: {{ toYaml . | nindent 12 }}
          {{- end }}
```

daichimukai (changes requested):
Other configuration items related to csi-resizer already exist under `resources.csi_resizer`, `env.csi_resizer`, and so on. It feels inconsistent to place only args under controller. I think it would be more consistent to place it under `args.csi_resizer`.

PR #1171, opened 2026-04-20
lvmd/command: drain stdout before returning from parseFullReportResult

### LINKED ISSUE #1161: Potential deadlock when JSON decoding fails in lvmd state command
**Describe the bug**

In `internal/lvmd/command/lvm_state_json.go`, when `parseFullReportResult` fails to decode JSON output, the function returns early while leaving unread data in `stdout`. The `defer streamed.Close()` statement then invokes `cmd.Wait()` without fully consuming the stdout pipe, which violates the `os/exec.StdoutPipe` contract.

The Go runtime documentation explicitly states:
> "It is incorrect to call Wait before all reads from the pipe have completed."

This can cause the process to deadlock/hang when:
1. LVM command produces output larger than the pipe buffer (~64KB)
2. JSON decoder encounters an error partway through parsing
3. Unread stdout fills the pipe buffer, blocking the child process from exiting
4. The parent blocks in `cmd.Wait()` (or earlier in `io.ReadAll(stderr)`)

**Environments**
- Version: topolvm-chart-v16.0.0, topolvm-v0.40.1
- OS: Flatcar

**To Reproduce**

Steps to trigger:
1. Inject a JSON parse error at arbitrary position in `parseFullReportResult`
2. With sufficient output to partially fill the pipe buffer
3. Observe: `getLVMState` may hang indefinitely instead of returning error

SAY-5:
Fixes #1161.

`getLVMState` streams the `lvm fullreport` output through an `os/exec.StdoutPipe`-backed `ReadCloser` and defers `streamed.Close()`, which internally calls `cmd.Wait()`. The Go documentation explicitly forbids calling `Wait` before all reads from the pipe complete, because an unread tail keeps the child process blocked on a full pipe buffer.

`parseFullReportResult` does not honour that contract. `json.Decoder` reads only up to the end of the top-level value and always leaves at least the trailing newline in the pipe; on the error path it typically leaves a much larger unread tail. For fullreports above ~64 KB (multi-VG or many-LV clusters) this deadlocks lvmd — the child cannot flush stdout, `cmd.Wait()` never returns, and the operator's lvmd loop hangs indefinitely.

This captures the decode error, drains the rest of the stream with `io.Copy(io.Discard, data)`, then returns. Decode errors still propagate; the drain just guarantees the pipe is fully consumed so the subsequent `Close` → `Wait` path completes cleanly.

### DIFF
--- a/internal/lvmd/command/lvm_command.go
+++ b/internal/lvmd/command/lvm_command.go
@@ -113,15 +113,9 @@ type commandReadCloser struct {
 
 // Close drains any unread stdout, reads stderr, and waits for the command to exit.
 func (p commandReadCloser) Close() error {
-	// Drain stdout before Wait() to prevent blocking when the caller
-	// stopped reading early and the child has more output than the
-	// pipe buffer (#1161).
-	if _, err := io.Copy(io.Discard, p.ReadCloser); err != nil {
-		// Caller may have already closed the reader; only that case is expected.
-		if !errors.Is(err, os.ErrClosed) {
-			return err
-		}
-	}
+	// Drain stdout before Wait() to prevent blocking when the caller stopped
+	// reading early and the child has more output than the pipe buffer.
+	_, _ = io.Copy(io.Discard, p.ReadCloser)
 
 	// Read the stderr output after the read has finished since we are sure by then the command must have run.
 	stderr, err := io.ReadAll(p.stderr)
--- a/internal/lvmd/command/lvm_command.go
+++ b/internal/lvmd/command/lvm_command.go
@@ -111,9 +111,18 @@ type commandReadCloser struct {
 	stderr io.ReadCloser
 }
 
-// Close closes stdout and stderr and waits for the command to exit. Close
-// should not be called before all reads from stdout have completed.
+// Close drains any unread stdout, reads stderr, and waits for the command to exit.
 func (p commandReadCloser) Close() error {
+	// Drain stdout before Wait() to prevent blocking when the caller
+	// stopped reading early and the child has more output than the
+	// pipe buffer (#1161).
+	if _, err := io.Copy(io.Discard, p.ReadCloser); err != nil {
+		// Caller may have already closed the reader; only that case is expected.
+		if !errors.Is(err, os.ErrClosed) {
+			return err
+		}
+	}
+
 	// Read the stderr output after the read has finished since we are sure by then the command must have run.
 	stderr, err := io.ReadAll(p.stderr)
 	if err != nil {
--- a/internal/lvmd/command/lvm_command_test.go
+++ b/internal/lvmd/command/lvm_command_test.go
@@ -157,9 +157,10 @@ func TestCallLVM(t *testing.T) {
 
 func TestCommandReadCloserClose(t *testing.T) {
 	t.Run("Close returns even when the caller stops reading early", func(t *testing.T) {
-		// Produce ~256 KiB of output, safely larger than the pipe buffer
-		// so cmd.Wait would deadlock on an undrained stdout (#1161).
-		ctx := log.IntoContext(context.Background(), testr.New(t))
+		// Regression for #1161.
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+		ctx = log.IntoContext(ctx, testr.New(t))
 		cmd := exec.CommandContext(ctx, "dd", "if=/dev/zero", "bs=1024", "count=256", "status=none")
 
 		rc, err := runCommand(ctx, verbosityLVMStateNoUpdate, cmd)
@@ -167,23 +168,13 @@ func TestCommandReadCloserClose(t *testing.T) {
 			t.Fatalf("runCommand: %v", err)
 		}
 
-		// Read only a handful of bytes, simulating a parser that bails
-		// out on the first value it decodes.
 		buf := make([]byte, 4)
 		if _, err := io.ReadFull(rc, buf); err != nil {
 			t.Fatalf("read prefix: %v", err)
 		}
 
-		done := make(chan error, 1)
-		go func() { done <- rc.Close() }()
-		select {
-		case err := <-done:
-			if err != nil {
-				t.Fatalf("close returned error: %v", err)
-			}
-		case <-time.After(5 * time.Second):
-			_ = cmd.Process.Kill()
-			t.Fatal("commandReadCloser.Close deadlocked with unread stdout")
+		if err := rc.Close(); err != nil {
+			t.Fatalf("close returned error: %v", err)
 		}
 	})
 }
--- a/internal/lvmd/command/lvm_command_test.go
+++ b/internal/lvmd/command/lvm_command_test.go
@@ -4,9 +4,11 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"os/exec"
 	"regexp"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/go-logr/logr/funcr"
 	"github.com/go-logr/logr/testr"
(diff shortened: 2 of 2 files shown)


llamerada-jp (changes requested):
Thank you for your help.

To resolve this problem, it seems good to update  `commandReadCloser.Close()` instead of `parseFullReportResult()`. Because it is useful in other cases. Users of `commandReadCloser` don't need to be aware of the limitation.

SAY-5:
Thanks @llamerada-jp — moved the drain into `commandReadCloser.Close` so every caller of `callLVMStreamed` benefits automatically, and reverted the `parseFullReportResult` change so the logic now lives in one place.

Verified locally on macOS, go 1.26.2:

- `gofmt -s -l`: clean
- `go vet ./internal/lvmd/command/...`: clean
- `SKIP_TESTS_USING_ROOT=1 go test -race -count=1 ./internal/lvmd/command/...`: pass

Added a regression test `TestCommandReadCloserCloseDrainsUnreadStdout` that spawns a real `sh` child which writes 256 KiB to stdout, reads only four bytes from the `commandReadCloser`, then calls `Close()` under a 5-second watchdog. On the pre-fix Close the test trips the watchdog with `commandReadCloser.Close deadlocked with unread stdout (issue #1161)`; with the drain in place Close returns cleanly within milliseconds.

Net diff: +73 / -2 across lvm_command.go, lvm_command_test.go, lvm_state_json.go.

internal/lvmd/command/lvm_command.go
```diff
@@ -111,9 +111,26 @@ type commandReadCloser struct {
 	stderr io.ReadCloser
 }
 
-// Close closes stdout and stderr and waits for the command to exit. Close
-// should not be called before all reads from stdout have completed.
+// Close drains any unread stdout, reads stderr, and waits for the command
+// to exit. Callers no longer need to drain stdout themselves before calling
+// Close; any leftover output is discarded so cmd.Wait does not deadlock on
+// a child that produced more than the pipe buffer (#1161).
```
llamerada-jp:
```suggestion
// Close drains any unread stdout, reads stderr, and waits for the command to exit.
```

internal/lvmd/command/lvm_command.go
```diff
@@ -111,9 +111,26 @@ type commandReadCloser struct {
...
+	// whatever is left so the underlying process can finish.
+	if _, err := io.Copy(io.Discard, p.ReadCloser); err != nil {
+		// Drop the error if the underlying reader was already closed
+		// by the caller (which is valid usage) - we only care about
+		// getting cmd.Wait unblocked.
+		if !errors.Is(err, os.ErrClosed) {
+			return err
+		}
+	}
+
```
llamerada-jp:
```suggestion
	// Drain stdout before Wait() to prevent blocking when the caller stopped
	// reading early and the child has more output than the pipe buffer.
	_, _ = io.Copy(io.Discard, p.ReadCloser)
```

internal/lvmd/command/lvm_command_test.go
```diff
@@ -4,16 +4,70 @@ import (
...
 	"github.com/topolvm/topolvm/internal/lvmd/testutils"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+// TestCommandReadCloserCloseDrainsUnreadStdout is a regression test for
+// issue #1161 / PR #1171. When a command writes more stdout than the OS
+// pipe buffer (typically 64 KiB) and the caller stops reading early, Close
+// must drain the rest of the pipe; otherwise cmd.Wait blocks forever on
+// the child, the ReadCloser's Close blocks with it, and lvmd deadlocks.
+func TestCommandReadCloserCloseDrainsUnreadStdout(t *testing.T) {
```
llamerada-jp: Please align the structure of the test with the others. Also, could you move the test to the end, so it’s placed after the function?
```suggestion
func TestCommandReadCloserClose(t *testing.T) {
  t.Run("...", ...
```

internal/lvmd/command/lvm_command_test.go
```diff
@@ -4,16 +4,70 @@ import (
...
+	// Produce ~256 KiB of output - safely larger than any pipe buffer
+	// we expect to see in production. The child exits immediately after
+	// the stdout write returns, so if Close drains stdout it sees EOF
+	// and cmd.Wait unblocks. If Close does not drain, the child is
+	// still blocked in write(2) and we deadlock below.
+	const payloadBytes = 256 * 1024
+	cmd := exec.Command("sh", "-c", fmt.Sprintf("yes y | head -c %d", payloadBytes))
```
llamerada-jp:
```suggestion
	cmd := exec.Command("dd", "if=/dev/zero", "bs=1024", "count=256", "status=none")
```

internal/lvmd/command/lvm_command_test.go
```diff
@@ -4,16 +4,70 @@ import (
...
+	}
+	stderr, err := cmd.StderrPipe()
+	if err != nil {
+		t.Fatalf("stderr pipe: %v", err)
+	}
+	if err := cmd.Start(); err != nil {
+		t.Fatalf("start: %v", err)
+	}
+
+	rc := commandReadCloser{cmd: cmd, ReadCloser: stdout, stderr: stderr}
```
llamerada-jp: You can write it more simply using the `runCommand` function.

internal/lvmd/command/lvm_command_test.go
```diff
@@ -4,16 +4,70 @@ import (
...
+
+	select {
+	case err := <-done:
+		if err != nil {
+			t.Fatalf("close returned error: %v", err)
+		}
+	case <-time.After(5 * time.Second):
+		_ = cmd.Process.Kill()
+		t.Fatal("commandReadCloser.Close deadlocked with unread stdout (issue #1161)")
+	}
```
llamerada-jp: Isn't this block unnecessary?

internal/lvmd/command/lvm_command_test.go
```diff
@@ -4,16 +4,70 @@ import (
...
 	"testing"
+	"time"
 
 	"github.com/go-logr/logr/funcr"
 	"github.com/go-logr/logr/testr"
 	"github.com/topolvm/topolvm/internal/lvmd/testutils"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+// TestCommandReadCloserCloseDrainsUnreadStdout is a regression test for
```
llamerada-jp: Since there is a link to the issue, please keep your comments brief.

llamerada-jp (changes requested):

SAY-5:
Thanks for the thorough review @llamerada-jp. Force-pushed with all your inline suggestions applied:

- `commandReadCloser.Close()` now owns the drain (was already on that path, but collapsed the comment and removed the `errors.Is(err, os.ErrClosed)` branch in favour of the one-line `_, _ = io.Copy(io.Discard, p.ReadCloser)` you suggested)
- Removed the `parseFullReportResult` change; no consumer needs to know about the pipe-drain constraint anymore.
- Renamed the test to `TestCommandReadCloserClose` with a `t.Run` subtest, moved it after the subject function, switched the producer to `dd if=/dev/zero bs=1024 count=256 status=none`, dropped the unnecessary block, shortened the comment to just reference \#1161.

`go test ./internal/lvmd/command/ -run TestCommandReadCloserClose` passes locally; the existing `TestCallLVMStreamed`/`TestCallLVM` suites are untouched and continue to pass.

SAY-5:
Thanks for the detailed review, @llamerada-jp. Addressed all four inline suggestions in a force-push:

1. Shortened `Close` doc comment to the one-liner you proposed.
2. Replaced the long drain comment with your shorter version that only keeps the issue link.
3. Renamed the test to `TestCommandReadCloserClose`, moved it to the end of the file, wrapped the assertion in `t.Run`, and rewrote it to use `runCommand` instead of the manual `exec.Start` + pipe setup (no more duplicate plumbing).
4. Switched the payload command to `dd if=/dev/zero bs=1024 count=256 status=none` per your suggestion.

Also dropped the redundant recovery block and pared the remaining comments down to the issue link.

internal/lvmd/command/lvm_command_test.go
```diff
@@ -152,3 +154,36 @@ func TestCallLVM(t *testing.T) {
...
+		go func() { done <- rc.Close() }()
+		select {
+		case err := <-done:
+			if err != nil {
+				t.Fatalf("close returned error: %v", err)
+			}
+		case <-time.After(5 * time.Second):
+			_ = cmd.Process.Kill()
+			t.Fatal("commandReadCloser.Close deadlocked with unread stdout")
+		}
```
llamerada-jp: You can use `context.WithTimeout` instead of timeout using select to simplify the test.

```suggestion
		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
		defer cancel()
		ctx = log.IntoContext(ctx, testr.New(t))
		cmd := exec.CommandContext(ctx, "dd", "if=/dev/zero", "bs=1024", "count=256", "status=none")

		rc, err := runCommand(ctx, verbosityLVMStateNoUpdate, cmd)
		if err != nil {
			t.Fatalf("runCommand: %v", err)
		}

		// Read only a handful of bytes, simulating a parser that bails
		// out on the first value it decodes.
		buf := make([]byte, 4)
		if _, err := io.ReadFull(rc, buf); err != nil {
			t.Fatalf("read prefix: %v", err)
		}

		if err := rc.Close(); err != nil {
			t.Fatalf("close returned error: %v", err)
		}
```

llamerada-jp (changes requested):

llamerada-jp (approved):

toshipp (approved):

PR #1172, opened 2026-04-20
lvmd/command: use exec.CommandContext so cancelled RPCs kill their LVM child

### LINKED ISSUE #1163: lvmd: LVM child processes not cleaned up when kubelet RPC timeout occurs
**Describe the bug**

When kubelet calls NodePublishVolume with a timeout (typically 2 minutes), and lvmd's gRPC handler takes longer than expected, the gRPC connection is closed by the client with a context canceled error. However, lvmd's child LVM processes (vgs, lvs, etc.) continue executing indefinitely because `internal/lvmd/command/lvm_command.go` uses `exec.Command()` instead of `exec.CommandContext()`, which ignores context cancellation.

**Symptom:**
- CSI driver receives: `failed to list LV: rpc error: code = Canceled desc = context canceled`
- lvmd side: child LVM processes remain running
- System impact: Resource leaks as processes accumulate

**Environments**
- Version: All versions using `callLVMStreamed` pattern
- OS: Flatcar

**To Reproduce**

Steps to reproduce:
1. Kubernetes node is running with TopoLVM CSI plugin
2. Kubelet calls NodePublishVolume with a timeout (default: ~2 minutes, configurable via `csiTimeout`)
3. lvmd's GetLVList RPC handler takes longer than the configured `csiTimeout` to respond
   - Can occur during high LVM activity or system load
4. After `csiTimeout` is exceeded, kubelet's context times out and closes the gRPC connection
5. CSI driver receives: `failed to list LV: rpc error: code = Canceled desc = context canceled`
6. Observe: lvmd side still has `vgs` or `lvs` child processes running, consuming resources

**Expected behavior**

- `exec.CommandContext()` should be used instead of `exec.Command()` to respect context cancellation
- When kubelet's RPC timeout is exceeded and context is canceled:
  - lvmd's RPC handler context is canceled
  - Child LVM processes (vgs, lvs, etc.) receive SIGTERM from the kernel
  - Processes are cleaned up immediately, not left running
- No child LVM processes should remain after gRPC connection is closed
- System should not accumulate orphaned processes even under repeated timeout scenarios

**Root Cause**

`internal/lvmd/command/lvm_command.go:76`

```go
cmd := exec.Command(wholeCommand[0], wholeCommand[1:]...)  // ← context is ignored!
```

The `ctx` parameter is passed through the call stack (`callLVMStreamed()` → `runCommand()`) but **never used to create the command**. This means:
- When kubelet's context times out, the RPC handler context is canceled
- But `exec.Command()` ignores context
- Child processes are never killed
- They continue running even after gRPC connection is closed

**Solution:**

Replace with:
```go
cmd := exec.CommandContext(ctx, wholeCommand[0], wholeCommand[1:]...)  // ← respects context
```

This ensures that when context is canceled, the child process receives SIGTERM and is automatically cleaned up.

**Additional context**

**Execution Flow:**

```
kubelet (2min timeout)
  ↓ NodePublishVolume(ctx) with csiTimeout
  ↓
CSI Driver (topolvm-node)
  ↓ GetLVList(ctx) RPC call to lvmd
  ↓
lvmd (gRPC server)
  ↓ vgservice.GetLVList(ctx) handler
    ↓ storagePoolForDeviceClass(ctx)
      ↓ FindVolumeGroup(ctx)
        ↓ getVGReport(ctx)
          ↓ callLVMInto(ctx)
            ↓ callLVMStreamed(ctx)
              ↓ exec.Command() ← CONTEXT IGNORED!
                ↓ vgs process starts
                  (running for duration > csiTimeout in kubelet)

After csiTimeout is exceeded:
  kubelet context timeout
  → gRPC connection close
  → CSI driver: "context canceled" error
  ← lvmd: vgs process still running
            (never received context cancellation)
```

**Why child processes remain:**
1. gRPC client (CSI driver) detects context timeout and closes connection
2. gRPC server (lvmd) RPC handler is still executing, unaware of the network-level closure
3. `exec.Command()` at line 76 ignores context, so spawned vgs/lvs processes are never sent SIGTERM/SIGKILL
4. Child processes continue running to completion (or indefinitely if blocked)
5. When handler eventually finishes and returns to client, connection is already dead
6. Over multiple timeout incidents, processes accumulate as resource leaks

**Impact**

- **Immediate:** CSI driver unable to publish volumes: `failed to list LV: context canceled` errors
- **Resource accumulation:** Each timeout incident leaves child LVM processes (vgs, lvs) running
- **System degradation:**
  - File descriptors consumed by orphaned processes
  - LVM lock contention worsens as processes accumulate
  - Future LVM operations become slower
- **Cascading failures:** As system degrades, more RPC requests timeout, creating more orphaned processes
- **Recovery:** May require manual cleanup (`pkill vgs`, `pkill lvs`) or lvmd restart to recover

SAY-5:
Fixes #1163.

`callLVMStreamed` threads `ctx` through its signature but hands it to `exec.Command`, which ignores context cancellation. When kubelet's `csiTimeout` (default 2 minutes) expires mid-RPC, the gRPC handler returns with `context canceled` but the underlying `vgs` / `lvs` / etc. process keeps running on the node. Every subsequent timeout adds another orphan, and under high LVM activity the accumulation is visible as unreaped processes consuming resources.

This switches to `exec.CommandContext(ctx, ...)` so a cancelled context triggers the usual Kill-on-cancel path. No behaviour change on the happy path; only the RPC-timeout path now cleans up the child.

### DIFF
--- a/internal/lvmd/command/lvm_command.go
+++ b/internal/lvmd/command/lvm_command.go
@@ -73,7 +73,11 @@ func callLVMInto(ctx context.Context, into any, logVerbosity int, args ...string
 func callLVMStreamed(ctx context.Context, logVerbosity int, args ...string) (io.ReadCloser, error) {
 	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithCallDepth(1))
 	wholeCommand := slices.Concat(lvmCommandPrefix, args)
-	cmd := exec.Command(wholeCommand[0], wholeCommand[1:]...)
+	// Use CommandContext so kubelet timing out on an RPC (default 2 min
+	// csiTimeout) cancels the underlying vgs/lvs/... subprocess instead
+	// of leaving it running on the node. exec.Command ignores ctx and
+	// accumulates orphans under repeated timeout scenarios (#1163).
+	cmd := exec.CommandContext(ctx, wholeCommand[0], wholeCommand[1:]...)
 	cmd.Env = os.Environ()
 	cmd.Env = append(cmd.Env, "LC_ALL=C")
 	return runCommand(ctx, logVerbosity, cmd)


cupnes:
@SAY-5 
Thanks for the PR. Could you please add a "Signed-off-by" line to your commit message? You can do this by running git commit --amend -s and pushing again.

SAY-5:
Done - added the Signed-off-by trailer and re-pushed. Also applied the same to #1171 preemptively.

cupnes (approved):
Thank you! LGTM.

satoru-takeuchi (approved):

PR #1196, opened 2026-07-03
Pin image digests

toshipp:
This is a practice to mitigate the impact of a compromised repository

### DIFF
--- a/.github/workflows/create-chart-update-pr.yaml
+++ b/.github/workflows/create-chart-update-pr.yaml
@@ -35,6 +35,18 @@ jobs:
             git switch main
             git pull
             git switch -c bump-chart-${{ inputs.chart-version }}
+      - name: "Resolve image digest"
+        id: image
+        run: |
+          digest=$(gh api --paginate "/orgs/topolvm/packages/container/topolvm-with-sidecar/versions" \
+            --jq "[.[] | select(.metadata.container.tags[]? == \"${{ inputs.app-version }}\") | .name] | first")
+          if [ -z "$digest" ] || [ "$digest" = "null" ]; then
+            echo "::error::Could not find a digest for tag ${{ inputs.app-version }} in ghcr.io/topolvm/topolvm-with-sidecar" >&2
+            exit 1
+          fi
+          echo "digest=${digest}" >> "$GITHUB_OUTPUT"
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
       - name: "Update files"
         run: |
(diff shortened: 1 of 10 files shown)


.github/workflows/create-chart-update-pr.yaml
```diff
@@ -35,13 +35,26 @@ jobs:
...
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
       - name: "Update files"
         run: |
           sed -r -i "s/checkout topolvm-chart-v[[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+/checkout topolvm-chart-v${{ inputs.chart-version }}/g" example/README.md
           sed -r -i "s/appVersion: [[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+/appVersion: ${{ inputs.app-version }}/g" charts/topolvm/Chart.yaml
           sed -r -i "s/^version: [[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+/version: ${{ inputs.chart-version }}/g" charts/topolvm/Chart.yaml
           sed -r -i "s/ghcr.io\/topolvm\/topolvm-with-sidecar:[[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+/ghcr.io\/topolvm\/topolvm-with-sidecar:${{ inputs.app-version }}/g" charts/topolvm/Chart.yaml
           sed -r -i "s/ghcr.io\/topolvm\/topolvm:[[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+/ghcr.io\/topolvm\/topolvm:${{ inputs.app-version }}/g" charts/topolvm/Chart.yaml
+          sed -r -i "s|^(  reference:).*|\1 ${{ inputs.app-version }}@${{ steps.image.outputs.digest }}|" charts/topolvm/values.yaml
```
cupnes: For consistency with the existing `sed` commands above (which all use `s/.../.../`), could you consider using `/` as the delimiter here as well instead of `|`?
```suggestion
          sed -r -i "s/^(  reference:).*/\1 ${{ inputs.app-version }}@${{ steps.image.outputs.digest }}/" charts/topolvm/values.yaml
```

cupnes (changes requested):
`example/Makefile:24` still uses `--set image.tag=devel`. Since the values now use `image.reference`, could you update that line too?

https://github.com/topolvm/topolvm/blob/401772d33fcae08eb1c09c8c3344d47384da599b/example/Makefile#L24

.github/workflows/create-chart-update-pr.yaml
toshipp: fair enough

toshipp:
> example/Makefile:24 still uses --set image.tag=devel. Since the values now use image.reference, could you update that line too?

Thanks, I'll fix it.

cupnes (approved):

satoru-takeuchi (approved):

PR #1197, opened 2026-07-06
Reduce the number of CI Jobs

toshipp:
This reduces the number of jobs from 28 to 24.

### DIFF
--- a/.github/workflows/build-images.yaml
+++ b/.github/workflows/build-images.yaml
@@ -17,17 +17,15 @@ on:
         description: "Push images"
         default: "false"
         type: string
+      test:
+        description: "Run container-structure-test"
+        default: "false"
+        type: string
 
 jobs:
   build-images:
     name: "build-images"
     runs-on: "ubuntu-22.04"
-    strategy:
-      fail-fast: true
-      matrix:
-        image:
-          - "normal"
-          - "with-sidecar"
     steps:
       - name: Disk cleanup
         run: |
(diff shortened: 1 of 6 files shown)


.github/workflows/build-images.yaml
```diff
@@ -41,16 +39,32 @@ jobs:
...
         env:
           TOPOLVM_VERSION: ${{ inputs.topolvm_version }}
           IMAGE_TAG: ${{ inputs.image_tag }}
           IMAGE_PREFIX: ${{ inputs.image_prefix }}
           PUSH: ${{ inputs.push }}
+      - name: Run container-structure-test
+        if: ${{ inputs.test == 'true' }}
+        run: |
+          make install-container-structure-test
+          make container-structure-test STRUCTURE_TEST_TARGET=with-sidecar
```
peng225, line 68: Previously, container structure test was executed with the target "normal" and "with-sidecar".
https://github.com/topolvm/topolvm/blob/401772d33fcae08eb1c09c8c3344d47384da599b/.github/workflows/main.yaml#L50

However, this PR changes the condition and now only the test with "with-sidecar" target runs. Is it intentional?

Makefile
```diff
@@ -23,9 +23,11 @@ ORIGINAL_IMAGE_TAG ?=
 STRUCTURE_TEST_TARGET ?= normal
 
 PUSH ?= false
-BUILDX_PUSH_OPTIONS := "-o type=tar,dest=build/topolvm.tar"
+BUILDX_BAKE_OPTIONS := --load
+# omit --load when pushing, because it implies --load
+# and the loading will failed due to the already exists error.
```
peng225:
```suggestion
# and the loading will fail due to "already exists" error.
```

.github/workflows/build-images.yaml
toshipp: Yes. The normal variant is always executed.
https://github.com/topolvm/topolvm/blob/main/Makefile#L191-L199

toshipp: Ah, I misunderstood the previous code. We should test the topolvm and topolvm-with-sidecar images. I'll fix it.

Makefile
toshipp: fixed https://github.com/topolvm/topolvm/compare/f8b958aa99766eb258b1b0d47fd1a93bd8dbc260..208f20d150b42ef092c6551f667f10e5b0f3959d

.github/workflows/build-images.yaml
toshipp: fixed: https://github.com/topolvm/topolvm/compare/f8b958aa99766eb258b1b0d47fd1a93bd8dbc260..208f20d150b42ef092c6551f667f10e5b0f3959d

peng225 (approved):

satoru-takeuchi (approved):

PR #1200, opened 2026-07-07
Add overhead-factor to device class for RAID capacity reporting

qJkee:
RAID logical volumes consume more raw space than their usable size (e.g., raid1 doubles every write). Without RAID awareness, TopoLVM reports raw VG free space, overstating usable capacity and causing provisioning failures when the scheduler places pods on nodes that cannot actually fulfill the request.

Add an overhead-factor field to the lvmd.yaml device class config. When set, reported available capacity is divided by this factor (available = vg_free / overhead_factor), giving the scheduler an accurate view of usable space. Defaults to 1.0 for backward compatibility.

### DIFF
--- a/internal/lvmd/device_class_manager.go
+++ b/internal/lvmd/device_class_manager.go
@@ -32,6 +32,15 @@ func GetSpare(dc *lvmdTypes.DeviceClass) uint64 {
 	return *dc.SpareGB << 30
 }
 
+// GetOverheadFactor returns the overhead factor for the device-class.
+// Returns 1.0 (no overhead) when not set.
+func GetOverheadFactor(dc *lvmdTypes.DeviceClass) float64 {
+	if dc.OverheadFactor == nil {
+		return 1.0
+	}
+	return *dc.OverheadFactor
+}
+
 // ValidateDeviceClasses validates device-classes
 func ValidateDeviceClasses(deviceClasses []*lvmdTypes.DeviceClass) error {
 	if len(deviceClasses) < 1 {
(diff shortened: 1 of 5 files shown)


qJkee:
@jakobmoellerdev @molpako can i get a review please?

qJkee:
@molpako  i see that error in CI because of github throttling (429 error). Unfortunately i can't re run the job, can you trigger it please?

qJkee:
All tests passed, can i get a review please?

satoru-takeuchi:
@qJkee I'll get back to you by the end of this week. I'm checking whether this could break the consistency of storage capacities.

qJkee:
@satoru-takeuchi sure, no problem. Thank you.
This change is RAID specific so it should not affect anything else

satoru-takeuchi:
@qJkee 

I hesitate to add a field like `overhead-factor` as an official feature for lvcreate options that users should be used [at their own risk](https://github.com/topolvm/topolvm/blob/18a0e461917c94bceec8f35a97dfb92a34c4845a/docs/limitations.md#use-lvcreate-options-at-your-own-risk).
 
I believe your problem will be mitigated by using [`StorageCapacityScoring`](https://kubernetes.io/blog/2025/04/30/kubernetes-v1-33-storage-capacity-scoring-feature/
). This feature will be beta in K8s v1.36.

I guess you're not using the topolvm scheduler, or your pods are created before PVCs. In that case, the pods might be deployed on nodes without enough free capacity. With `StorageCapacityScoring`, the kube-scheduler tries to deploy your pods on the node with the most available capacity, even in the situations described above.

Would it be acceptable for you to use `StorageCapacityScoring` instead?

qJkee:
@satoru-takeuchi 
Totally understand. Proposed `StorageCapacityScoring` solves another problem - eases to schedule pods on nodes with either the most or least available storage capacity. Even with scoring enabled, if topolvm reports 100GB free but only 80GB is usable due to overhead factor setting in RAID config, a 90GB PVC will still be scheduled there and fail.

That said, I completely understand the hesitation to add overhead-factor as an official feature in the lvcreate-options space. It's your call — happy to close this if you'd prefer not to go down this path.

qJkee:
@satoru-takeuchi @jakobmoellerdev @molpako 
Can we finalize this discussion?

satoru-takeuchi:
@qJkee Yes, I close this cussion and thank you for providing your feedback!

PR #1205, opened 2026-07-11
Micro-fix docs

mtchuikov:
There is no red arrows in the diagram, only the black ones.


llamerada-jp:
Could you sign off your commit? https://github.com/topolvm/topolvm/blob/main/CONTRIBUTING.md#dco

Issue #1201, opened 2026-07-07
CSIStorageCapacity objects missing after adding k8s nodes


briend:
**Describe the bug**
After adding nodes to Kubernetes, the CSIStorageCapacity objects may be missing until the topolvm-controller pods are restarted. The problem re-occurs after some amount of time (days or weeks). Without the `CSIStorageCapacity` objects, no stateful workloads are scheduled to the nodes. 

**Environments**
- Version: v0.41.0
- OS: Baremetal Kubernetes v1.34.8 on Ubuntu 24.04

**To Reproduce**
1. Upgrade Topolvm to `v0.41.0`
2. Wait a few days or weeks (unclear how to reproduce).
3. Add or re-add nodes to k8s
4. Observe `CSIStorageCapacity` objects and verify they are created for the new or re-provisioned nodes, if they are not, restart topolvm-controller pods.

**Expected behavior**
CSIStorageCapacity should be created within a few minutes after a node is added and the topolvm pods are running on it.

**Additional context**
- Previously we ran Topolvm `v0.36.1` for over a year without this problem ever happening, despite adding and re-adding many  k8s nodes
- We have over 1800 Kubernetes nodes, which might be a factor just due to scale
- We have fairly frequent node disk hardware failures.  Topolvm will produce errors trying to cleanup the PVC that it cannot due to the state of LVM on the node.  We're not sure but this could be a factor:

Example logs from csi-provisioner, we haven't seen anything specifically when adding the new nodes to k8s:

```
kubectl -n topolvm logs -l app.kubernetes.io/component=controller,app.kubernetes.io/instance=topolvm -c csi-provisioner -f --since=10m --timestamps
2026-07-07T18:58:17.149112411Z E0707 18:58:17.149055       1 controller.go:1054] "Unhandled Error" err="error syncing volume \"pvc-39c5b5b3-75b3-41a0-909a-83638d4a5eeb\": rpc error: code = DeadlineExceeded desc = context deadline exceeded" logger="UnhandledError"
2026-07-07T19:00:16.203704652Z E0707 19:00:16.203597       1 controller.go:1633] "Volume deletion failed" err="rpc error: code = DeadlineExceeded desc = stream terminated by RST_STREAM with error code: CANCEL" PV="pvc-39c5b5b3-75b3-41a0-909a-83638d4a5eeb"
[...]

PR #1207, opened 2026-07-15
add csi-provisoner metric scraping and extraArgs to containers

briend:
This PR adds additional csi-provisioner metric to the PodMonitor, as well as exposing `extraArgs` to the csi containers to allow configuration.  This helps resolve the issue with large k8s clusters here: https://github.com/topolvm/topolvm/issues/1201

### DIFF
--- a/charts/topolvm/README.md
+++ b/charts/topolvm/README.md
@@ -49,6 +49,9 @@ See [Getting Started](https://github.com/topolvm/topolvm/blob/topolvm-chart-v17.
 | env.topolvm_controller | list | `[]` | Specify environment variables for topolvm_controller container. |
 | env.topolvm_node | list | `[]` | Specify environment variables for topolvm_node container. |
 | env.topolvm_scheduler | list | `[]` | Specify environment variables for topolvm_scheduler container. |
+| extraArgs.csi_provisioner | list | `[]` | Specify extra arguments for csi_provisioner container. |
+| extraArgs.csi_resizer | list | `[]` | Specify extra arguments for csi_resizer container. |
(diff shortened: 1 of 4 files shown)


llamerada-jp:
Could you sign off your commit? https://github.com/topolvm/topolvm/blob/main/CONTRIBUTING.md#dco

charts/topolvm/Chart.yaml
```diff
@@ -4,7 +4,7 @@ type: application
 home: https://github.com/topolvm/topolvm
 name: topolvm
 description: Topolvm
-version: 17.0.0
```
llamerada-jp, line 7: We'll update version when release timing. So would you revert it?

llamerada-jp (changes requested):

charts/topolvm/Chart.yaml
briend: Will do, and I'll sign off.  Thanks!

briend: signed off and reverted the version bump

llamerada-jp (changes requested):
Would you run `make generate` to generate a document for Helm.

briend:
> Would you run `make generate` to generate a document for Helm.

I had trouble with `make generate` as I don't have an Ubunutu VM handy, but I ran `helm-docs -c charts/topolvm/` and that seems to have done the trick.

cupnes:
> I had trouble with make generate as I don't have an Ubunutu VM handy, but I ran helm-docs -c charts/topolvm/ and that seems to have done the trick.

Thanks for your work on this, and sorry for the late reply.
Could you also sign off that commit?

PR #1209, opened 2026-07-17
feat(chart): add podAnnotations for controller and node pods

wilfriedroset:
The controller and node pod templates only exposed podLabels, so there was no way to set pod annotations such as fluentbit.io/parser_<container> to select a log parser. The topolvm-controller and topolvm-node containers log JSON, but the CSI sidecars (csi-provisioner, csi-resizer, csi-snapshotter, csi-registrar, liveness-probe) emit klog, which a log collector cannot structure without a per-container parser annotation.

Add controller.podAnnotations and node.podAnnotations. On the node DaemonSet they merge with the existing metrics annotations rather than replacing them. Both default to {} and render no annotations block when unset.

### DIFF
--- a/charts/topolvm/README.md
+++ b/charts/topolvm/README.md
@@ -23,6 +23,7 @@ See [Getting Started](https://github.com/topolvm/topolvm/blob/topolvm-chart-v17.
 | controller.minReadySeconds | int | `nil` | Specify minReadySeconds. |
 | controller.nodeFinalize.skipped | bool | `false` | Skip automatic cleanup of PhysicalVolumeClaims when a Node is deleted. |
 | controller.nodeSelector | object | `{}` | Specify nodeSelector. # ref: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/ |
+| controller.podAnnotations | object | `{}` | Annotations to be set on the controller pod. |
 | controller.podDisruptionBudget.enabled | bool | `true` | Specify podDisruptionBudget enabled. |
 | controller.podLabels | object | `{}` | Additional labels to be set on the controller pod. |
 | controller.priorityClassName | string | `nil` | Specify priorityClassName. |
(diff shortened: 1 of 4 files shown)


charts/topolvm/README.md
```diff
@@ -106,6 +107,7 @@ See [Getting Started](https://github.com/topolvm/topolvm/blob/topolvm-chart-v17.
 | node.metrics.annotations | object | `{"prometheus.io/port":"metrics"}` | Annotations for Scrape used by Prometheus. |
 | node.metrics.enabled | bool | `true` | If true, enable scraping of metrics by Prometheus. |
 | node.nodeSelector | object | `{}` | Specify nodeSelector. # ref: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/ |
+| node.podAnnotations | object | `{}` | Annotations to be set on the node pods. |
```
molpako: Thanks for making the chart release non-breaking.

One thing to note is that, when `node.metrics.annotations` and `node.podAnnotations` share the same key, `node.podAnnotations` takes precedence. Could you update the description to make this clear?

molpako (changes requested):

charts/topolvm/Chart.yaml
```diff
@@ -4,7 +4,7 @@ type: application
 home: https://github.com/topolvm/topolvm
 name: topolvm
 description: Topolvm
-version: 17.0.0
+version: 17.1.0
```
molpako: Since Helm chart releases are handled by my team, would you drop this change?

molpako (changes requested):

wilfriedroset:
Thank you for your quick feedback, both of them are taken into account. I've amended my commit

molpako (dismissed):

ushitora-anqou (changes requested):
Could you fix the CI errors?

ushitora-anqou (approved):

molpako (approved):

PR #1227, opened 2026-09-02
dependabot: Add a gomod configuration to dependabot

llamerada-jp:
I tried to exclude K8s-related packages from dependabot, I found that Go lang wasn't listed as a target, so I added it with the exclusion settings.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,3 +12,25 @@ updates:
           - "*"
     cooldown:
       default-days: 14
+  - package-ecosystem: "gomod"
+    directory: "/"
+    rebase-strategy: "disabled"
+    schedule:
+      interval: "monthly"
+    groups:
+      gomod-dependencies-update:
+        patterns:
+          - "*"
+    cooldown:
+      default-days: 14
+    ignore:
+      - dependency-name: "github.com/prometheus/*"
+      - dependency-name: "google.golang.org/grpc"
+      - dependency-name: "google.golang.org/grpc/cmd/protoc-gen-go-grpc"
+      - dependency-name: "google.golang.org/protobuf"
+      - dependency-name: "k8s.io/api"
+      - dependency-name: "k8s.io/apimachinery"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -12,3 +12,18 @@ updates:
           - "*"
     cooldown:
       default-days: 14
+  - package-ecosystem: "gomod"
+    directories:
+      - "/"
```
cupnes: The `"github-actions"` ecosystem defined above uses `directory: "/"`. Could you use the same style here for consistency?

.github/dependabot.yml
```diff
@@ -12,3 +12,18 @@ updates:
...
+    rebase-strategy: "disabled"
+    schedule:
+      interval: "monthly"
+    groups:
+      gomod-dependencies-update:
+        patterns:
+          - "*"
+    cooldown:
+      default-days: 14
+    ignore:
```
cupnes, line 26: Could you apply the same change as the comment on the corresponding PR in pie?

https://github.com/topolvm/pie/pull/196#discussion_r3920043574

In topolvm as well, prometheus/* depends on the Kubernetes version, so it should be ignored here too.

.github/dependabot.yml
llamerada-jp: updated:
https://github.com/topolvm/topolvm/compare/5d8c34d7f0d0b156c2be47eb81b2eb9fe5883562..f7328f4298368b8a4f527c5b5d1e77963fe75386

.github/dependabot.yml
llamerada-jp: updated:
https://github.com/topolvm/topolvm/compare/5d8c34d7f0d0b156c2be47eb81b2eb9fe5883562..f7328f4298368b8a4f527c5b5d1e77963fe75386

cupnes (approved):

satoru-takeuchi (approved):

